PR #2, opened 2017-11-30
Add travis integration

andrewmil:
Added an initial "test" test to exercise this
Also added codecov and codacy integration


andrewmil:
I think coverage should be working but it's servers are having trouble atm

andrewmil:
I'm adding in more frontend setup first and after that I'll create a full Readme explaining how the application works.

PR #6, opened 2017-12-04
Add info endpoint

andrewmil:
Also added tests for this endpoint

### DIFF
--- a/response_operations_ui/views/info.py
+++ b/response_operations_ui/views/info.py
@@ -1,0 +1,23 @@
+import json
+from pathlib import Path
+
+from flask import make_response, jsonify
+
+from response_operations_ui import app
+
+
+@app.route('/info', methods=['GET'])
+def get_info():
+
+    _health_check = {}
+    if Path('git_info').exists():
+        with open('git_info') as io:
+            _health_check = json.loads(io.read())
+
+    info = {
+        "name": 'response-operations-ui',
+        "version": '0.0.1',
+    }
+    info = {**_health_check, **info}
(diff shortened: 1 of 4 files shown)


response_operations_ui/views/info.py
```diff
@@ -0,0 +1,23 @@
...
+from response_operations_ui import app
+
+
+@app.route('/info', methods=['GET'])
+def get_info():
+
+    _health_check = {}
+    if Path('git_info').exists():
+        with open('git_info') as io:
+            _health_check = json.loads(io.read())
```
JamesGardiner: probably want to catch the https://docs.python.org/3.6/library/json.html#json.JSONDecodeError here

response_operations_ui/views/info.py
```diff
@@ -1,18 +1,26 @@
...
 def get_info():
 
     _health_check = {}
     if Path('git_info').exists():
         with open('git_info') as io:
-            _health_check = json.loads(io.read())
+            try:
+                _health_check = loads(io.read())
+            except JSONDecodeError as e:
+                logger.exception('Failed to decode git_info json', exception=e)
```
JamesGardiner: You don't need the exception kwarg as the exception level logger will print any `exc_info` present anyway. See https://github.com/ONSdigital/eq-survey-runner/blob/e920e1dbfc633e7e6f6ed25002f139672674a8ae/application.py#L47 for an example of how to get the exception into the logged event in a splunk friendly way using `format_exc_info`.

PR #7, opened 2017-12-04
Prepare repo for cloudfoundry push

andrewmil:
Added Procfile and manifest.yml for cloudfoundry deployment

There is a version of the application currently pushed to the dev environment [here](http://response-operations-ui-dev.apps.devtest.onsclofo.uk)

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,6 +1,5 @@
 applications:
 - name: response-operations-ui
-  buildpack: https://github.com/cloudfoundry/staticfile-buildpack.git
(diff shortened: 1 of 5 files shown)


manifest.yml
```diff
@@ -0,0 +1,8 @@
+applications:
+- name: response-operations-ui
+  buildpack: https://github.com/cloudfoundry/staticfile-buildpack.git
```
tej99: No buildpack required for this service, it is only used in the backstage-ui (Angular app).

PR #8, opened 2017-12-08
Us001b view survey data

andrewmil:
Added view of surveys page, initially at the base url `/`

[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/166920222/US001b+View+Survey+Details)
[Trello](https://trello.com/c/4uPzDISJ/23-us001b-view-survey-data-8pts)

Styling for the view can be found in the `static/scss` folder.

This PR works in conjunction with https://github.com/ONSdigital/ras-backstage/pull/20 in backstage
If this is run with that PR then data can be passed to the table in the view, however, the surveys service is not returning the required data currently so it may look a bit like below.
screen shot 2017-12-08 at 15 46 05


andrewmil:
[US001b: View Survey Data (8pts)](https://trello.com/c/4uPzDISJ/23-us001b-view-survey-data-8pts)

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_surveys_list():
+    logger.debug('Retrieving surveys list')
+    url = '{}/{}'.format(app.config['BACKSTAGE_API_URL'], 'survey/surveys')
+
+    response = requests.get(url)
```
JamesGardiner: Are request exceptions handled anywhere?

JamesGardiner: Just seen the errorhandler below :+1

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -0,0 +1,24 @@
...
+    if response.status_code != 200:
+        logger.error('Failed to retrieve surveys list')
```
benjefferies: Is there value in logging here as `@app.errorhandler(ApiError)` will also log?

response_operations_ui/error_handlers.py
```diff
@@ -0,0 +1,23 @@
...
+
+from requests import RequestException
+from structlog import wrap_logger
+
+from response_operations_ui import app
+from response_operations_ui.exceptions.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
```
benjefferies: We probably want a `@app.errorhandler(Exception)` here too?

response_operations_ui/error_handlers.py
```diff
@@ -0,0 +1,23 @@
...
+
+@app.errorhandler(ApiError)
+def api_error(error):
+    logger.error('Api failed to retrieve required data',
```
benjefferies: It might be worth logging with logger.exception to capture the traceback

response_operations_ui/error_handlers.py
```diff
@@ -0,0 +1,23 @@
...
+                 url=error.url, status_code=str(error.status_code))
+    return "FAIL"
```
benjefferies: Might be cleaner to return a 500 status code here `return Response(status_code=500)`

tests/views/test_survey.py
```diff
@@ -0,0 +1,47 @@
...
+
+from response_operations_ui import app
+
+
+url_get_survey_list = '{}/{}'.format(app.config['BACKSTAGE_API_URL'], 'survey/surveys')
+with open('tests/test_data/survey/survey_list.json') as json_data:
+    survey_list = json.load(json_data)
+
+
+class TestInfo(unittest.TestCase):
```
benjefferies: TestInfo -> TestSurveys?

response_operations_ui/error_handlers.py
andrewmil: We'll probably add this in when we get our generic error pages from RAD

response_operations_ui/error_handlers.py
andrewmil: Is there any useful traceback coming from here? We log out the url we tried to hit and the status code that we got back, is there any benefit in seeing the code which triggered this?

response_operations_ui/error_handlers.py
andrewmil: The "FAIL" text is just a placeholder for actual error pages which we'll add in later stories

response_operations_ui/controllers/survey_controllers.py
andrewmil: probably not, I'll remove

response_operations_ui/error_handlers.py
benjefferies: I guess it's useful debug context to figure out the path the code followed before getting an error. It's probably fine for now because the api will only be called from one place but if we got to a point where the api was being called in different places it would definitely aid the debugging process

andrewmil: I've added it in. Instead of logging at exception level I've used the `format_exc_info` processor to include it in a error level log

benjefferies: 👍 but I thought you still had to `log.exception` to see it. @JamesGardiner do you know from when you did that little test?

response_operations_ui/static/scss/components/table.scss
```diff
@@ -0,0 +1,43 @@
+table {
```
tej99: Has this way of dealing with the css been agreed with @nicprice?

andrewmil: Meeting with Nic later today :+1:

response_operations_ui/error_handlers.py
andrewmil: You can see it from this, If you run the tests you can see it in the logs

response_operations_ui/static/scss/components/table.scss
andrewmil: @tej99 Confirmed current pattern with Nic

PR #10, opened 2017-12-15
Us002b view collection exercises

andrewmil:
Added view of single survey at /surveys/<short_name>
One extra function added to the controller to retrieve required data from backstage
This backstage endpoint depends on this [PR](https://github.com/ONSdigital/ras-backstage/pull/22)

[trello](https://trello.com/c/TZoUzqkn/24-us002b-view-collection-exercises-13pts)

### DIFF
--- a/tests/views/test_survey.py
+++ b/tests/views/test_survey.py
@@ -8,9 +8,12 @@ import requests_mock
 from response_operations_ui import app
 
 
-url_get_survey_list = '{}/{}'.format(app.config['BACKSTAGE_API_URL'], 'survey/surveys')
+url_get_survey_list = f'{app.config["BACKSTAGE_API_URL"]}/survey/surveys'
 with open('tests/test_data/survey/survey_list.json') as json_data:
     survey_list = json.load(json_data)
+url_get_survey_by_short_name = f'{app.config["BACKSTAGE_API_URL"]}/survey/shortname/bres'
+with open('tests/test_data/survey/survey.json') as json_data:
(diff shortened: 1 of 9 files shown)


andrewmil:
[US002b: View Collection Exercise(s) (13pts)](https://trello.com/c/TZoUzqkn/24-us002b-view-collection-exercises-13pts)

tests/views/test_survey.py
```diff
@@ -45,3 +48,20 @@ def test_survey_list_connection_error(self, mock_request):
...
+
+        self.assertEqual(response.status_code, 200)
+
+    @requests_mock.mock()
+    def test_survey_view_fail(self, mock_request):
+        mock_request.get(url_get_survey_by_short_name, status_code=500)
+
+        response = self.app.get("/surveys/bres")
+
+        self.assertEqual(response.status_code, 200)
```
benjefferies: Should the status_code be 200 when there is a failure?

andrewmil: I'm not sure what best practice is here. This is where we'll be rendering error screens, if we successfully render the error screen do we return a failing status code with it?

benjefferies: Yeah not massively sure. Can't find much around it. The main benefit I'd see is it will probably make testing easier when we test errors. Means we can just assert for http status codes vs parsing html?

andrewmil: We'd likely still have to parse html as we'll have different error pages in the future (404, 500, more specific ones).

PR #12, opened 2017-12-22
Create Dockerfile and update legalBasis

benjefferies:
Creating a dockerfile to be able to automatically create docker images with docker automation build https://hub.docker.com/r/sdcplatform/response-operations-ui/builds/ and updating legalbasis to match rm-survey-service json https://github.com/ONSdigital/rm-survey-service/pull/22/files#diff-4ec151648ae4c97b4d1668a4def08459R31

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3.6
+
+WORKDIR /response_operations_ui
+COPY . /response_operations_ui
+EXPOSE 8085
+RUN pip3 install pipenv==8.3.1 && pipenv install --deploy --system
+
+ENTRYPOINT ["python3"]
+CMD ["run.py"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.6
+
+WORKDIR /response_operations_ui
+COPY . /response_operations_ui
+EXPOSE 8085
+RUN pip3 install pipenv==8.3.1 && pipenv install --deploy --system
```
jcox-dev: Could start using the latest version (9.0.1) here.

benjefferies: Is there much value in pinning the version if we always want to use the latest? As it's for local development there is less risk in this. I don't see this being updated often otherwise

jcox-dev: I guess the only issue that could arise is if the flags (--deploy; --system) are changed in future versions but like you said this would be picked up in development.

PR #20, opened 2018-01-15
View list of collection instruments uploaded

benjefferies:
* User story *
As a Collection Exercise Coordinator
I need: to view the collection instruments that the system holds for the
given collection exercise
So that: I am happy that the collection instrument(s) loaded are correct

* Acceptance criteria *
Given the collection instruments have been loaded
When the user navigates to the collection exercise details page
Then they are able to see the filename (surveyid_formtype_period) of all
collection instruments that have been loaded

### DIFF
--- a/response_operations_ui/controllers/collection_instrument_controllers.py
+++ b/response_operations_ui/controllers/collection_instrument_controllers.py
@@ -1,4 +1,3 @@
-import json
 import logging
 
 import requests
--- a/response_operations_ui/controllers/collection_instrument_controllers.py
+++ b/response_operations_ui/controllers/collection_instrument_controllers.py
@@ -16,12 +16,3 @@ def upload_collection_instrument(short_name, period, file):
     response = requests.post(url, files={"file": (file.filename, file.stream, file.mimetype)})
     if response.status_code != 201:
         raise ApiError(response)
-
-
-def get_collection_instruments(short_name, period):
-    logger.debug('Getting collection instruments', short_name=short_name, period=period)
-    url = f'{app.config["BACKSTAGE_API_URL"]}/collection-instrument/{short_name}/{period}'
-    response = requests.get(url)
-    if response.status_code != 200:
-        raise ApiError(response)
-    return json.loads(response.text)
--- a/response_operations_ui/controllers/collection_instrument_controllers.py
+++ b/response_operations_ui/controllers/collection_instrument_controllers.py
@@ -1,3 +1,4 @@
+import json
 import logging
 
 import requests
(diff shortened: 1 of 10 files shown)


response_operations_ui/controllers/collection_instrument_controllers.py
```diff
@@ -15,3 +16,12 @@ def upload_collection_instrument(short_name, period, file):
...
         raise ApiError(response)
+
+
+def get_collection_instruments(short_name, period):
+    logger.debug('Getting collection instruments', short_name=short_name, period=period)
+    url = f'{app.config["BACKSTAGE_API_URL"]}/collection-instrument/{short_name}/{period}'
+    response = requests.get(url)
+    if response.status_code != 200:
+        raise ApiError(response)
+    return json.loads(response.text)
```
andrewmil: A debug level log line saying getting the collection instruments was successful here would be good for consistency with the other controller functions

response_operations_ui/views/collection_exercise.py
```diff
@@ -0,0 +1,47 @@
...
+from response_operations_ui.controllers import collection_exercise_controllers
+from response_operations_ui.controllers import collection_instrument_controllers
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+@app.route('/surveys/<short_name>/<period>', methods=['GET'])
+def view_collection_exercise(short_name, period):
+    ce_details = collection_exercise_controllers.get_collection_exercise(short_name, period)
+    cis = collection_instrument_controllers.get_collection_instruments(short_name, period)
```
andrewmil: I think there's a question around whether getting `cis` should be a separate call or if it should be returned as part of `ce_details` from the original backstage call. 
The pattern we've been using up til now would be to return it in the same call

tests/test_data/collection_exercise/collection_exercise_details.json
```diff
@@ -30,5 +30,19 @@
...
+          "e33daf0e-6a27-40cd-98dc-c6231f50e84a"
+        ],
+        "RU_REF": [],
+        "SURVEY_ID": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87"
+      },
+      "file_name": "collection_instrument.xlsx",
+      "id": "f732afbe-c710-4c95-a8d3-6644833195a7",
+      "surveyId": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87"
+    }
+  ]
```
rhiGriff: No new line at end of file.

tests/test_data/collection_exercise/collection_exercise_details_no_ci.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "survey": {
+    "id": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
+    "longName": "Business Register and Employment Survey",
+    "shortName": "BRES",
+    "surveyRef": "221"
+  },
+  "collection_instruments": [
+  ]
+}
```
rhiGriff: No new line at end of file.

tests/test_data/collection_exercise/collection_exercise_details.json
benjefferies: Don't think it's necessary in this case as it's json and it's only pep8 thing

tests/test_data/collection_exercise/collection_exercise_details_no_ci.json
benjefferies: Don't think it's necessary in this case as it's json and it's only pep8 thing

PR #21, opened 2018-01-16
Us008 9/preview and load sample

tej99:
DO NOT MERGE:- there are still a couple of questions around the wording and sample file contents.

As a: Collection Exercise Coordinator
I need: To be able to check the sample details before loading into the system
So that: I am reassured that the right sample is presented for loading

As a: Collection Exercise Coordinator
I need: To be able to load the sample 
So that: a sample file exists in the system


response_operations_ui/controllers/sample_controllers.py
```diff
@@ -0,0 +1,33 @@
...
+import requests
+from structlog import wrap_logger
+
+from response_operations_ui import app
+from response_operations_ui.exceptions.exceptions import ApiError
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def upload_sample(ce_id, file, total_businesses, total_ci):
```
andrewmil: We don't actually need `total_businesses` or `total_ci` to upload a sample. 
It might be better to separate out building the sample summary away from this function and building it in the view maybe?

response_operations_ui/static/js/read-csv.js
```diff
@@ -0,0 +1,99 @@
+
+
+
+function checkSampleContents(){
```
andrewmil: blank lines at the top and this function is empty

response_operations_ui/static/js/read-csv.js
```diff
@@ -0,0 +1,99 @@
...
+	reader.readAsText(fileToRead);
+}
+
+function loadHandler(event) {
+	var csv = event.target.result;
+	processData(csv);
+}
+
+function processData(csv) {
+    var allTextLines = csv.split(/\r\n|\n/);
```
andrewmil: There's some inconsistency in how this javascript is indented

response_operations_ui/static/js/read-csv.js
```diff
@@ -0,0 +1,99 @@
...
+	var ciCount = formTypes.filter(function(val, i, arr) {
+	    return arr.indexOf(val) === i;
+	}).length;
+
+
+	//Clear previous data
+	document.getElementById("sample-preview").innerHTML = "";
+
+	var preview = "";
+	preview = preview + "<div class='panel panel--info'>"
```
andrewmil: This all works fine so I think it's okay but directly injecting html like this feels a little bit messy

response_operations_ui/static/scss/components/load-button.scss
```diff
@@ -36,17 +37,99 @@
...
   background:#6a6a6a;
 }
 
+button.hidden{
+  display: none;
+}
+
 button.unready:hover{
   background:#6a6a6a;
 }
```
andrewmil: Missing newline

response_operations_ui/views/surveys.py
```diff
@@ -38,7 +38,33 @@ def view_collection_exercise(short_name, period):
...
+        return _upload_collection_instrument(short_name, period)
+
+
+def _upload_sample(short_name, period):
+    error = _validate_sample()
+    sample = False
+    total_businesses = request.form.get('sample-businesses')
+    total_ci = request.form.get('sample-collection-instruments')
+
+    ce_details = collection_exercise_controllers.get_collection_exercise(short_name, period)
```
jcox-dev: Currently the [PR in backstage](https://github.com/ONSdigital/ras-backstage/pull/29) expects the same URL structure ('/<short_name>/<period>') for uploading collection instruments as samples. Perhaps we don't need to get the relevant collection exercise here?

response_operations_ui/controllers/sample_controllers.py
tej99: Leaving this as it is because the submission time is linked to the upload.

response_operations_ui/static/js/read-csv.js
tej99: I kind of agree, but i'm just leaving it works fine.

response_operations_ui/static/scss/base.scss
```diff
@@ -175,17 +175,6 @@ img {
   width: auto;
 }
 
-.u-vh, .u-visuallyhidden {
```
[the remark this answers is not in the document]
tej99: line 154 has this

jcox-dev:
[Backstage updated](https://github.com/ONSdigital/ras-backstage/pull/29) ready to accept sample file uploads with the new endpoint.

PR #24, opened 2018-01-24
Usxxx internal user sign in

insacuri:
Adds the ability to sign-in / sign-out of the application using flask_login and blueprinted the views.

[Link to Trello card](https://trello.com/c/rxVbBK97)

Currently waiting on RAD for UI designs, so looks are basic for now, will only work correctly with [this](https://github.com/ONSdigital/ras-backstage/pull/39) pr in ras-backstage, which will need to be merged before this one.


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
+max-line-length = 160
```
andrewmil: 160 is a pretty long line length, we've been using 120 across most repos

Makefile
```diff
@@ -0,0 +1,7 @@
+build:
+	pipenv install
+	pipenv install --dev
+
+test:
+	pipenv check --style ./response_operations_ui ./tests
+	pipenv run python run_tests.py
```
andrewmil: Should we have a make command to run the application as well?

config.py
```diff
@@ -8,6 +8,12 @@ class Config(object):
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
 
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL', 'http://localhost:8001/backstage-api/v1')
+    RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
+
+    PASSWORD_MATCH_ERROR_TEXT = 'Your passwords do not match'
```
andrewmil: I think this text should be kept somewhere else, it's not really config

config.py
```diff
@@ -17,4 +23,8 @@ class DevelopmentConfig(Config):
 
 class TestingConfig(DevelopmentConfig):
     DEBUG = False
-    Testing = True
+    TESTING = True
+    LOGIN_DISABLED = True
+    WTF_CSRF_ENABLED = False
+
+    BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL', 'http://localhost:8001/backstage-api/v1')
```
andrewmil: This variable is inherited from the `Config` object so doesn't need to be defined here

response_operations_ui/__init__.py
```diff
@@ -2,11 +2,16 @@
...
 from flask import Flask
 from flask_assets import Bundle, Environment
+from flask_login import LoginManager
 
 from response_operations_ui.logger_config import logger_initial_config
+from response_operations_ui.user import User
 
 
 app = Flask(__name__)
+login_manager = LoginManager(app)
```
andrewmil: All this looks good, potentially we could pull all this initialisation around login into it's own file/function? Or at least group it together in this file so it's clear that it's all related

response_operations_ui/error_handlers.py
```diff
@@ -14,10 +16,22 @@
...
 @app.errorhandler(RequestException)
 def connection_error(error):
     logger.error('Failed to connect to external service', url=error.request.url)
-    return "FAIL"
+    return redirect(url_for('error_bp.server_error_page',
+                            _external=True,
+                            _scheme=getenv('SCHEME', 'http')))
+
+
+@app.errorhandler(Exception)
```
andrewmil: :+1:

response_operations_ui/templates/errors/500-error.html
```diff
@@ -0,0 +1,15 @@
+{% extends "layouts/base.html" %}
```
andrewmil: Did this error page come from RAD?
Definitely a lot better than what it was before :+1:

response_operations_ui/templates/home.html
```diff
@@ -1,16 +1,20 @@
 {% extends "layouts/base.html" %}
```
andrewmil: Did these UI changes come through RAD? This home page is already in the RAD designs so this should probably be passed by @nicprice

response_operations_ui/views/__init__.py
```diff
@@ -0,0 +1,16 @@
+from response_operations_ui import app
+from response_operations_ui.views.collection_exercise import collection_exercise_bp
+from response_operations_ui.views.home import home_bp
+from response_operations_ui.views.info import info_bp
+from response_operations_ui.views.errors import error_bp
+from response_operations_ui.views.logout import logout_bp
+from response_operations_ui.views.sign_in import sign_in_bp
+from response_operations_ui.views.surveys import surveys_bp
+
+app.register_blueprint(error_bp, url_prefix='/errors')
```
andrewmil: :+1: It's about time we were using blueprints in here

response_operations_ui/views/errors.py
```diff
@@ -0,0 +1,15 @@
...
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+error_bp = Blueprint('error_bp', __name__, template_folder='templates/errors')
+
+
+@error_bp.route('/500', methods=['GET', 'POST'])
+def server_error_page():
+    return render_template('errors/500-error.html', _theme='default',
+                           data={"error": {"type": "failed"}}), 500
```
andrewmil: Does this `data` variable do anything here?

response_operations_ui/views/info.py
```diff
@@ -1,31 +1,28 @@
...
+@info_bp.route('/', methods=['GET'])
+def get_info():
     info = {
-        "name": 'response-operations-ui',
-        "version": '0.0.1',
+        "name": "response-operations-ui",
+        "version": "0.0.1",
     }
-    info = {**_health_check, **info}
+    info = dict(_health_check, **info)
```
andrewmil: This all works fine but just out of interest is there an advantage to this syntax over the previous one?

response_operations_ui/views/sign_in.py
```diff
@@ -0,0 +1,44 @@
...
+    form = LoginForm(request.form)
+
+    if form.validate_on_submit():
+        sign_in_data = {
+            "username": request.form.get('username'),
+            "password": request.form.get('password'),
+        }
+        url = f'{app.config["BACKSTAGE_API_URL"]}/sign-in-uaa'
+
+        response = requests.post(url, json=sign_in_data)
```
andrewmil: The pattern followed by the rest of the application is to have requests to backstage handled in the controllers.
For consistency we could break some of this out into a file there

response_operations_ui/views/sign_in.py
```diff
@@ -0,0 +1,44 @@
...
+        response_json = json.loads(response.text)
```
andrewmil: According to me codecov the code from this line onwards isn't being covered

Makefile
andrewmil: Also if we've got a makefile now we should probably update the readme to mention the new commands

Makefile
```diff
@@ -0,0 +1,7 @@
+build:
+	pipenv install
+	pipenv install --dev
```
JamesGardiner: Do you need both lines? Won't pipenv install --dev also install any usual requirements?

response_operations_ui/error_handlers.py
```diff
@@ -14,10 +16,22 @@
...
+def server_error(error):  # pylint: disable=unused-argument
+    logger.exception('Uncaught exception generated', exception=error)
```
JamesGardiner: don't log the exception twice. Using logger.exception already logs out the exc_info if any is present

.flake8
dcdarrell9: Ideally we want it standardised throughout sdc, as it's different in eq, ras, sdx, open for discussion?

config.py
dcdarrell9: It's not used for now, so I'll remove until we get the rad designs

response_operations_ui/views/info.py
dcdarrell9: More for readability, shouldn't be any difference

response_operations_ui/error_handlers.py
```diff
@@ -14,10 +16,22 @@
 def api_error(error):
     logger.error('Api failed to retrieve required data', url=error.url,
                  status_code=str(error.status_code), exc_info=error)
-    return "FAIL"
+    return redirect(url_for('error_bp.server_error_page',
```
gemmai95: Should these always be displaying a server error to the user. As Api error is regularly not a 500 server error.

response_operations_ui/views/__init__.py
```diff
@@ -0,0 +1,16 @@
+from response_operations_ui import app
+from response_operations_ui.views.collection_exercise import collection_exercise_bp
+from response_operations_ui.views.home import home_bp
+from response_operations_ui.views.info import info_bp
+from response_operations_ui.views.errors import error_bp
```
gemmai95: These should probably be in alphabetical order.

response_operations_ui/views/collection_exercise.py
```diff
@@ -33,7 +35,7 @@ def view_collection_exercise(short_name, period):
                            breadcrumbs=breadcrumbs)
 
 
-@app.route('/surveys/<short_name>/<period>', methods=['POST'])
+@collection_exercise_bp.route('/<short_name>/<period>', methods=['POST'])
```
gemmai95: Does collection exercise not require a login to view?

response_operations_ui/error_handlers.py
insacuri: I copied this form ras-frontstage.  In that repo an api error has the same page rendered so I wanted to make it consistant.

Makefile
insacuri: Removed the duplicate line

response_operations_ui/error_handlers.py
insacuri: Agreed.  Removed it.

response_operations_ui/error_handlers.py
andrewmil: It's a system error but not necessarily in the UI, it's a question that came up in frontstage a lot. 
If one of the external Api's fail do we consider that different from if there's a failure directly in the UI?

response_operations_ui/views/errors.py
insacuri: It does nothing, removed

response_operations_ui/templates/errors/500-error.html
insacuri: Copied this from ras-frontstage, this'll probably be restyled down the line

response_operations_ui/templates/home.html
insacuri: This page is identical except for the log out link (which will be moved to a more sensible place when the styling gets applied).

response_operations_ui/__init__.py
insacuri: I grouped it all together as it's fairly small.  If this ends up being much bigger then I'll split it out.

response_operations_ui/templates/home.html
insacuri: I've updated the home page to look like the design.  Currently the scss is in the repo (and was given to me by leah) though there is talk of it using the cdn instead.  Short term though this isn't an issue.

response_operations_ui/views/sign_in.py
```diff
@@ -0,0 +1,40 @@
...
+
+        response_json = get_sign_in(sign_in_data)
+
+        if 'token' in response_json:
+            user = User(response_json['token'])
+            login_user(user)
+
+            next_url = request.args.get('next')
+
+            # Do we test if the redirect is safe or just assume it's fine?
```
gemmai95: Why is this comment a question?

insacuri: Removed it

tests/controllers/test_collection_exercise.py
```diff
@@ -0,0 +1,39 @@
...
+        }, 200)
+
+    return MockResponse(None, 404)
+
+
+# Our test case class
+class TestSurveyControllers(unittest.TestCase):
+
+    # We patch 'requests.get' with our own method. The mock object is passed in to our test case method.
+    @mock.patch('requests.get', side_effect=mocked_requests_get)
```
andrewmil: For consistency we don't create a custom response object to mock object
We use the `requests_mock` package in the other tests and across much of the ras services. If you haven't seen it before it's really really good and makes this kind of thing really easy :+1:

tests/controllers/test_survey_controllers.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+from unittest import mock
+
+from response_operations_ui.controllers import survey_controllers
+from response_operations_ui import app
+
+
+def mocked_requests_get(*args, **kwargs):
+    class MockResponse:
```
andrewmil: Same comment as above about mocking requests ^

response_operations_ui/error_handlers.py
```diff
@@ -13,11 +15,23 @@
 @app.errorhandler(ApiError)
 def api_error(error):
     logger.error('Api failed to retrieve required data', url=error.url,
-                 status_code=str(error.status_code), exc_info=error)
-    return "FAIL"
+                 status_code=str(error.status_code))
+    return redirect(url_for('error_bp.server_error_page',
+                            _external=True,
+                            _scheme=getenv('SCHEME', 'http')))
```
benjefferies: Not needed if we remove _external

response_operations_ui/forms.py
```diff
@@ -0,0 +1,16 @@
...
+from structlog import wrap_logger
+from wtforms import PasswordField, StringField, SubmitField
+from wtforms.validators import InputRequired
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+class LoginForm(FlaskForm):
+    id = StringField('id')
```
benjefferies: Linters might complain about reusing id as it will be already defined

response_operations_ui/views/errors.py
```diff
@@ -0,0 +1,14 @@
...
+from flask import Blueprint, render_template
+from structlog import wrap_logger
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+error_bp = Blueprint('error_bp', __name__, template_folder='templates/errors')
+
+
+@error_bp.route('/500', methods=['GET', 'POST'])
```
benjefferies: I guess this only needs to be GET, it's unlikely something will post to this page?

response_operations_ui/__init__.py
```diff
@@ -21,10 +23,21 @@
...
-from response_operations_ui.views import info  # NOQA # pylint: disable=wrong-import-position
+
+@login_manager.user_loader
+def user_loader(user_id):
+    # TODO This will need to be replaced with a call to reddis to get the token
+    # as we can't leave that sort of information in the cookie.  Will be implemented
+    # down the line once uaa is sorted out.
+    return User(user_id)
+
+import response_operations_ui.views  # NOQA # pylint: disable=wrong-import-position
```
warrenbailey: why are these imports at the bottom?

response_operations_ui/__init__.py
```diff
@@ -21,10 +23,21 @@
...
 
-from response_operations_ui import error_handlers  # NOQA # pylint: disable=wrong-import-position
-from response_operations_ui.views import collection_exercise, surveys  # NOQA # pylint: disable=wrong-import-position
-from response_operations_ui.views import info  # NOQA # pylint: disable=wrong-import-position
+
+@login_manager.user_loader
+def user_loader(user_id):
+    # TODO This will need to be replaced with a call to reddis to get the token
+    # as we can't leave that sort of information in the cookie.  Will be implemented
+    # down the line once uaa is sorted out.
```
warrenbailey: Need to add a card/task to the backlog so we don't forget this

response_operations_ui/controllers/sign_in_controller.py
```diff
@@ -0,0 +1,17 @@
...
+from structlog import wrap_logger
+
+from response_operations_ui import app
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_sign_in(sign_in_data):
+    logger.debug('Retrieving sign in authorisation')
+    url = f'{app.config["BACKSTAGE_API_URL"]}/sign-in-uaa'
```
warrenbailey: why is uaa in the url? shouldn't it just be sign-in?

response_operations_ui/forms.py
```diff
@@ -0,0 +1,16 @@
...
+from wtforms import PasswordField, StringField, SubmitField
+from wtforms.validators import InputRequired
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+class LoginForm(FlaskForm):
+    user_id = StringField('user_id')
+    username = StringField('Username', [InputRequired("Username is required")])
```
warrenbailey: Why do we need user_id and username?

response_operations_ui/templates/errors/500-error.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% block page_title %}Server error (Error 500) - ONS Business Surveys{% endblock %}
+
+{% block main %}
+        <h1 class="saturn">Error 500 - Server error</h1>
+
+        <div class="panel panel--simple panel--error">
+
+            <p>Something has gone wrong with the website.</p>
```
warrenbailey: Is this the correct wording? Doesn't read that well.

Also why website and not application?

response_operations_ui/controllers/sign_in_controller.py
```diff
@@ -0,0 +1,17 @@
...
+import requests
+from structlog import wrap_logger
+
+from response_operations_ui import app
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_sign_in(sign_in_data):
+    logger.debug('Retrieving sign in authorisation')
```
warrenbailey: I don't think this is the correct name for the function. Please rename to sign_in or something similar

response_operations_ui/controllers/sign_in_controller.py
```diff
@@ -0,0 +1,17 @@
...
+    url = f'{app.config["BACKSTAGE_API_URL"]}/sign-in-uaa'
+
+    response_json = requests.post(url, json=sign_in_data)
+    return json.loads(response_json.text)
```
warrenbailey: If use response.json() instead, then you don't need the json.loads call

response_operations_ui/__init__.py
dcdarrell9: I **think** it's something to do with the way Flask is structured and the views need to imported after `app = Flask(__name__)`

response_operations_ui/controllers/sign_in_controller.py
dcdarrell9: We added this as a new resource, but I think we can remove the old sign-in, and if we can, just replace with this

response_operations_ui/templates/errors/500-error.html
dcdarrell9: We haven't had the final designs from RAD yet or whether the stylesheets will be pulled from elsewhere

config.py
```diff
@@ -17,4 +18,6 @@ class DevelopmentConfig(Config):
 
 class TestingConfig(DevelopmentConfig):
     DEBUG = False
-    Testing = True
+    TESTING = True
```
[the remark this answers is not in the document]
andrewmil: It's a standard flask environment variable but I'm not sure if it's used anywhere

response_operations_ui/error_handlers.py
```diff
@@ -13,11 +15,23 @@
 @app.errorhandler(ApiError)
 def api_error(error):
     logger.error('Api failed to retrieve required data', url=error.url,
-                 status_code=str(error.status_code), exc_info=error)
-    return "FAIL"
+                 status_code=str(error.status_code))
+    return redirect(url_for('error_bp.server_error_page',
+                            _external=True,
```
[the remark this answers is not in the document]
andrewmil: This was related to a weird bug we were having in production a while ago, I'd recommend against removing it

response_operations_ui/__init__.py
andrewmil: Yeah this is a common pattern in flask, you need to import the endpoints to register them but you need the app object to be created first. Otherwise you get weird cyclical import problems

response_operations_ui/controllers/sign_in_controller.py
andrewmil: We cannot remove the old sign-in endpoint. It's being used by the angular application which is getting used for BRES

response_operations_ui/controllers/sign_in_controller.py
```diff
@@ -0,0 +1,16 @@
...
+from response_operations_ui import app
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def sign_in(sign_in_data):
+    logger.debug('Retrieving sign in authorisation')
+    url = f'{app.config["BACKSTAGE_API_URL"]}/sign-in-uaa'
+
+    response = requests.post(url, json=sign_in_data)
```
andrewmil: Our controllers generally have a debug level success log on exit

tests/views/test_collection_exercise.py
```diff
@@ -18,6 +19,9 @@
 class TestCollectionExercise(unittest.TestCase):
 
     def setUp(self):
+        app_config = TestingConfig()
```
andrewmil: If this method of setting things up to use TestingConfig works it's definitely the way to go. We should probably do this in all our other test files as well

dcdarrell9: I'm pretty sure it works, but looking at the config documentation on flask, they have something like this which we could look at...   `app.config.from_object('configmodule.ProductionConfig')`

response_operations_ui/controllers/sign_in_controller.py
warrenbailey: We need some error handling in this method too

tests/views/test_collection_exercise.py
JamesGardiner: The above is functionally the same as passing a string to a module that contains a config class.

config.py
```diff
@@ -17,4 +18,5 @@ class DevelopmentConfig(Config):
 
 class TestingConfig(DevelopmentConfig):
     DEBUG = False
-    Testing = True
```
JamesGardiner: `TESTING` should be True. This stops any exceptions being handled by the flask app object and allows them to propagate to the unittest module.

config.py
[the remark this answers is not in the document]
JamesGardiner: See my comment above @benjefferies

response_operations_ui/__init__.py
JamesGardiner: Done

tests/test_error_handlers.py
```diff
@@ -0,0 +1,17 @@
+import unittest
+from unittest.mock import patch
+
+from response_operations_ui import app
+
+
+class TestErrorHandlers(unittest.TestCase):
+    def setUp(self):
+        app.config['WTF_CSRF_ENABLED'] = False
```
andrewmil: This is set in `TestingConfig` probably just set this test up to use that

PR #26, opened 2018-01-24
US0010 view sample

tej99:
As a CEC when a sample is loaded in to a CE

I need to understand the content of a sample

So that I am reasonably assured that it is the right sample file for the collection exercise


tej99:
[US010: View sample in context of a CE](https://trello.com/c/9zdn4R3Q/26-us010-view-sample-in-context-of-a-ce)

benjefferies:
Samples that have already been uploaded the rm-sample-service changes will look like
image

PR #30, opened 2018-01-26
Handle date format changes services with iso8601 package

jcox-dev:
The Java services are now consolidating the datetime format that is returned in their JSON to be more flexible but a side-effect of this is that in the main Java returns "Zulu time" using ISO8601. 

This PR adds a package for specifically parsing strings in this format (`%Y-%m-%dT%H:%M:%S.%f%z`)  and returning datetime objects. (NB: Let's not use this as an excuse to mangle things on backstage).

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,7 +20,7 @@ cssmin = "==0.2.0"
 jsmin = "==2.2.2"
 gunicorn = "==19.7.1"
 requests = "==2.18.4"
-"iso8601" = "*"
+"iso8601" = "==0.1.12"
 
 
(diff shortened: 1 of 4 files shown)


Pipfile
```diff
@@ -20,6 +20,7 @@ cssmin = "==0.2.0"
 jsmin = "==2.2.2"
 gunicorn = "==19.7.1"
 requests = "==2.18.4"
+"iso8601" = "*"
```
tej99: Can we pin this to a specific version? "==0.1.12"

PR #32, opened 2018-01-31
Fixes redirect on unauthorised error

insacuri:


response_operations_ui/views/sign_in.py
```diff
@@ -28,8 +28,9 @@ def sign_in():
         if 'token' in response_json:
             user = User(response_json['token'])
             login_user(user)
-
-            next_url = request.args.get('next')
-            return redirect(next_url or url_for('home_bp.home'))
+            if 'next' in session:
+                return redirect(session['next'])
```
andrewmil: This case isn't tested

response_operations_ui/views/sign_in.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-from flask import Blueprint, redirect, render_template, request, url_for
+from flask import Blueprint, redirect, render_template, request, url_for, session
```
JamesGardiner: I'm going to be nit picking but this should be alphabetized

insacuri: Fixed

response_operations_ui/views/sign_in.py
insacuri: Tested now

response_operations_ui/views/sign_in.py
```diff
@@ -31,8 +31,9 @@ def sign_in():
         if 'token' in response_json:
             user = User(response_json['token'])
             login_user(user)
-
-            next_url = request.args.get('next')
-            return redirect(next_url or url_for('home_bp.home'))
+            if 'next' in session:
+                return redirect(session['next'])
+            else:
```
jcox-dev: 🎨 superfluous `else` statement here.

PR #36, opened 2018-02-05
For story us040 updated collection-exercise.html

muhamf:


muhamf:
Please review collection-exercise.html for story us040

PR #35, opened 2018-02-02
US011 - Display state of a collection exercise

emilioward:
[https://trello.com/c/kJO9QkLd](https://trello.com/c/kJO9QkLd)

Added status column and data to the survey details page.

Data needs to be loaded in to test this functionality.

Don't merge until other user stories related to this card have been tested against this PR.
These states might change as they are, but the mapping functionality works.

PR #37, opened 2018-02-06
For story us040 change in collection-exercise.html

muhamf:


response_operations_ui/templates/collection-exercise.html
```diff
@@ -6,6 +6,18 @@
...
+    {% if ce.state == "INIT" %}
+       <dl class="survey-info info-panel" id="survey-attributes">
+           <dt class="survey-info__title">
+             Status:
+           </dt>
+          <dd class="survey-info__data">
+             Created
+          </dd>
+       </dl>
+  {% endif %}
```
davidmort: formatting - this should be in line with open brackets above

response_operations_ui/templates/collection-exercise.html
```diff
@@ -6,6 +6,18 @@
 <div role="main" id="main" class="page__main">
     <br/>
     <h1 class="jupiter" name="page-ce-title">{{survey.shortName}} {{ce.exerciseRef}}</h1>
+
+      {% if ce.state == "INIT" %}
```
jcox-dev: Rather than mapping the states in the template, I wonder if we should do something similar to [the PR for US011](https://github.com/ONSdigital/response-operations-ui/pull/35/files).

davidmort:
looks like this has been implemented in https://github.com/ONSdigital/response-operations-ui/pull/49, so will close this PR

PR #39, opened 2018-02-06
Apply additional styling to login page

insacuri:
First attempt to make it look like the designs in confluence (Found here for those who have access: https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/226230282/US048+Internal+User+Signs+into+SDC)

The styles aren't perfect as a number of stylesheets are only in the CDN, which I think we're still on the fence about using.

### DIFF
--- a/response_operations_ui/templates/sign_in.html
+++ b/response_operations_ui/templates/sign_in.html
@@ -2,44 +2,71 @@
 
 {% block page_title %} Sign-In | Survey Data Collection {% endblock %}
 
-
 {% block main %}
 
 <div class="container page__container">
-  <div class="grid">
-    <div class="grid__col col-12@m">
-      {% with messages = get_flashed_messages(with_categories=true) %}
-        {% if messages %}
-          <ul class=flashes>
-            {% for category, message in messages %}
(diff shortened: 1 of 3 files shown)


dcdarrell9:
Couple of things...

screen shot 2018-02-07 at 08 10 25

- I think there's too much of a gap above and below the error message, and are we going to be clicking on the 'username required' link and then being put into the username entry box.

screen shot 2018-02-07 at 08 10 03

- If you enter a user and no pass the error box is too close to the entry field.

- The entry fields should have the same style as from frontstage / pattern library, *I think* something like `class=input input--text input-type__input`

screen shot 2018-02-07 at 08 32 24

- Should the successfully signed out have a darker green banner title at the top?

- Also I think we may want a gap before the 'Sign in' after an error message?

insacuri:
@dcdarrell9 you're not wrong, but I also don't want to fix them at the moment.  A bunch of these styles are in the sdc-global-design-patterns repo that we're not currently using (and we might be using... still need to find out if we're going to use the CDN or not).
This PR is less 'here's how it's going to look' and more 'it was a million miles out before, and now it isn't'

PR #42, opened 2018-02-08
Sm111 messages

insacuri:
Unit tests will be with this PR soon but I'm putting it up now for early code review so it can go in asap.


response_operations_ui/templates/messages.html
```diff
@@ -0,0 +1,70 @@
...
+      <div class="messages-list">
+
+        <table id="tbl-messages" class="table__dense" summary="messages in the system">
+          <thead class="table--head">
+            <tr class="table--row">
+              <th scope="col" class="table--header--cell tbl-messages-RU-Ref">
+                RU_Ref
+              </th>
+              <th scope="col" class="table--header--cell tbl-messages-business">
+                Business
```
gemmai95: Is this meant to be business name rather than business?

tests/views/test_message.py
```diff
@@ -0,0 +1,38 @@
...
+
+get_message_list = f'{app.config["BACKSTAGE_API_URL"]}/v1/secure-message/messages'
+with open('tests/test_data/survey/survey_list.json') as json_data:
+    message_list = json.load(json_data)
+
+
+class TestMessage(unittest.TestCase):
+
+    def setUp(self):
+        app_config = TestingConfig()
```
gemmai95: Not sure this is needed as this will run as part of run_tests which is set up to use testing config.

response_operations_ui/views/messages.py
```diff
@@ -0,0 +1,31 @@
...
+logger = wrap_logger(logging.getLogger(__name__))
+
+messages_bp = Blueprint("messages_bp", __name__, static_folder='Static',
+                        template_folder='templates')
+
+
+@messages_bp.route('/', methods=['GET'])
+@login_required
+def view_messages():
+    messages = message_controllers.get_message_list()
```
JamesGardiner: Should the messages not be refined by the controller? I don't see why this is being changed here to fit a template.

tests/views/test_message.py
insacuri: Fixed

response_operations_ui/templates/messages.html
```diff
@@ -0,0 +1,70 @@
...
+              </th>
+              <th scope="col" class="table--header--cell tbl-messages-received">
+                Received
+              </th>
+            </tr>
+          </thead>
+
+              <tbody>
+              {% if not messages %}
+                <p>There are no messages for this user</p>
```
gemmai95: This text doesn't match up with the design.

response_operations_ui/views/messages.py
```diff
@@ -0,0 +1,31 @@
...
+    refined_messages = []
+    for message in messages:
+        refined_message = {
+            'ru_ref': message.get('ru_id'),
+            'business_name': message.get('@ru_id').get('name'),
+            'subject': message.get('subject'),
+            'from': message.get('msg_from'),
+            'to': message.get('@msg_to')[0].get('firstName') + ' ' + message.get('@msg_to')[0].get('lastName'),
+            'sent_date': message.get('sent_date')
+        }
```
JamesGardiner: defined_messages = [define(msg) for msg in messages]

response_operations_ui/controllers/message_controllers.py
```diff
@@ -0,0 +1,26 @@
+import logging
+import jwt
+import requests
+from structlog import wrap_logger
+
+from response_operations_ui import app
```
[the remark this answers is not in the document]
insacuri: There isn't a current_app in response_operations_ui.  This could be tweaked down the line if it's needed but shouldn't impact this PR

response_operations_ui/controllers/message_controllers.py
```diff
@@ -0,0 +1,39 @@
+import jwt
```
JamesGardiner: jwt, requests and structlog  are not in python's standard library

response_operations_ui/controllers/message_controllers.py
```diff
@@ -0,0 +1,39 @@
...
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_message_list():
+    logger.debug("Retrieving Message list")
+
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/messages'
+    # This will be removed once UAA is completed.  For now we need the call to backstage to include
+    # an Authorization in its header a JWT that includes party_id and role.
+    encoded_jwt = jwt.encode({'party_id': 'BRES', 'role': 'internal'}, 'testsecret', algorithm='HS256')
```
JamesGardiner: When we start properly encoding JWT payloads, this will need to handle a number of possible exceptions. From looking at the JWT docs, TypeError and ValueError (from the mapping of the JSON) are the two obvious ones.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -0,0 +1,39 @@
...
+    except HTTPError:
+        logger.exception("Message retrieval failed")
+        raise ApiError(response)
+
+    logger.debug("Retrieval successful")
+    try:
+        messages = response.json()['messages']
+        return messages
+    except KeyError:
+        # TODO: Look to fail more gracefully.  Returning an empty list will display
```
JamesGardiner: This comment is spot on, but should this be part of the implementation?

insacuri: It's not essential and doesn't have a design, but when there is a one then that can go into a short lived improvement branch.  I just want this merged so other stories can start using it.

response_operations_ui/controllers/message_controllers.py
insacuri: Agreed, but does this need to be done now?

PR #45, opened 2018-02-08
Us044 display ru details

andrewmil:
Added new page at /reporting-units/<ru_ref>
Currently it only displays the most recent reporting unit name and the RU_REF

[Trello](https://trello.com/c/zHV82xsH/77-us044-int-view-ru-details-med)
[Confluence](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/248643696/US044+View+RU+Details)

### DIFF
--- a/tests/views/test_reporting_units.py
+++ b/tests/views/test_reporting_units.py
@@ -28,3 +28,13 @@ class TestReportingUnits(unittest.TestCase):
 
         self.assertEqual(response.status_code, 200)
         self.assertIn("Bolts and Ratchets Ltd".encode(), response.data)
+        self.assertIn("50012345678".encode(), response.data)
+
+    @requests_mock.mock()
+    def test_get_reporting_unit_fail(self, mock_request):
+        mock_request.get(url_get_reporting_unit, status_code=500)
+
+        response = self.app.get("/reporting-units/50012345678", follow_redirects=True)
+
+        self.assertEqual(response.status_code, 500)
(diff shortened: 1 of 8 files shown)


andrewmil:
[US044: INT View RU Details [Med]](https://trello.com/c/zHV82xsH/77-us044-int-view-ru-details-med)

tests/views/test_reporting_units.py
```diff
@@ -0,0 +1,40 @@
...
+url_get_reporting_unit = f'{app.config["BACKSTAGE_API_URL"]}/v1/reporting-unit/50012345678'
+with open('tests/test_data/reporting_units/reporting_unit.json') as json_data:
+    reporting_unit = json.load(json_data)
+
+
+class TestReportingUnits(unittest.TestCase):
+
+    def setUp(self):
+        app_config = TestingConfig()
+        app.config.from_object(app_config)
```
gemmai95: Why are you specifying here to use the TestingConfig when this will only be run in the test suite already using TestingConfig?

andrewmil: If we want to run individual tests then we'd have to manually set `APP_SETTINGS=TestingConfig` with environment variables. I think this removes this requirement?

andrewmil: This will allow the tests to be run individually/by case without using the run_tests.py script

On a similar note I think we could maybe get rid of that script in favour of using a Makefile

dcdarrell9: I agree with that ^

PR #51, opened 2018-02-12
Sm101 create message

AdamHawtin:
Created a page for creating a secure message in the context of an RU/Respondent.

Also adds mock create message links to the 'reporting-unit' page as a placeholder until this page is created properly. RU details are currently passed to the create-message page by a query string dictionary, this is not a final implementation, just a placeholder way to pass the information we need until a final solution is decided.

Currently the uuid's for the 'to' and 'from' are hardcoded to match uuid's in ras-secure-message mock party service. If you run ras-secure-message locally with the mock party service toggled on you can successfully send a message. This will be fixed when we are being passed the uuid's resolved by the reporting-unit page.

There is a link on the landing page to create a message with hardcoded details.

There is a branch of ras-integration-tests with feature tests for this story:
https://github.com/ONSdigital/ras-integration-tests/tree/SM101-create-message

Story Confluence page:
https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/223281382/SM101+Internal+User+to+Send+Message

Trello Card:
https://trello.com/c/ig9tz8A8/21-sm101-internal-user-to-send-message-8


response_operations_ui/templates/create-message.html
```diff
@@ -0,0 +1,112 @@
+{% extends "layouts/base.html" %}
+
+{% block page_title %} Create Message | Survey Data Collection {% endblock %}
+
+{% block main %}
+<style>
```
insacuri: These should be in their own .scss file

response_operations_ui/templates/create-message.html
```diff
@@ -0,0 +1,112 @@
...
+            {% if form.errors|length > 1 %}
+                <h1 class="panel__title venus">This page has {{ form.errors|length }} errors</h1>
+            {% elif form.errors %}
+                <h1 class="panel__title venus">This page has {{ form.errors|length }} error</h1>
+            {% endif %}
+        </div>
+        <div class="panel__body">
+            <ol>
+                {% for error in form.errors %}
+                <li class="panel__title venus">{{ form.errors[error][0] }}</li>
```
insacuri: Can a field produce multiple errors at the same time?  If so then this'll need to change.  If not, then theres no issue.

response_operations_ui/templates/create-message.html
```diff
@@ -0,0 +1,112 @@
...
+            {% if form.subject.errors %}
+                </div>
+            {% endif %}
+            <br />
+
+
+
+            {% if form.body.errors %}
+                <div class="panel panel--simple panel--error panel--spacious" id="body-error">
+                    <p class="error-message">{{ form.errors['body'][0] }}</p>
```
insacuri: If you're getting errors from a single field, it's something ilke form.<fieldname>.errors  and you loop over it

response_operations_ui/templates/create-message.html
```diff
@@ -0,0 +1,112 @@
...
+            {% endif %}
+            <table width="40%">
+                <tr align="left" valign="bottom">
+                    <td align="left" valign="bottom">
+                        <label class="venus" for="body">Message</label>
+                        <p class="u-vh">Message body is required</p>
```
insacuri: Is this right?  Do we let screen readers know that a field is mandatory with hidden fields?

response_operations_ui/templates/create-message.html
```diff
@@ -0,0 +1,112 @@
...
+                <tr>
+                    <td align="left" valign="top">
+                        {{ form.subject(id='secure-message-subject',
+                        class_='secure-message-form__subject input input--text',
+                        maxlength='96', size=80) }}
+                    </td>
+                </tr>
+            </table>
+            {% if form.subject.errors %}
+                </div>
```
insacuri: I'm not a fan of this way of doing this.  It seems messy to have 2 {% if error %} blocks.  I might be wrong though and the whole thing needs to be encapsulated in a single div

response_operations_ui/templates/partials/mock-create-message-link.html
```diff
@@ -0,0 +1,12 @@
+<div class="grid">
```
insacuri: Is this necessary?  This HTML could just be in the reporting unit template with a comment saying 'remove me once X happens'.  This might get forgotten about when that partial is removed and then we'll have an unused file messing up the place

response_operations_ui/views/messages.py
```diff
@@ -1,15 +1,92 @@
...
+            form.ru_ref.text = form.hidden_ru_ref.data
+            form.business.text = form.hidden_business.data
+            form.to.text = form.hidden_to.data
+            form.body.data = g.form_body_data
+            form.subject.data = g.form_subject_data
+
+            form.errors['sending'] = ["Something went wrong: Message failed to send"]
+
+            return render_template('create-message.html',
+                                   _theme='default',
```
insacuri: I think this _theme line can be removed

response_operations_ui/views/messages.py
```diff
@@ -1,15 +1,92 @@
...
+            form.hidden_to_uuid.data = ru_dict.get('to_uuid')[0]
+            form.hidden_to_ru_id.data = ru_dict.get('to_ru_id')[0]
+
+        form.survey.text = form.hidden_survey.data
+        form.ru_ref.text = form.hidden_ru_ref.data
+        form.business.text = form.hidden_business.data
+        form.to.text = form.hidden_to.data
+
+        return render_template('create-message.html',
+                               _theme='default',
```
insacuri: This can be removed

response_operations_ui/templates/create-message.html
AdamHawtin: I was told to include this be Leah

response_operations_ui/templates/create-message.html
AdamHawtin: This is temporary css until the styling is sorted out properly. This css shouldn't be necessary but we're still waiting on the UX people to hear how the css should be done.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -37,3 +37,17 @@ def get_message_list(params):
...
+    logger.debug("Sending messsage")
+
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/send-message'
+    # This will be removed once UAA is completed.  For now we need the call to backstage to include
+    # an Authorization in its header a JWT that includes party_id and role.
+    encoded_jwt = jwt.encode({'user': 'BRES', 'party_id': 'BRES', 'role': 'internal'}, 'testsecret', algorithm='HS256')
+
+    response = requests.post(url, headers={'Authorization': encoded_jwt, 'Content-Type': 'application/json',
+                                           'Accept': 'application/json'}, data=message_json)
+    if response.status_code != 201:
```
gavinleeedwards: Following on from James' change, should this now be checking for !=200 rather than 201?

response_operations_ui/views/messages.py
```diff
@@ -1,15 +1,92 @@
...
+
+        # Keep the message subject and body
+        g.form_subject_data = form.subject.data
+        g.form_body_data = form.body.data
+
+        try:
+            message_controllers.send_message(message_json)
+
+        except ApiError:
+
```
JamesGardiner: Additional logging here for debugging would be useful

response_operations_ui/views/messages.py
```diff
@@ -1,15 +1,92 @@
...
+            return render_template('create-message.html',
+                                   _theme='default',
+                                   form=form,
+                                   breadcrumbs=breadcrumbs)
+        return redirect(url_for('messages_bp.view_messages'))
+
+    else:
+        if not form.is_submitted():
+            query_ru = request.args.get('ru_details')
+            ru_dict = parse_qs(query_ru)
```
JamesGardiner: This should be available to you as an attribute on the request itself. See https://github.com/pallets/werkzeug/blob/8b1799738de72dd45d1136d943a049d45617ad09/werkzeug/wrappers.py#L651

response_operations_ui/views/messages.py
```diff
@@ -1,16 +1,93 @@
...
+        form.ru_ref.text = form.hidden_ru_ref.data
+        form.business.text = form.hidden_business.data
+        form.to.text = form.hidden_to.data
+
+        return render_template('create-message.html',
+                               form=form,
+                               breadcrumbs=breadcrumbs)
+
+
+def _build_create_message_breadcrumbs():
```
dcdarrell9: I'd argue these breadcrumbs shouldn't be hardcoded and this is a step in the right way, but for now would it be worth not having it as a function just to keep it consistent with the rest of the repo?

insacuri: I disagree, before putting this (and other things) in their own function the create_message function was super long and hard to read.   This keeps the core of that function small and the functions name is highly descriptive so I see no reason to change it.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -33,3 +33,20 @@ def get_message_list(params):
...
         raise NoMessagesError
+
+
+def send_message(message_json):
+    logger.debug("Sending message")
+
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/send-message'
+    # This will be removed once UAA is completed.  For now we need the call to backstage to include
+    # an Authorization in its header a JWT that includes party_id and role.
+    encoded_jwt = jwt.encode({'user': 'BRES', 'party_id': 'BRES', 'role': 'internal'}, 'testsecret', algorithm='HS256')
```
gemmai95: Is there a reason this is still hardcoded to BRES?

response_operations_ui/controllers/message_controllers.py
```diff
@@ -33,3 +33,20 @@ def get_message_list(params):
     except KeyError:
         logger.exception("Response was successful but didn't contain a 'messages' key")
         raise NoMessagesError
+
+
+def send_message(message_json):
+    logger.debug("Sending message")
+
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/send-message'
+    # This will be removed once UAA is completed.  For now we need the call to backstage to include
```
gemmai95: Would it be better putting this as a todo rather than a comment?

response_operations_ui/templates/home.html
```diff
@@ -10,6 +10,7 @@
       <ul class="u-pt-l">
         <li><a name="surveys-link" href="/surveys/">View list of business surveys</a></li>
         <li><a name="messages-link" href="/messages/">View messages</a></li>
+        <li><a name="create-message-link" href="/messages/create-message?ru_details=survey%3DBRES%2B2017%26ru_ref%3D49900000280%26business%3DBolts%2B%2526%2BRachets%2BLtd%26to%3DJacky%2BTurner%26to_uuid%3Df62dfda8-73b0-4e0e-97cf-1b06327a6712%26to_ru_id%3Dc614e64e-d981-4eba-b016-d9822f09a4fb">Create message link</a>
```
gemmai95: I don't know that the parameters need to be there. Are they there for a reason?

response_operations_ui/views/messages.py
```diff
@@ -1,16 +1,93 @@
...
-                        template_folder='templates')
+
+@messages_bp.route('/create-message', methods=['GET', 'POST'])
+@login_required
+def create_message():
+    form = SecureMessageForm(request.form)
+    breadcrumbs = _build_create_message_breadcrumbs()
+
+    if form.validate_on_submit():
+        # Hard coded id's until we can fetch them from UAA or passed by Reporting Units page
```
gemmai95: Should this also be a todo?

response_operations_ui/views/reporting_units.py
```diff
@@ -26,7 +27,17 @@ def view_reporting_unit(ru_ref):
             "title": f"{ru_ref}"
         }
     ]
-    return render_template('reporting-unit.html', ru=reporting_unit, breadcrumbs=breadcrumbs)
+
+    # Details for mock create message link
+    ru_details = {'survey': 'BRES 2017',
+                  'ru_ref': ru_ref,
+                  'business': 'Bolts & Rachets Ltd',
+                  'to': 'Jacky Turner',
```
gemmai95: Should the respondent be hardcoded?

AdamHawtin: For the moment yes, this is information we need to be passed from pages that don't exist yet.

response_operations_ui/templates/home.html
AdamHawtin: This page should normally only be accessed through a page that passes it these parameters, it doesn't work without them

response_operations_ui/controllers/message_controllers.py
AdamHawtin: We have tasks on the board to make these changes so I would argue a comment is sufficient.

response_operations_ui/controllers/message_controllers.py
AdamHawtin: Until a new auth server is in place this is the only way to work with secure message

response_operations_ui/forms.py
```diff
@@ -2,9 +2,8 @@
 
 from flask_wtf import FlaskForm
 from structlog import wrap_logger
-from wtforms import PasswordField, StringField, SubmitField
-from wtforms.validators import InputRequired
-
+from wtforms import PasswordField, StringField, SubmitField, TextAreaField, Label, HiddenField
```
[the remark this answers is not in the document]
JamesGardiner: Still needs alphabetising.

response_operations_ui/forms.py
```diff
@@ -15,6 +14,26 @@ class LoginForm(FlaskForm):
     submit = SubmitField('Sign in')
 
 
+class SecureMessageForm(FlaskForm):
+    send = SubmitField(label='Send')
+    subject = StringField('Subject',
+                          validators=[InputRequired(message="Please enter a subject"),
+                                      Length(max=96, message="Please enter a subject less than 100 characters")])
```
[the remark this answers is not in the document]
JamesGardiner: We've covered this outside of this PR

response_operations_ui/views/messages.py
```diff
@@ -1,15 +1,93 @@
+import json
 import logging
+from urllib.parse import parse_qs
 
-from flask import Blueprint, render_template, request
+from flask import Blueprint, render_template, request, redirect, url_for, g
```
JamesGardiner: Same here

response_operations_ui/forms.py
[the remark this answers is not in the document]
AdamHawtin: I don't like this but I don't see how else to do it. This matches what is done on frontstage.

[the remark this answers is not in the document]
JamesGardiner: Ok - this has been covered outside of this PR

config.py
```diff
@@ -7,14 +7,15 @@ class Config(object):
     PORT = os.getenv('PORT', 8085)
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
     USE_SESSION_FOR_NEXT = True
-
-    BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL', 'http://localhost:8001/backstage-api')
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
+    BACKSTAGE_BASE_URL = os.getenv('BACKSTAGE_BASE_URL')
+    BACKSTAGE_API_SEND = '/v1/secure-message/send-message'
```
insacuri: I still don't like this.  Can anyone else weigh in? @JamesGardiner @warrenbailey?  The endpoint changes so infrequently that it's just adding config for the heck of it.  If we were to go in this direction we'd have to do it for every single endpoint that we point to in backstage which would massively bloat the config file for no reason

JamesGardiner: Agreed; we have not long completed an exercise to reduce our config overhead. Config should be exclusively for things that change between environments or things that change frequently.

atrozzo: I disagree totaly. Everybody knows that String is Evil , and when you test something , you do not want to test a String but a configuration . Why do you want to change Strings over a Constant based on your running porfile .

config.py
```diff
@@ -7,14 +7,14 @@ class Config(object):
     PORT = os.getenv('PORT', 8085)
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
     USE_SESSION_FOR_NEXT = True
-
-    BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL', 'http://localhost:8001/backstage-api')
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
+    BACKSTAGE_BASE_URL = os.getenv('BACKSTAGE_BASE_URL')
```
benjefferies: We should be cautious when changing the name of environmental variables. We will either need to change this back or propagate these changes to all of the jenkins deployments jobs; dev,ci,int,sit,pre-prod,prod and the docker repo: https://github.com/ONSdigital/ras-rm-docker-dev

response_operations_ui/views/messages.py
andrewmil: We can add a TBL to update this across the other endpoints. We want the core view code to be as human readable as possible

andrewmil:
The code here is good and works fine by the looks of things but we need to have unit tests for any new functionality we're putting in. There's too large a drop in code coverage here

PR #53, opened 2018-02-12
Sm111 graceful exit

insacuri:


response_operations_ui/views/messages.py
```diff
@@ -22,10 +22,13 @@ def view_messages():
         'label': request.args.get('label'),
         'survey': request.args.get('survey')
     }
-    messages = message_controllers.get_message_list(params)
-    refined_messages = [_refine(msg) for msg in messages]
     breadcrumbs = [{"title": "Messages"}]
-    return render_template("messages.html", breadcrumbs=breadcrumbs, messages=refined_messages)
+    messages = message_controllers.get_message_list(params)
```
JamesGardiner: I would make `get_message_list` throw a `NoMessages` that subclasses `Exception` error or a `KeyError` and catch that here, taking away the if statements.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -31,9 +31,5 @@ def get_message_list(params):
...
         return messages
     except KeyError:
-        # TODO: Look to fail more gracefully.  Returning an empty list will display
-        # 'you have no mail' on the screen, which isn't accurate as it's more accurately
-        # 'we don't know if you have messages, try again later'.  This should error for the
-        # user but not give a server error page.
-        logger.exception("Response was successful but didn't contain messages element")
-        return []
+        logger.exception("Response was successful but didn't contain a 'messages' key")
+        raise NoMessages
```
JamesGardiner: Might be worth calling this NoMessagesError

response_operations_ui/exceptions/exceptions.py
```diff
@@ -3,3 +3,7 @@ class ApiError(Exception):
     def __init__(self, response):
         self.url = response.url
         self.status_code = response.status_code
+
+
+class NoMessages(Exception):
+    pass
```
JamesGardiner: Good practice is to put a simple docstring here describing when the exception should be raised.

PR #55, opened 2018-02-15
add respondents for ru per survey

tej99:
As an: internal user
I need: to be able to view respondent data for a Reporting Unit for a specific survey
So that: I understand the relationships with respondents for the Reporting Unit across surveys

Dependencies:
https://github.com/ONSdigital/ras-backstage/pull/53

### DIFF
--- a/response_operations_ui/views/reporting_units.py
+++ b/response_operations_ui/views/reporting_units.py
@@ -29,6 +29,10 @@ def view_reporting_unit(ru_ref):
             collection_exercise['responseStatus'] = map_ce_response_status(collection_exercise['responseStatus'])
             collection_exercise['companyRegion'] = map_region(collection_exercise['companyRegion'])
 
+        for respondent in survey['respondents']:
+            respondent['status'] = map_respondent_status(respondent['status'])
+            respondent['enrolmentStatus'] = map_respondent_status(respondent['enrolmentStatus'])
+
     breadcrumbs = [
         {
             "title": "Reporting units",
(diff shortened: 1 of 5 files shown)


response_operations_ui/views/reporting_units.py
```diff
@@ -29,6 +29,10 @@ def view_reporting_unit(ru_ref):
             collection_exercise['responseStatus'] = map_ce_response_status(collection_exercise['responseStatus'])
             collection_exercise['companyRegion'] = map_region(collection_exercise['companyRegion'])
 
+        for respondent in survey['respondents']:
+            respondent['status'] = map_respondent_status(respondent['status'])
+            respondent['enrolmentStatus'] = map_respondent_status(respondent['enrolmentStatus'])
```
andrewmil: Do we need to map both statuses? What's the difference?

response_operations_ui/views/reporting_units.py
```diff
@@ -76,3 +80,20 @@ def map_region(region):
...
+    elif status == "PENDING":
+        status = "Pending"
+    elif status == "ENABLED":
+        status = "Enabled"
+    elif status == "DISABLED":
+        status = "Disabled"
+    elif status == "SUSPENDED":
+        status = "Suspended"
+
+    return status
```
andrewmil: We could just have a return in each if statement rather than setting a variable each time

response_operations_ui/views/reporting_units.py
andrewmil: And if they're different surely they should have separate mapping functions

tej99: One is the account status and the other is the enrolment status. Could separate it out into two functions but i didn't see any point in doing it.

tej99: CREATED, ACTIVE, SUSPENDED  are for account
PENDING,  ENABLED, DISABLED,  SUSPENDED are for enrolment

tej99:
[US046: INT View Enrolled Respondents for a Survey for an RU[Med]](https://trello.com/c/DkC0zhwn/41-us046-int-view-enrolled-respondents-for-a-survey-for-an-rumed)

response_operations_ui/views/reporting_units.py
```diff
@@ -77,3 +81,18 @@ def map_region(region):
         region = "GB"
 
     return region
+
+
+def map_respondent_status(status):
+    if status == "CREATED":
```
warrenbailey: All of this can be replaced with 

return status.title()

PR #56, opened 2018-02-15
Eq cdn

andrewmil:
We now link to the ons global design patterns stylesheet and have replaced the scss files in the repo with css files directly from the prototype in dropbox.

Because of this we no longer need to parse scss into css


response_operations_ui/static/scss/components/search.scss
```diff
@@ -1,9 +0,0 @@
-.ru-search-box {
```
tej99: This is needed

response_operations_ui/templates/layouts/base.html
```diff
@@ -19,7 +20,7 @@
     <link href="/static/images/favicon.ico" rel="Shortcut Icon" />
   </head>
 
-  <body>
+  <body style="position: relative; height: auto; min-height: 100%;">
```
insacuri: Is this right?  I've always been told that inline styling is bad

response_operations_ui/static/css/components/nav.css
```diff
@@ -34,15 +42,25 @@
...
+/* Using BEM         */
+
+.nav-tabs{
     list-style: none;
     padding: 0;
     border-bottom: 1px solid #666;
 }
 
-.navigation-tabs__tab, .navigation-tabs__tab--active{
+.nav-tabs__tab, .nav-tabs__tab--active{
```
insacuri: This will break the styling on the messages tabs.  Either change the name back or change the class on the page.

insacuri:
Is it correct to change the scss folder to css?  I thought those files can use scss features... but I'm not well versed in it so I might be wrong.

response_operations_ui/static/css/components/nav.css
```diff
@@ -63,30 +81,53 @@
...
 }
 
-.navigation-tabs__tab.create{
+
+
+.nav-tabs__tab.create{
   float: right;
 }
+
+.tab-label.create-new-message{
```
insacuri: With these new ones, I don't think they're being used anywhere on the site at the moment.  Would it be better to add them down the line to keep these files as clean as possible?

response_operations_ui/static/css/components/footer.css
```diff
@@ -0,0 +1,75 @@
+
+/* Styles for the footer */
+
+
+
+footer{
```
insacuri: This should be deleted as it does nothing

insacuri:
Aaaalso, take a look at the codacy warnings, some of them are legitimate

response_operations_ui/templates/layouts/base.html
andrewmil: I'll discuss with Nic when he;s back in, this is from his templating

response_operations_ui/static/css/components/nav.css
andrewmil: works on my branch?

response_operations_ui/static/scss/components/search.scss
andrewmil: ru search box is being changed, I discussed these changes with @nicprice on Thursday

andrewmil:
The secure messaging stuff isn't working with these changes, I'll have a chat with @nicprice next time he's in

response_operations_ui/static/css/components/nav.css
andrewmil: The messaging css is all messed up, I'll have to have a chat with @nicprice 
Currently it's all in a separate override.css file from what nic was using but it should really mostly come from ons stylesheets and the rest should be in components. There's no rush for this PR atm but once that issue has been fixed we'll get it in

andrewmil:
There are a couple of small UI bugs introduced by this, but after a discussion with Nic Price we've decided to go with this version and refine it in a later PR

PR #58, opened 2018-02-16
Us018 load variable event dates

emilioward:
This PR adds in the ability for the user to view variable dates for a collection exercise (`employment`, `ref p start`, `ref p end`, `reminders`). 

You'll need to load in the new event dates from the RM tools to be able to view the dates.

Best to look at the qbs survey to view the dates as that all all the new variable dates added.


emilioward:
[US018: INT Load variable event dates for a collection exercise [M]](https://trello.com/c/n4uTwf2f/43-us018-int-load-variable-event-dates-for-a-collection-exercise-m)

PR #60, opened 2018-02-19
View active IAC code

tej99:
GIVEN
The user understands which surveys the RU Ref has been selected for 
WHEN
The surveys are displayed back to the user
THEN
For each survey, the user is displayed the most recent Enrolment Code issued for that RU for that Survey
Enrollment Code displayed is based on the issued code at notification which hasn't yet been used
Once code is used and respondent is fully active and enabled suppress the code
Enrollment Code displayed is last code unused using 'Generate Code'

Dependencies: https://github.com/ONSdigital/ras-backstage/pull/55

### DIFF
--- a/response_operations_ui/templates/reporting-unit.html
+++ b/response_operations_ui/templates/reporting-unit.html
@@ -66,6 +66,12 @@
 
       <h4 class="neptune">Respondents</h4>
 
+      <div class="ru-survey-enrolment-code u-pb-s">
+        {% if survey.activeIacCode %}
+        Unused enrolment code: {{survey.activeIacCode}}
+        {% endif %}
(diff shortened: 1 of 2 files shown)


response_operations_ui/templates/reporting-unit.html
```diff
@@ -66,6 +66,12 @@ <h4 class="neptune">Collection exercises</h4>
 
       <h4 class="neptune">Respondents</h4>
 
+      <div class="ru-survey-enrolment-code u-pb-s">
+        {% if survey.activeIacCode %}
```
andrewmil: This probably shouldnt surround the whole div so we dont have unnecessary whitespace

PR #68, opened 2018-02-22
Us029 select eq ci

tej99:
Scenario: The user is able to select the eQ collection instruments to load into the collection exercise
    Given the internal user navigates to the collection exercise details page for QBS 1803
    When the user selects the collection instrument
    Then the collection instrument is added into the collection exercise


tej99:
[US029: INT Select/Load EQ Collection Instruments [XXL]](https://trello.com/c/QharabAR/42-us029-int-select-load-eq-collection-instruments-xxl)

PR #69, opened 2018-02-23
Flask Server side sessions with REDIS

warrenbailey:
Implementing server side sessions with REDIS. This is too allow us to store the UAA token in redis for later use.

### DIFF
--- a/config.py
+++ b/config.py
@@ -9,12 +9,16 @@ class Config(object):
     USE_SESSION_FOR_NEXT = True
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
+    SESSION_TYPE = "redis"
 
 
 class DevelopmentConfig(Config):
     DEBUG = os.getenv('DEBUG', True)
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'DEBUG')
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL', 'http://localhost:8001/backstage-api')
+    REDIS_HOST = os.getenv('REDIS_HOST', "localhost")
+    REDIS_PORT = os.getenv('REDIS_PORT', 7379)
+    REDIS_DB = os.getenv('REDIS_DB', 3)
 
 
 class TestingConfig(DevelopmentConfig):
--- a/config.py
+++ b/config.py
@@ -11,7 +11,7 @@ class Config(object):
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
     SESSION_TYPE = "redis"
     PERMANENT_SESSION_LIFETIME = os.getenv('PERMANENT_SESSION_LIFETIME', 43200)
-    REDIS_SERVICE = os.getenv('REDIS_SERVICE', "res-ops-redis")
+    REDIS_SERVICE = os.getenv('REDIS_SERVICE')
     REDIS_HOST = os.getenv('REDIS_HOST')
     REDIS_PORT = os.getenv('REDIS_PORT')
     REDIS_DB = os.getenv('REDIS_DB', 0)
--- a/config.py
+++ b/config.py
@@ -10,6 +10,10 @@ class Config(object):
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
     SESSION_TYPE = "redis"
+    SESSION_PERMANENT = False
+    REDIS_HOST = os.getenv('REDIS_HOST')
+    REDIS_PORT = os.getenv('REDIS_PORT')
+    REDIS_DB = os.getenv('REDIS_DB', 3)
 
 
 class DevelopmentConfig(Config):
--- a/config.py
+++ b/config.py
@@ -30,3 +30,5 @@ class TestingConfig(DevelopmentConfig):
     TESTING = True
     LOGIN_DISABLED = True
     WTF_CSRF_ENABLED = False
+    SESSION_TYPE = "filesystem"
+    SESSION_PERMANENT = False
--- a/config.py
+++ b/config.py
@@ -10,7 +10,7 @@ class Config(object):
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
     SESSION_TYPE = "redis"
-    SESSION_PERMANENT = False
+    PERMANENT_SESSION_LIFETIME = os.getenv('PERMANENT_SESSION_LIFETIME', 43200)
     REDIS_HOST = os.getenv('REDIS_HOST')
     REDIS_PORT = os.getenv('REDIS_PORT')
     REDIS_DB = os.getenv('REDIS_DB', 3)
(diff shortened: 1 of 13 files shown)


warrenbailey:
@AndrewTorrance there are no v2 endpoints on backstage api. The v2 endpoints are only on secure-messaging so the endpoints are technically correct in this PR. I do wonder if we should upgrade the backstage endpoints but that's definitely out of scope of this PR.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -60,6 +60,6 @@ def _get_url():
 
 
 def _get_jwt():
```
insacuri: should this method name be changed?

response_operations_ui/__init__.py
```diff
@@ -26,17 +31,31 @@
...
+        app.config['REDIS_PORT'] = cf.redis.credentials['port']
+
+    redis = redis.StrictRedis(host=app.config['REDIS_HOST'],
+                              port=app.config['REDIS_PORT'],
+                              db=app.config['REDIS_DB'])
+
+    # wrap in the flask server side session manager and back it by redis
+    app.config['SESSION_REDIS'] = redis
+
+Session(app)
```
ajmaddaford: By wrapping the app with Session all endpoints will create a session, including `/info` - if this endpoint is hit by an automated check e.g. to check if the service is still up, then we could end up with a lot of sessions being created.

config.py
```diff
@@ -9,16 +9,26 @@ class Config(object):
     USE_SESSION_FOR_NEXT = True
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
+    SESSION_TYPE = "redis"
+    PERMANENT_SESSION_LIFETIME = os.getenv('PERMANENT_SESSION_LIFETIME', 43200)
+    REDIS_HOST = os.getenv('REDIS_HOST')
+    REDIS_PORT = os.getenv('REDIS_PORT')
+    REDIS_DB = os.getenv('REDIS_DB', 3)
```
ajmaddaford: I think we should assume there is a Redis instance for backstage sessions, or for the response-operations-ui service rather than for all of RAS and default this to 0. If the app is using a shared Redis instance, this parameter can be set in the environment.

warrenbailey: Agreed, I'll change and make the service name configurable in the cloud_foundry.py file

response_operations_ui/__init__.py
warrenbailey: Good spot, will investigate remove certain URLs

warrenbailey: It's difficult to remove urls from the flask session as it's app based. However i've cleared the session if you hit the info endpoint, that should stop them building up unnecessarily in REDIS. The only downside being if the user manual hits /info it'll sign them out. Though they shouldn't be doing that anyway.

response_operations_ui/__init__.py
```diff
@@ -26,19 +31,42 @@
...
 
 @login_manager.user_loader
 def user_loader(user_id):
-    # TODO This will need to be replaced with a call to reddis to get the token
-    # as we can't leave that sort of information in the cookie.  Will be implemented
-    # down the line once uaa is sorted out.
     return User(user_id)
 
 
+@app.after_request
```
ajmaddaford: Can the after request be done on the info blueprint, removing the need for the info global?

config.py
```diff
@@ -9,16 +9,27 @@ class Config(object):
     USE_SESSION_FOR_NEXT = True
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
+    SESSION_TYPE = "redis"
+    PERMANENT_SESSION_LIFETIME = os.getenv('PERMANENT_SESSION_LIFETIME', 43200)
+    REDIS_HOST = os.getenv('REDIS_SERVICE')
```
ajmaddaford: Should this be REDIS_SERVICE rather than REDIS_HOST?

response_operations_ui/__init__.py
ajmaddaford: I think this is fine for now.

config.py
ajmaddaford: Also, should we default this to something so it doesn't need to be set for every environment?

warrenbailey: Yes and yes

response_operations_ui/__init__.py
JamesGardiner: https://github.com/pallets/flask/blob/d72612d11e4f4e5d682c8986acb42ae5e8e97e4a/flask/blueprints.py#L327

config.py
JamesGardiner: The RAS/RM standard for now has been to *not* use defaults in base config classes, so that defaults aren't deployed to prod by mistake. Not saying I agree with that approach, but it's certainly the expected behaviour of services.

response_operations_ui/__init__.py
warrenbailey: Yes good idea

response_operations_ui/__init__.py
```diff
@@ -1,12 +1,17 @@
 import os
+import redis
```
JamesGardiner: Redis isn't stdlib

response_operations_ui/cloud/cloudfoundry.py
```diff
@@ -0,0 +1,16 @@
...
+from flask import current_app
+
+
+class ONSCloudFoundry(object):
+
+    def __init__(self):
+        self._cf_env = cfenv.AppEnv()
+
+    @property
+    def detected(self):
```
JamesGardiner: An ONSCloudFoundry object could probably just evaluate to True if the environment is detected, rather than passing back the app attribute:

```Python
def __bool__(self):
    return True if self._cf_env.app
```

JamesGardiner: Then do

```Python
cf = ONSCloudfoundry()
if cf:
```

response_operations_ui/controllers/message_controllers.py
```diff
@@ -1,8 +1,8 @@
 import logging
 
-import jwt
 import requests
-from flask import current_app
+from flask import current_app, session
```
JamesGardiner: alphabetised imports

response_operations_ui/views/sign_in.py
```diff
@@ -29,7 +29,9 @@ def sign_in():
         response_json = sign_in_controller.sign_in(sign_in_data)
 
         if 'token' in response_json:
```
JamesGardiner:
```Python
try:
    user = User(response_json.get('user_id'))
	# store the token in the session (it's server side and stored in redis)
    session['token'] = response_json['token']
    login_user(user)	         
    if 'next' in session:
        return redirect(session['next'])
    return redirect(url_for('home_bp.home'))
except KeyError:
    return render_template('sign_in.html', form=form)
```

tests/views/test_message.py
```diff
@@ -18,6 +19,13 @@ def setUp(self):
         self.app = app.test_client()
         self.app.testing = True
         app.config.from_object(TestingConfig)
+        self.before()
+
+    @requests_mock.mock()
```
JamesGardiner: Are we going to use requests_mock or responses going forward? My vote goes for responses, it seems to be much more fully featured and comes from Sentry, so should be decent quality: https://github.com/getsentry/responses

JamesGardiner:
Yeah agree with @insacuri on the above - might be a wider discussion and treat it as a small tech debt task

tests/views/test_message.py
warrenbailey: We'll discuss in the tech session on Monday then. For now I've followed the convention in the rest of the code.

response_operations_ui/controllers/message_controllers.py
JamesGardiner: Ok - we should probably agree a standard so we can get it in sync across the teams linters

warrenbailey: Agreed

warrenbailey:
Rebasing to master

PR #71, opened 2018-02-26
Sm101 collection case id fix

AdamHawtin:
Adds the case ID to the data expected on the reporting units page which is then passed to the create message page through the create message button. This field is required by secure-message to log case events for message exchanges.

Story page:
https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/223281382/SM101+Internal+User+to+Send+Message


AdamHawtin:
It turns out the current UI does not allow a user to explicitly choose a collection case so these changes are being put back on the backlog until we can get the correct case ID.

PR #77, opened 2018-03-02
Fix bug in secure message inbox

AdamHawtin:
There is a bug in the messages inbox where messages sent from the 'create message' option in frontstage didn't contain the name fields it expected, causing a 500 error on loading the page.

This fixes it by taking the 'msg_to' field if the message does not have names.

### DIFF
--- a/response_operations_ui/views/messages.py
+++ b/response_operations_ui/views/messages.py
@@ -130,6 +130,14 @@ def _refine(message):
         'business_name': message.get('@ru_id').get('name'),
         'subject': message.get('subject'),
         'from': message.get('msg_from'),
-        'to': message.get('@msg_to')[0].get('firstName') + ' ' + message.get('@msg_to')[0].get('lastName'),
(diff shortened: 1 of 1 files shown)


response_operations_ui/views/messages.py
```diff
@@ -130,6 +130,14 @@ def _refine(message):
...
         'from': message.get('msg_from'),
-        'to': message.get('@msg_to')[0].get('firstName') + ' ' + message.get('@msg_to')[0].get('lastName'),
+        'to': _get_name_from_message(message),
         'sent_date': message.get('sent_date').split(".")[0]
     }
+
+
+def _get_name_from_message(message):
+    try:
+        name = message.get('@msg_to')[0].get('firstName') + ' ' + message.get('@msg_to')[0].get('lastName')
```
JamesGardiner: Nice use of the index error here.

warrenbailey:
As discussed in stand up we'll do both

PR #79, opened 2018-03-02
US028 Add error panel to Collection Exercise details failed validation

jcox-dev:
Also removes list element to remove indentation.

image


jcox-dev:
[US028: Set and View 'Ready for Live' State of a CE [Large]](https://trello.com/c/X6SD8Tmu/28-us028-set-and-view-ready-for-live-state-of-a-ce-large)

jcox-dev:
Changes to the [Collection Exercise service](https://github.com/ONSdigital/rm-collection-exercise-service) incoming (https://github.com/ONSdigital/rm-collection-exercise-service/pull/38) means we will shortly be able to interrogate what caused a FAILEDVALIDATION. 

Some discussion required around number of requests that may be performed to achieve detailed messages on the frontend (whether we return the validation errors in the first CE detail GET JSON response or, upon checking the state hitting the new /errors endpoint before rendering the rest of the page).

jcox-dev:
Error wording may still yet change pending consideration from @nicprice.

response_operations_ui/views/collection_exercise.py
```diff
@@ -38,9 +38,14 @@ def view_collection_exercise(short_name, period, error=None, ci_loaded=False, ex
     ]
 
     ce_state = ce_details['collection_exercise']['state']
-    show_set_live_button = ce_state == 'READY_FOR_REVIEW'
+    show_set_live_button = ce_state in ('READY_FOR_REVIEW', 'FAILEDVALIDATION')
     locked = ce_state in ('LIVE', 'READY_FOR_LIVE', 'EXECUTION_STARTED', 'VALIDATED', 'EXECUTED')
     processing = ce_state in ('EXECUTION_STARTED', 'EXECUTED', 'VALIDATED')
+    validation_failed = ce_state == 'FAILEDVALIDATION'
```
insacuri: 🎨 Super minor, but maybe `ce_state is 'FAILEDVALIDATION'` just so there is less = all over the place ^^

response_operations_ui/views/collection_exercise.py
```diff
@@ -38,9 +38,14 @@ def view_collection_exercise(short_name, period, error=None, ci_loaded=False, ex
...
+
+    validation_errors = ce_details['collection_exercise']['validationErrors']
```
insacuri: Is it possible for ce_details['collection_exercise'] to be empty?  If it is then it can throw a KeyError (..or is it TypeError... I can't remember sorry) and it needs to be handled.

jcox-dev: @insacuri a KeyError would be thrown before reaching this line and handled by the [error handler](https://github.com/ONSdigital/response-operations-ui/blob/81261b366e028354709d4fe9870b3b6d482f216b/response_operations_ui/error_handlers.py#L27).

insacuri: oki dokes ^^  I figured that was the case but I've been seeing a LOT of thing['something']['somethingelse'] and can't stop worrying that on the off chance these exceptions aren't being handled, and if they are, nothing is being logged when it really should be.

jcox-dev: Mm. Symptomatic of passing JSON about from service to service, I suppose.

PR #83, opened 2018-03-06
SM: Display first and last name instead of UUID in from

warrenbailey:
In Secure message display first and last name in the inbox view rather than UUID

### DIFF
--- a/response_operations_ui/views/messages.py
+++ b/response_operations_ui/views/messages.py
@@ -126,7 +126,7 @@ def _refine(message):
         'ru_ref': message.get('ru_id'),
         'business_name': message.get('@ru_id').get('name'),
         'subject': message.get('subject'),
-        'from':_get_from_name(message),
+        'from': _get_from_name(message),
         'to': _get_to_name(message),
         'sent_date': message.get('sent_date').split(".")[0]
     }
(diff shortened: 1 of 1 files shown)


response_operations_ui/views/messages.py
```diff
@@ -126,13 +126,21 @@ def _refine(message):
         'ru_ref': message.get('ru_id'),
         'business_name': message.get('@ru_id').get('name'),
         'subject': message.get('subject'),
-        'from': message.get('msg_from'),
-        'to': _get_name_from_message(message),
+        'from':_get_from_name(message),
```
JamesGardiner: missing some whitespace here

response_operations_ui/views/messages.py
```diff
@@ -126,13 +126,21 @@ def _refine(message):
...
     }
 
 
-def _get_name_from_message(message):
+def _get_from_name(message):
+    try:
+        msg_from = message['@msg_from']
+        return f"{msg_from.get('firstName')} {msg_from.get('lastName')}"
+    except KeyError:
+        return message.get('msg_from')
```
JamesGardiner: return None? More efficient than trying to access the dict object again.

warrenbailey: It's return the uuid (same as the method below)?

JamesGardiner: Ah I missed the '@'

PR #86, opened 2018-03-08
SM111 Update Inbox View to Show Threads

AdamHawtin:
Story Card:
[SM111: INT - View All Inbox Messages](https://trello.com/c/gfaiP90r/60-sm111-int-view-all-inbox-messages-from-threads-13)

As part of the feature to show message conversations the secure message inbox view needs to get threads not messages. This update changes it to hit the backstage [get threads endpoint](https://github.com/ONSdigital/ras-backstage/pull/70) instead of get messages and updates the view to show the correct fields. Also tries to make the inbox more resilient to malformed data, previously it had issues with the page refusing to load completely if a single message was missing some expected fields. Now it should always load and display 'Unavailable' in any fields that are missing.

PR for updates to acceptance tests for this update:
[SM111 Internal Inbox Now Shows Threads #70](https://github.com/ONSdigital/rasrm-acceptance-tests/pull/70)


response_operations_ui/controllers/survey_controllers.py
```diff
@@ -30,3 +30,25 @@ def get_survey(short_name):
 
     logger.debug('Successfully retrieved survey', short_name=short_name)
     return response.json()
+
+
+def get_surveys_dict():
+    # TODO get and cache this dictionary at app startup
+    logger.debug('Retrieving surveys list')
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/survey/surveys'
```
AndrewTorrance: This code is not DRY , we repeat the code in get_service_list and get_service_dict . We should only be doing this in one place . This will also make it easier when we add caching of this list which I am assuming is a follow up ticket ?

response_operations_ui/views/messages.py
```diff
@@ -123,12 +125,12 @@ def view_messages():
...
 def _refine(message):
     return {
-        'ru_ref': message.get('ru_id'),
-        'business_name': message.get('@ru_id').get('name'),
+        'ru_ref': _get_ru_ref_from_message(message),
+        'business_name': _get_business_name_from_message(message),
         'subject': message.get('subject'),
         'from': _get_from_name(message),
         'to': _get_to_name(message),
-        'sent_date': message.get('sent_date').split(".")[0]
```
AndrewTorrance: given we have switched to the threads view , then its not a given that the latest message retrieved has been sent . It could for example be a draft in which case it would have a DRAFT_SAVED event.

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -30,3 +30,17 @@ def get_survey(short_name):
...
+    surveys_list = get_surveys_list()
+    return {survey['id']: survey for survey in surveys_list}
+
+
+def get_survey_short_name_by_id(survey_id):
+    try:
+        return get_surveys_dict()[survey_id]['shortName']
+    except KeyError:
+        logger.exception("failed to resolve survey short name", survey_id=survey_id)
+        return 'Unavailable'
```
JamesGardiner: In the case of a KeyError would this result in the display of "Unavailable Team" in the UI. You might want to speak to UI about what to display in this instance.

response_operations_ui/views/messages.py
```diff
@@ -137,12 +139,41 @@ def _get_from_name(message):
...
 
 
 def _get_to_name(message):
     try:
-        name = message.get('@msg_to')[0].get('firstName') + ' ' + message.get('@msg_to')[0].get('lastName')
-    except IndexError:
-        name = message.get('msg_to')[0]
-    return name
+        if message.get('msg_to')[0] == 'GROUP':
+            return f"{get_survey_short_name_by_id(message.get('survey'))} Team"
```
dcdarrell9: Do we want to add a log here to know if this is getting called?

response_operations_ui/views/messages.py
```diff
@@ -137,12 +139,41 @@ def _get_from_name(message):
...
+        return f"{message.get('@msg_to')[0].get('firstName')} {message.get('@msg_to')[0].get('lastName')}"
+    except (IndexError, TypeError):
+        logger.exception("Failed to retrieve message to name ", message_id=message['msg_id'])
+        return 'Unavailable'
```
dcdarrell9: Do we know what will go here in the future?

response_operations_ui/views/messages.py
```diff
@@ -114,7 +116,7 @@ def view_messages():
     }
     breadcrumbs = [{"title": "Messages"}]
     try:
-        messages = message_controllers.get_message_list(params)
+        messages = message_controllers.get_thread_list(params)
```
JamesGardiner: Don't bother assigning this. You could just do:

```Python
messages = [_refine(msg) for msg in message_controllers.get_thread_list(params)]
```

response_operations_ui/views/messages.py
AdamHawtin: I don't think so, this is normal behaviour for if a respondent creates a message. If the survey service is called logs are generated then.

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -30,3 +30,15 @@ def get_survey(short_name):
...
+def get_survey_short_name_by_id(survey_id):
+    # TODO cache the surveys dictionary at app start up
+    try:
+        surveys_dict = {survey['id']: survey for survey in get_surveys_list()}
+        return surveys_dict[survey_id]['shortName']
+    except ApiError:
+        return 'Unavailable'
+    except KeyError:
+        logger.exception("failed to resolve survey short name", survey_id=survey_id)
+        return 'Unavailable'
```
JamesGardiner: I think this dismissed my last statement - is this correct in the UI when returning unavailable?

response_operations_ui/views/messages.py
AdamHawtin: When drafts exist we'll have to have some logic to deal with them, but I don't think the UI behaviour for that has been decided yet.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -11,10 +11,10 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-def get_message_list(params):
```
dcdarrell9: Will we ever need to use get_message_list again? (I'm guessing no)

tests/views/test_message.py
```diff
@@ -27,38 +29,126 @@ def before(self, mock_request=None):
...
 
+    surveys_list_json = [
+        {
+            "id": "f235e99c-8edf-489a-9c72-6cabe6c387fc",
+            "shortName": "QBS",
+            "longName": "Quarterly Business Survey"
+        }
+    ]
+
+    # Test showing that the threads list loads into the website and displays User, business name and subject
```
dcdarrell9: Probably don't need this

response_operations_ui/controllers/message_controllers.py
AdamHawtin: Not as far as I know, the only use of it became the threads views.

PR #87, opened 2018-03-08
Sm124 conversation view

ryangrundy7:
As an: Internal User
I need: To be able to see all of the previous messages in a conversation
So that: I have the full context of a message

https://digitaleq.atlassian.net/wiki/x/oIBiDQ


response_operations_ui/controllers/message_controllers.py
```diff
@@ -11,6 +11,25 @@
...
+
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/threads/{thread_id}'
+
+    response = requests.get(url, headers={'Authorization': _get_jwt()})
+
+    try:
+        response.raise_for_status()
+    except HTTPError:
+        logger.exception("Conversation retrieval failed")
+        raise ApiError(response)
```
JamesGardiner: Worth also catching RequestException here, which will deal with issues such as connection errors.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -11,6 +11,25 @@
...
+
+    logger.debug("Retrieval successful")
+    conversation = response.json()
+
+    return conversation
```
JamesGardiner: Don't assign response.json to conversation as it's never used apart from in the return statement.

response_operations_ui/views/messages.py
```diff
@@ -146,3 +163,20 @@ def _get_to_name(message):
...
+        'username': message.get('@msg_from').get('emailAddress'),
+        # TODO use survey ref instead of survey id
+        'survey': message.get('survey'),
+        'ruref': message.get('@ru_id').get('sampleUnitRef'),
+        'business': message.get('@ru_id').get('name'),
+        'to': _get_to_name(message),
+        'sentDate': message.get('sent_date'),
+        'internal': message.get('from_internal')
+
+
```
JamesGardiner: No blank lines required here

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,52 @@
...
+
+<div class="secure-message-conversation block">
+    <div class="secure-message-sent-message-meta">
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">Survey: </span><span id="survey">{{ messages[0].get('survey') }}</span></div>
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">RU Ref: </span><span id="ruref">    {{ messages[0].get('ruref') }}</span></div>
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">Business: </span><span id="ru">{{ messages[0].get('business') }}</span></div>
+         <div class="secure-message--recipient venus"><span class="sm-form-recipient-label">To: </span><span id="recipient">    {{ messages[0].get('to') }}  </span></div>
+
+    </div>
+    {% for msg in messages %}
```
ajmaddaford: I don't think we should be shortening words, message is more readable than msg and is consistent with other uses e.g. https://github.com/ONSdigital/response-operations-ui/blob/3db04ba8f4c434d30c4ad88b0bb36a20ece144c2/response_operations_ui/templates/messages.html#L28

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,52 @@
...
+    {% if msg.get('internal') == True %}
+    <div id="smC1M1" class="secure-message-sent-message">
+        <div class="secure-message-sent-message-meta sm-from-ons">
+    {% else %}
+    <div id="smC1M2" class="secure-message-sent-message">
```
ajmaddaford: - What do these id's mean on the divs?
- Why do the divs have different ids? Can they be the same and just vary the class on the inner div?

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,52 @@
...
+        <div class="secure-message-sent-message-meta sm-from-respondent">
+    {% endif %}
+            <span class="secure-message-sent-message-meta-from venus ">
+                <span class="tempHack">From:</span>
```
ajmaddaford: tempHack?

response_operations_ui/views/messages.py
```diff
@@ -121,9 +121,34 @@ def view_messages():
         return render_template("messages.html", breadcrumbs=breadcrumbs, response_error=True)
 
 
+@messages_bp.route('/threads/<thread_id>', methods=['GET'])
+@login_required
+def view_conversation(thread_id):
+    try:
+        messages = [_thread_refine(message) for message in message_controllers.get_conversation(thread_id)['messages']]
+    except KeyError as ex:
+        # TODO change exception message
```
ajmaddaford: Why TODO?

response_operations_ui/views/messages.py
```diff
@@ -121,9 +121,34 @@ def view_messages():
         return render_template("messages.html", breadcrumbs=breadcrumbs, response_error=True)
 
 
+@messages_bp.route('/threads/<thread_id>', methods=['GET'])
+@login_required
+def view_conversation(thread_id):
+    try:
+        messages = [_thread_refine(message) for message in message_controllers.get_conversation(thread_id)['messages']]
```
ajmaddaford: There's quite a lot going on in this line, I think it would be clearer with:
```
messages = message_controllers.get_conversation(thread_id)
refined_messages = [_thread_refine(message) for message in messages]
```

response_operations_ui/views/messages.py
```diff
@@ -146,3 +171,18 @@ def _get_to_name(message):
...
+        'subject': message.get('subject'),
+        'username': message.get('@msg_from').get('emailAddress'),
+        # TODO use survey ref instead of survey id
+        'survey': message.get('survey'),
+        'ruref': message.get('@ru_id').get('sampleUnitRef'),
+        'business': message.get('@ru_id').get('name'),
+        'to': _get_to_name(message),
+        'sentDate': message.get('sent_date'),
+        'internal': message.get('from_internal')
+    }
```
ajmaddaford: Some of this functionality duplicates the `_refine` method, but with some subtle differences:
- `business_name` has become `business`
- `ru_ref` is now `ruref` and is `@ru_id. business` rather than `ru_id`

We aim to make these consistent, possibly sharing the common mappings.

Also, `sentDate` should be `sent_date`.

response_operations_ui/templates/conversation-view.html
ryangrundy7: tempHack was used on the UX guys template which I used as a template. This is what it does in the css:
.tempHack{
  width: 2.5rem;
  display: inline-block;
}

response_operations_ui/views/messages.py
ryangrundy7: Removing the TODO now

response_operations_ui/templates/conversation-view.html
ajmaddaford: Can we update the css from tempHack to 
```
secure-message-sent-message-meta-from > span {
  width: 2.5rem;
  display: inline-block;
}
```

From https://stackoverflow.com/questions/15853899/styling-span-within-a-div

response_operations_ui/templates/conversation-view.html
atrozzo: Divs need a different ID because they represent differt kind of messages. One if for internal another one for external. It s justa convention for allow the Slenium test to diferenciate between them.

response_operations_ui/views/messages.py
atrozzo: I am not sure , but feels more like a personal preference in style. To me the feels more pythonic in single line and it s still really easy to read.

response_operations_ui/templates/conversation-view.html
ajmaddaford: I agree in isolation it's a preference thing, but it is inconsistent with other places in the codebase (where message in messages is used).

response_operations_ui/controllers/message_controllers.py
```diff
@@ -3,14 +3,36 @@
...
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/threads/{thread_id}'
+
+    response = requests.get(url, headers={'Authorization': _get_jwt()})
+
+    try:
+        response.raise_for_status()
+    except (HTTPError, RequestException):
+        logger.exception("Thread retrieval failed ")
+        raise ApiError(response)
+    logger.debug("Thread Retrieval successful")
```
gavinleeedwards: Could we include thread_id in the log messages?  It would be hard to investigate errors without specifics.

response_operations_ui/views/messages.py
ajmaddaford: Indeed, we've agreed to leave as it is and discuss as a team next week.

response_operations_ui/templates/conversation-view.html
ajmaddaford: Thanks @atrozzo 
@ryangrundy7 can we use a common outer div and just have a specific id on the inner div?

response_operations_ui/views/messages.py
```diff
@@ -127,27 +127,28 @@ def view_conversation(thread_id):
...
 
 def _get_messages_subject(messages):
     try:
-        return messages[0].get("subject")
-    except (IndexError, AttributeError):
-        return 'Unavailable'
+        return messages[0]["subject"]
+    except KeyError:
+        logger.exception("Failed to retrieve RU ref from message")
+        return "Unavailable"
```
ajmaddaford: It would be better if we returned None and had the fallback text in the template

response_operations_ui/views/messages.py
```diff
@@ -180,9 +181,9 @@ def _thread_refine(message):
...
         # TODO use survey ref instead of survey id
         'survey': message.get('survey'),
-        'ruref': message.get('@ru_id').get('sampleUnitRef'),
-        'business': message.get('@ru_id').get('name'),
+        'ru_ref': message.get('@ru_id').get('sampleUnitRef'),
+        'business_name': message.get('@ru_id').get('name'),
         'to': _get_to_name(message),
-        'sentDate': message.get('sent_date'),
+        'sent_date': message.get('sent_date'),
         'internal': message.get('from_internal')
```
ajmaddaford: Do we need a separate refine method, or can we just add the extra fields to `_refine`?

response_operations_ui/static/css/override.css
```diff
@@ -778,3 +778,8 @@ textarea.secure-message--message{
   margin-left: -0.75rem;
   margin-right: .2rem;
 }
+
+.secure-message-sent-message-meta-from > span{
+  width: 2.5rem;
+  display: inline-block;
+}
```
insacuri: Why no newline?

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,51 @@
+{% extends "layouts/base.html" %}
+
+{% block page_title %} Create Message | Survey Data Collection {% endblock %}
+
+{% block main %}
+
+
```
insacuri: The spacing in this file seems odd.  Can you tidy it up?

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,51 @@
...
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">Survey: </span><span id="survey">{{ messages[0].get('survey') }}</span></div>
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">RU Ref: </span><span id="ruref">    {{ messages[0].get('ru_ref') }}</span></div>
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">Business: </span><span id="ru">{{ messages[0].get('business_name') }}</span></div>
+         <div class="secure-message--recipient venus"><span class="sm-form-recipient-label">To: </span><span id="recipient">    {{ messages[0].get('to') }}  </span></div>
+
+    </div>
+    {% for message in messages %}
+       <div class="secure-message-sent-message">
+    {% if message.get('internal') == True %}
+        <div id="smC1M1" class="secure-message-sent-message-meta sm-from-ons">
```
insacuri: What's the deal with these ids?  Isn't there a more human readable way to describe this field?

ryangrundy7: @insacuri id's have been changed but they were originally ones the UX guys have used in their templates

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,51 @@
...
+{% block main %}
+
+
+
+<h1 class="saturn" name="page-messages-title">{{ subject }}</h1>
+
+<div class="secure-message-conversation block">
+    <div class="secure-message-sent-message-meta">
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">Survey: </span><span id="survey">{{ messages[0].get('survey') }}</span></div>
+         <div class="secure-message--meta venus"><span class="sm-form-meta-label">RU Ref: </span><span id="ruref">    {{ messages[0].get('ru_ref') }}</span></div>
```
insacuri: Random spacing after ruref id span

response_operations_ui/controllers/message_controllers.py
```diff
@@ -3,14 +3,36 @@
...
+    logger.debug("Retrieving conversation")
+
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/threads/{thread_id}'
+
+    response = requests.get(url, headers={'Authorization': _get_jwt()})
+
+    try:
+        response.raise_for_status()
+    except (HTTPError, RequestException):
+        logger.exception("Thread retrieval failed ", thread_id=thread_id)
```
insacuri: Remove random space after failed

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,51 @@
...
+
+            <p>{{ message.get('body') }}</p>
+
+        </div>
+    </div>
+    {% endfor %}
+        </div>
+        <!--TODO Add in the reply box -->
+    </div>
+    <a id="bottom"></a>
```
insacuri: Is this needed?

response_operations_ui/templates/messages.html
```diff
@@ -106,4 +106,4 @@ <h1 class="saturn u-mt-l">Oops! Something went wrong</h1>
     </div>
   </div>
 </div>
-{% endblock %}
+{% endblock %}
```
insacuri: Add newline

response_operations_ui/templates/conversation-view.html
```diff
@@ -1,51 +1,41 @@
...
-            <span class="secure-message-sent-message-meta-from venus ">
-                <span>From:</span>
-
+            <span class="secure-message-sent-message-meta-from venus">
+                <span id="from">From:</span>
                 {{ message.get('username') }}
             </span>
             <span class="secure-message-sent-message-meta-from mercury">
-                <span>Sent:</span>
+                <span id="sent">Sent:</span>
```
insacuri: Still inside a loop, can't use id unless you do something to make it unique

response_operations_ui/controllers/message_controllers.py
```diff
@@ -3,14 +3,36 @@
...
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/threads/{thread_id}'
+
+    response = requests.get(url, headers={'Authorization': _get_jwt()})
+
+    try:
+        response.raise_for_status()
+    except (HTTPError, RequestException):
+        logger.exception("Thread retrieval failed", thread_id=thread_id)
+        raise ApiError(response)
+    logger.debug("Thread Retrieval successful")
```
insacuri: lower case r in retrieval

response_operations_ui/views/messages.py
```diff
@@ -121,14 +121,46 @@ def view_messages():
...
+
+
 def _refine(message):
     return {
-        'ru_ref': message.get('ru_id'),
+        'ru_ref': message.get('@ru_id').get('sampleUnitRef'),
+        'thread_id': message.get('thread_id'),
         'business_name': message.get('@ru_id').get('name'),
-        'subject': message.get('subject'),
+        'subject': _get_message_subject(message),
```
ajmaddaford: If we don't care about logging an exception, we could replace this method with `message.get('subject')`

response_operations_ui/views/messages.py
```diff
@@ -121,14 +121,46 @@ def view_messages():
         return render_template("messages.html", breadcrumbs=breadcrumbs, response_error=True)
 
 
+@messages_bp.route('/threads/<thread_id>', methods=['GET'])
+@login_required
+def view_conversation(thread_id):
+    try:
+        thread_conversation = message_controllers.get_conversation(thread_id)['messages']
+        refined_thread = [_refine(thread) for thread in reversed(thread_conversation)]
```
ajmaddaford: Isn't thread here actually message, so should be `[_refine(message) for message in reversed(thread_conversation)]`?

warrenbailey: Rather than reverse the conversation, I think we should change the API to return it in the correct order for the UI

response_operations_ui/templates/messages.html
atrozzo: @insacuri not sure if this a cosmetic thing or mandatory checkstyle. I mean it s html/flask page. I would recomend a checkstyle at python level.

response_operations_ui/templates/conversation-view.html
atrozzo: @insacuri as below, not sure if in HTML we should keep this level of accurancy, and we should idstinguish from suggestion and mandatory changes.

response_operations_ui/controllers/message_controllers.py
atrozzo: @gavinleeedwards that is a debug level log, you would not find it in case of error. Do you want it actually to change the logging level in case ?

response_operations_ui/views/messages.py
atrozzo: @warrenbailey I agree, but should we not put down as tech debt external to this story. For the amount messages passed, has not memory or speed downside at the moment.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -3,14 +3,36 @@
...
+        response.raise_for_status()
+    except (HTTPError, RequestException):
+        logger.exception("Thread retrieval failed", thread_id=thread_id)
+        raise ApiError(response)
+    logger.debug("Thread retrieval successful")
+
+    try:
+        return response.json()
+    except JSONDecodeError:
+        logger.exception("the response could not be decoded")
```
dcdarrell9: 🎨 Might be worth logging the thread_id in a few more places, and capital letter here

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,43 @@
...
+{% block page_title %} {{ subject }} | Survey Data Collection {% endblock %}
+
+{% block main %}
+
+
+<h1 class="saturn" name="page-messages-title">{{ subject }}</h1>
+
+<div class="secure-message-conversation block">
+    <div class="secure-message-sent-message-meta">
+         <div class="secure-message--meta venus"><span class="secure-message-conversation-meta__label">Survey: </span><span id="survey">{{ messages[0].get('survey') }}</span></div>
```
dcdarrell9: 🎨 I would probably do it like `messages[0]['survey']` unless we're looking for a default value

response_operations_ui/templates/conversation-view.html
```diff
@@ -0,0 +1,43 @@
+{% extends "layouts/base.html" %}
+
+{% set subject = messages[-1]["subject"] or "No Subject" %}
```
dcdarrell9: Not sure if you need the "No Subject" as I don't think we can send a message without a subject.
🎨 And I think you can set default like `{% set subject = messages[-1].get("subject", "No Subject") %}`

response_operations_ui/views/messages.py
```diff
@@ -121,14 +121,46 @@ def view_messages():
         return render_template("messages.html", breadcrumbs=breadcrumbs, response_error=True)
 
 
+@messages_bp.route('/threads/<thread_id>', methods=['GET'])
+@login_required
+def view_conversation(thread_id):
+    try:
+        thread_conversation = message_controllers.get_conversation(thread_id)['messages']
+        refined_thread = [_refine(message) for message in reversed(thread_conversation)]
```
dcdarrell9: 🎨 Could do `refined_thread = [_refine(message) for message in reversed(message_controllers.get_conversation(thread_id)['messages'])]`

response_operations_ui/controllers/message_controllers.py
gavinleeedwards: I'd say that all errors should be exception level (which this one is) so that they are logged by default and will be picked up by any alerting that we may set up in future.  The log level for the successful retrieval is fine as debug.  Adding the thread_id to both logs would be useful though.

response_operations_ui/templates/conversation-view.html
ajmaddaford: If you assign `messages[0]` to a variable
```
{% set first_message = messages[0] %}
```
Then you should be able to use `{{ first_message.survey }}`

response_operations_ui/templates/conversation-view.html
insacuri: Is there any chance that messages could be None?  If it somehow ends up not being a list then I think this is going to break in a fairly big way and we won't be able to handle errors as we're at the template

response_operations_ui/views/messages.py
ryangrundy7: all in one refine now

response_operations_ui/views/messages.py
ryangrundy7: got fallback text in html

response_operations_ui/views/messages.py
```diff
@@ -127,27 +127,28 @@ def view_conversation(thread_id):
...
 
 
 def _get_messages_subject(messages):
     try:
-        return messages[0].get("subject")
-    except (IndexError, AttributeError):
-        return 'Unavailable'
+        return messages[0]["subject"]
+    except KeyError:
+        logger.exception("Failed to retrieve RU ref from message")
```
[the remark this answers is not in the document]
ryangrundy7: error message changed

response_operations_ui/templates/conversation-view.html
```diff
@@ -6,40 +6,39 @@
...
-    <div id="smC1M1" class="secure-message-sent-message">
-        <div class="secure-message-sent-message-meta sm-from-ons">
+    {% for message in messages %}
+       <div class="secure-message-sent-message">
+    {% if message.get('internal') == True %}
+        <div id="smC1M1" class="secure-message-sent-message-meta sm-from-ons">
     {% else %}
-    <div id="smC1M2" class="secure-message-sent-message">
-        <div class="secure-message-sent-message-meta sm-from-respondent">
+        <div id="smC1M2" class="secure-message-sent-message-meta sm-from-respondent">
```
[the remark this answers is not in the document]
ryangrundy7: id's changed

response_operations_ui/views/messages.py
```diff
@@ -121,14 +121,46 @@ def view_messages():
...
+def view_conversation(thread_id):
+    try:
+        thread_conversation = message_controllers.get_conversation(thread_id)['messages']
+        refined_thread = [_refine(message) for message in reversed(thread_conversation)]
+    except KeyError as ex:
+        logger.exception("A key error occurred")
+        raise ApiError(ex)
+    breadcrumbs = [
+        {"title": "Messages", "link": "/messages"},
+        {"title": refined_thread[0]["subject"]}
```
insacuri: Put this in a try/catch block.  If there were no threads returned, then you'll get an IndexError by trying to get the subject key on an empty list.  Maybe there is a better way then that.... maybe test the length of the refined thread and only make the subject breadcrumb if it's populated?  Might be worth asking the other guys what to do if that happens

response_operations_ui/views/messages.py
ryangrundy7: @JamesGardiner done your suggested changes

response_operations_ui/views/messages.py
ryangrundy7: @dcdarrell9 i've done it this way in the past but i've been suggested to change it so i'll leave it for now

response_operations_ui/views/messages.py
dcdarrell9: We could default it to 'Message' or something similar if subject is unavailable?

insacuri: Yea, that sounds good

PR #88, opened 2018-03-08
Us057 change respondent name and number

rhiGriff:
This PR is dependent on the following PRs:

https://github.com/ONSdigital/ras-backstage/pull/73
https://github.com/ONSdigital/ras-party/pull/105


response_operations_ui/controllers/edit_contact_details_controller.py
```diff
@@ -0,0 +1,43 @@
...
+
+from response_operations_ui import app
+from response_operations_ui.exceptions.exceptions import ApiError
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def edit_contact_details(edit_details_data):
+    logger.debug('Validating contact details')
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/party/update-respondent-details'
```
gemmai95: Is this right?

rhiGriff:
[US057: INT Change Respondent Name & Number [M]](https://trello.com/c/1ZGcBo4W/45-us057-int-change-respondent-name-number-m)

response_operations_ui/templates/edit-contact-details.html
```diff
@@ -0,0 +1,124 @@
...
+                      </span>
+                    </label>
+                  {{ form.telephone(class_="input input--text", id='telephone', required=required, maxlength=254) }}
+                </div>
+
+                <button class="btn btn--primary u-mt-s" id="save-btn">Save changes</button>
+                </form>
+
+                <div>
+                  <a id="cancel-btn" href="{{ referrer }}">Cancel</a>
```
VibhorMehra: Could we please change the href URL to `"{{ url_for('reporting_unit_bp.view_reporting_unit', ru_ref=ru_ref) }}"`? We should not use `referrer` here because if an internal user is navigating to the `Edit contact details` page after successfully editing another respondent's details, the referrer would be set as `/reporting-units/<ru_ref>?edit_details=True`. In this case, the internal user will see `Contact details changed` message even after selecting the `cancel-btn`.

VibhorMehra:
@rhiGriff @emilioward it appears we are not using the validations used on `frontstage` while editing Respondent's Name and Contact number. Do we know why these are not included in the acceptance criteria? This seems a bit strange 😕
resp_edit_contact

rhiGriff:
@VibhorMehra The validation here is what's on the card and is correct

response_operations_ui/views/reporting_units.py
```diff
@@ -57,9 +60,44 @@ def view_reporting_unit(ru_ref):
...
+@reporting_unit_bp.route('/<ru_ref>/edit-contact-details/<respondent_id>', methods=['GET'])
+@login_required
+def view_contact_details(ru_ref, respondent_id):
+    respondent_details = edit_contact_details_controller.get_contact_details(respondent_id)
+
+    referrer = request.referrer
+    form = EditContactDetailsForm(form=request.form, default_values=respondent_details)
+
+    return render_template('edit-contact-details.html', ru_ref=ru_ref, respondent_details=respondent_details,
+                           referrer=referrer, form=form)
```
benjefferies: Why are we added referrer in the model? I can't see where it's used

response_operations_ui/controllers/edit_contact_details_controller.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def edit_contact_details(edit_details_data, respondent_id):
+    logger.debug('Validating contact details')
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/party/update-respondent-details/{respondent_id}'
+
+    response = requests.put(url, json=edit_details_data)
+
+    try:
+        response.raise_for_status()
```
benjefferies: Seems a strange way to do this. I think it would be cleaner if it was something along the lines of
```
response = requests.put(url, json=edit_details_data)
if response.status_code == 401:
    abort(401)
```

benjefferies: Also needs test coverage

response_operations_ui/controllers/edit_contact_details_controller.py
```diff
@@ -0,0 +1,43 @@
...
+
+def get_contact_details(respondent_id):
+    logger.debug('Retrieving contact details by id', id=respondent_id)
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/party/party-details'
+
+    param = {"respondent_party_id": respondent_id}
+    response = requests.get(url, params=param)
+
+    if response.status_code != 200:
+        raise ApiError(response)
```
benjefferies: Missing test coverage

response_operations_ui/static/css/components/spoke.css
```diff
@@ -1,16 +1,29 @@
+/* Spoke styles for hub and spoke patterns and modals */
```
benjefferies: 👍

response_operations_ui/templates/reporting-unit.html
```diff
@@ -2,161 +2,178 @@
...
+              <th scope="col" class="table--header--cell">
+                Status
               </th>
             </tr>
-            </thead>
-            <tbody>
-            {% for respondent in survey.respondents %}
+          </thead>
+          <tbody>
+            {% for ce in survey.collection_exercises[:4] %}
```
benjefferies: Why are we only showing 4?

response_operations_ui/views/reporting_units.py
```diff
@@ -57,9 +60,44 @@ def view_reporting_unit(ru_ref):
...
+              "first_name": request.form.get('first_name'),
+              "last_name": request.form.get('last_name'),
+              "email": request.form.get('email'),
+              "telephone": request.form.get('telephone'),
+              "respondent_id": respondent_id
+          }
+        edit_contact_details_controller.edit_contact_details(edit_details_data, respondent_id)
+
+    else:
+        respondent_details = edit_contact_details_controller.get_contact_details(respondent_id)
```
benjefferies: No test coverage for invalid form

response_operations_ui/templates/reporting-unit.html
rhiGriff: This is in master not my code

benjefferies: Oh didn't notice because of the reformatting

tej99: Requirement is to only show 4 collection exercises.

jwgwalton:
It does seem odd that it doesn't have validation on the email and phone number like the other pages, do we have some code that is reusable?

rhiGriff:
@jwgwalton the validation on frontstage is different to here, I've double checked with the BAs

response_operations_ui/controllers/edit_contact_details_controller.py
```diff
@@ -0,0 +1,37 @@
...
+from structlog import wrap_logger
+
+from response_operations_ui import app
+from response_operations_ui.exceptions.exceptions import ApiError
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def edit_contact_details(edit_details_data, respondent_id):
+    logger.debug('Validating contact details')
```
andrewmil: Probably want to log out `respondent_id` here

andrewmil: Also is `validating` the right word for this message? We are past validating here and we're actually changing them?

response_operations_ui/forms.py
```diff
@@ -40,6 +40,22 @@ class SearchForm(FlaskForm):
     submit = SubmitField('Search')
 
 
+class EditContactDetailsForm(FlaskForm):
+    first_name = StringField('first_name')
```
andrewmil: Should we have some kind of validation in this form?

response_operations_ui/templates/reporting-unit.html
```diff
@@ -2,161 +2,178 @@
...
-    <dt class="item-headlines-term">
-      Reference:
-    </dt>
-    <dd id="RU_REF" class="item-headlines-definition">
-      {{ ru.sampleUnitRef }}
-    </dd>
-  </dl>
 
-  <h2 class="saturn">Surveys</h2>
+{% if edit_details %}
```
andrewmil: This is fine for now but it would probably be better to trigger this with the `info-box` partial as it's doing the same thing just with a different message

response_operations_ui/views/reporting_units.py
```diff
@@ -17,6 +19,7 @@
 @reporting_unit_bp.route('/<ru_ref>', methods=['GET'])
 @login_required
 def view_reporting_unit(ru_ref):
+    edit_details = request.args.get('edit_details')
```
gemmai95: Can this be none?

response_operations_ui/views/reporting_units.py
```diff
@@ -57,9 +60,43 @@ def view_reporting_unit(ru_ref):
...
+          "email": request.form.get('email'),
+          "telephone": request.form.get('telephone'),
+          "respondent_id": respondent_id
+      }
+
+    edit_successfully = edit_contact_details_controller.edit_contact_details(edit_details_data, respondent_id)
+
+    if not edit_successfully:
+        respondent_details = edit_contact_details_controller.get_contact_details(respondent_id)
+        logger.info('Error submitting respondent details', respondent_id=respondent_id)
```
andrewmil: This line + the line under it are not covered in testing

response_operations_ui/forms.py
rhiGriff: This will be part of tech debt @davidmort made this decision, validation is currently done in the html

PR #90, opened 2018-03-09
SM111 cache surveys dictionary

AdamHawtin:
This is a branch from [SM111 Update Inbox View to Show Threads #86](https://github.com/ONSdigital/response-operations-ui/pull/86) so includes all changes from it and shouldn't be merged until that PR is complete.

This update adds lazy caching for the dictionary of surveys. Without, a call is made to the survey service for every message created by an external user every time the inbox is loaded. The dictionary is cached the first time it is needed, then only fetched again if the app restarts or it attempts to get a short name for a survey ID that doesn't currently exist in it's cache.


response_operations_ui/views/messages.py
```diff
@@ -4,10 +4,12 @@
 from flask import Blueprint, flash, g, render_template, request, redirect, url_for
 from flask_login import login_required, current_user
 from structlog import wrap_logger
+import maya
```
JamesGardiner: Alphabetize this 👍

response_operations_ui/views/messages.py
```diff
@@ -4,10 +4,12 @@
...
 
 from response_operations_ui.controllers import message_controllers
 from response_operations_ui.exceptions.exceptions import ApiError, InternalError, NoMessagesError
 from response_operations_ui.forms import SecureMessageForm
+from response_operations_ui.controllers.survey_controllers import get_survey_short_name_by_id
```
JamesGardiner: And this

response_operations_ui/views/messages.py
```diff
@@ -137,12 +138,41 @@ def _get_from_name(message):
         msg_from = message['@msg_from']
         return f"{msg_from.get('firstName')} {msg_from.get('lastName')}"
     except KeyError:
-        return message.get('msg_from')
+        logger.exception("Failed to retrieve message from name", message_id=message['msg_id'])
```
JamesGardiner: Is msg_id guaranteed to be present as a key?

response_operations_ui/views/messages.py
```diff
@@ -137,12 +138,41 @@ def _get_from_name(message):
...
-        name = message.get('@msg_to')[0].get('firstName') + ' ' + message.get('@msg_to')[0].get('lastName')
-    except IndexError:
-        name = message.get('msg_to')[0]
-    return name
+        if message.get('msg_to')[0] == 'GROUP':
+            return f"{get_survey_short_name_by_id(message.get('survey'))} Team"
+        return f"{message.get('@msg_to')[0].get('firstName')} {message.get('@msg_to')[0].get('lastName')}"
+    except (IndexError, TypeError):
+        logger.exception("Failed to retrieve message to name ", message_id=message['msg_id'])
+        return 'Unavailable'
```
JamesGardiner: Is there any reason to always return 'Unavailable'? Generally we could just return `None` and check in the calling function for that.

PR #92, opened 2018-03-12
Sm139 filter lists of messages

dcdarrell9:
Adds the ability to filter messages before viewing.
Currently there is WIP to look at caching the chosen survey to load up the same filters.


response_operations_ui/templates/message_select_survey.html
```diff
@@ -0,0 +1,49 @@
...
+                      <label class="label label--inline venus " for={{ survey }}>{{ survey }}</label>
+                  </div><br />
+                  {% endfor %}
+                  <div class="field__item js-focusable-box">
+                      <input class="input input--radio js-focusable" name="radio-answer" value="sand-and-gravel" id="sand-and-gravel" type="radio">
+                      <label class="label label--inline venus " for="sand-and-gravel">Sand & Gravel</label>
+                  </div><br />
+                  <div class="field__item js-focusable-box">
+                      <input class="input input--radio js-focusable" name="radio-answer" value="BRES" id="BRES" type="radio">
+                      <label class="label label--inline venus " for="BRES">BRES</label>
```
gavinleeedwards: Any reason why are Sand & Gravel and BRES still hard coded here?

response_operations_ui/views/messages.py
```diff
@@ -132,6 +177,11 @@ def _refine(message):
     }
 
 
+def _get_survey_id(selected_survey):
+    survey_messages = survey_controllers.get_survey(selected_survey)
+    return survey_messages['survey']['id']
```
insacuri: Need to wrap this in a try/catch block.  If survey_messages['survey'] is somehow empty, then when it tries to get the ['id'] then it'll throw a typeError

response_operations_ui/templates/message_select_survey.html
```diff
@@ -0,0 +1,49 @@
...
+                  <div class="field__item js-focusable-box">
+                      <input class="input input--radio js-focusable" name="radio-answer" value={{ survey }} id={{ survey }} type="radio">
+                      <label class="label label--inline venus " for={{ survey }}>{{ survey }}</label>
+                  </div><br />
+                  {% endfor %}
+                  <div class="field__item js-focusable-box">
+                      <input class="input input--radio js-focusable" name="radio-answer" value="sand-and-gravel" id="sand-and-gravel" type="radio">
+                      <label class="label label--inline venus " for="sand-and-gravel">Sand & Gravel</label>
+                  </div><br />
+                  <div class="field__item js-focusable-box">
```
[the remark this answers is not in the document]
dcdarrell9: I couldn't find a way to get the ID correctly, it would come up as 'sand' because it's stopping at the space in 'sand & gravel'... mostly for when we add the acceptance tests, but sand as an ID could be fine?

response_operations_ui/views/messages.py
dcdarrell9: `return survey_controllers.get_survey(selected_survey)['survey']['id']`
??

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
...
+    }
+
+    try:
+        refined_messages = [_refine(message) for message in message_controllers.get_thread_list(params)]
+        if selected_survey == 'FDI':
+            survey_id = _get_FDI_survey_id()
+        else:
+            survey_id = _get_survey_id(selected_survey)
+
+        filtered_messages = [messages for messages in refined_messages if messages['survey_id'] == survey_id]
```
insacuri: Is this right? It reads really weird to me.  Comprehensions are `[ expression for item in list if conditional ]`
but I don't get what the expression part of this one is doing.  Should the 2nd messages  actually be called 'message'?
The one above is clear, it's refining each message in the thread list.  It could also be me being dumb >_<

AdamHawtin: Is this a temporary solution? The get threads endpoint in SM can filter by survey. Backstage may need an update to pass that arg through but this filtering can be done at a database level rather than here.

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
...
+
+    params = {
+        'label': request.args.get('label'),
+        'survey': request.args.get('survey')
+    }
+
+    try:
+        refined_messages = [_refine(message) for message in message_controllers.get_thread_list(params)]
+        if selected_survey == 'FDI':
+            survey_id = _get_FDI_survey_id()
```
warrenbailey: Does this work for FDI message? Because in this scenario you have a list of survey ids and not a single id so I don't see how the following code would work:

e.g. messages['survey_id'] == survey_id]

response_operations_ui/common/surveys.py
```diff
@@ -0,0 +1,13 @@
...
+class Surveys(Enum):
+    """Includes all surveys for the api"""
+    ASHE = "ASHE"
+    BLOCKS = "Blocks"
+    BRICKS = "Bricks"
+    FDI = "FDI"
+    NBS = "NBS"
+    PCS = "PCS"
+    QBS = "QBS"
+    SANDANDGRAVEL = "Sand&Gravel"
```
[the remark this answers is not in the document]
warrenbailey: Does this match the survey short name in the survey service?

I think this needs to be sorted because in the UI it shouldn't display without spaces.

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
...
+    breadcrumbs = [{"title": selected_survey + " Messages"}]
+
+    params = {
+        'label': request.args.get('label'),
+        'survey': request.args.get('survey')
+    }
+
+    try:
+        refined_messages = [_refine(message) for message in message_controllers.get_thread_list(params)]
+        if selected_survey == 'FDI':
```
warrenbailey: You could use the enum here rather than the string FDI

response_operations_ui/views/messages.py
```diff
@@ -176,6 +234,16 @@ def _refine(message):
     }
 
 
+def _get_survey_id(selected_survey):
+    survey_messages = survey_controllers.get_survey(selected_survey)
+    return survey_messages['survey']['id']
+
+
+def _get_FDI_survey_id():
+    fdi_surveys = ["AOFDI", "AIFDI", "QIFDI", "QOFDI"]
```
warrenbailey: I think these values could also go in an enum or at least with the other survey data

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
...
+
+    try:
+        refined_messages = [_refine(message) for message in message_controllers.get_thread_list(params)]
+        if selected_survey == Surveys.FDI.value:
+            survey_id = _get_FDI_survey_id()
+        else:
+            survey_id = _get_survey_id(selected_survey)
+
+        filtered_messages = [message for message in refined_messages if message['survey_id'] in survey_id]
+
```
warrenbailey: As discussed we can't do this in memory in the UI, we need to pass the survey id to secure messaging and let it filter at the database level.

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
         refined_messages = [_refine(msg) for msg in message_controllers.get_thread_list(params)]
         return render_template("messages.html", breadcrumbs=breadcrumbs, messages=refined_messages)
     except NoMessagesError:
-        return render_template("messages.html", breadcrumbs=breadcrumbs, response_error=True)
+        return render_template("messages.html",
+                               breadcrumbs=breadcrumbs,
```
warrenbailey: This route can go, everything should go through the filter now

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
...
+                                selected_survey=selected_survey))
+    else:
+        survey_list = [survey.value for survey in Surveys]
+        return render_template("message_select_survey.html",
+                               breadcrumbs=breadcrumbs,
+                               selected_survey=None,
+                               survey_list=survey_list)
+
+
+@messages_bp.route('/<selected_survey>', methods=['GET'])
```
warrenbailey: I think survey selection should be via request args and not urls.

response_operations_ui/views/messages.py
```diff
@@ -126,7 +127,64 @@ def view_messages():
...
+@login_required
+def view_selected_survey(selected_survey):
+    breadcrumbs = [{"title": selected_survey + " Messages"}]
+
+    params = {
+        'label': request.args.get('label'),
+        'survey': request.args.get('survey')
+    }
```
warrenbailey: These params end up being set to None (as there are no request args).  if you follow them to secure messaging they are actually used in the database query.

In this scenario I don't think we need the label parameter and we should be setting the survey to the survey id (or ref??) that we're filtering for.

response_operations_ui/views/messages.py
warrenbailey: Discussed with Leah and the UX team would rather url parameters

tests/views/test_message.py
```diff
@@ -307,3 +311,32 @@ def test_conversation_subject_error(self, mock_request):
...
+        self.assertIn("Please select a survey to filter your messages.".encode(), response.data)
+        self.assertIn("Filter messages by survey".encode(), response.data)
+
+    @requests_mock.mock()
+    def test_get_messages_survey_does_not_exist(self, mock_request):
+        mock_request.get(shortname_url)
+
+        response = self.app.get("/messages/this-survey", follow_redirects=True)
+
+        self.assertEqual(response.status_code, 500)
```
emilioward: If this doesn't exist shouldn't it be a 404?

response_operations_ui/views/messages.py
```diff
@@ -176,6 +223,15 @@ def _refine(message):
     }
 
 
+def _get_survey_id(selected_survey):
+    survey_messages = survey_controllers.get_survey(selected_survey)
+    return survey_messages['survey']['id']
```
warrenbailey: We should add a method to the survey controller to get survey id by short name from the cache dict

response_operations_ui/views/messages.py
```diff
@@ -176,6 +223,15 @@ def _refine(message):
...
+
+
+def _get_FDI_survey_id():
+    return [survey_controllers.get_survey(fdi_survey.value)['survey']['id'] for fdi_survey in FDISurveys]
```
warrenbailey: Then we could use the new get_survey_id_by_shortname function here too

response_operations_ui/views/messages.py
```diff
@@ -111,22 +114,66 @@ def _populate_form_details_from_hidden_fields(form):
...
+    survey_list = [survey.value for survey in Surveys]
+
+    if request.method == 'POST':
+        selected_survey = request.form.get('radio-answer')
+        if not selected_survey:
+            return render_template("message_select_survey.html",
+                                   breadcrumbs=breadcrumbs,
+                                   selected_survey=None,
+                                   survey_list=survey_list,
+                                   response_error=True)
```
warrenbailey: We could tidy this up a bit by doing something like this:

    if request.method == 'POST':
        selected_survey = request.form.get('radio-answer')
        if selected_survey:
            return redirect(url_for("messages_bp.view_selected_survey",
                            selected_survey=selected_survey))
        else:
            response_error = True
    else:
        response_error = False

    survey_list = [survey.value for survey in Surveys]
    return render_template("message_select_survey.html",
                               breadcrumbs=breadcrumbs,
                               selected_survey=None,
                               response_error=response_error,
                               survey_list=survey_list)

tests/views/test_message.py
dcdarrell9: It should be, and we do get a 404, but rops changes it to 500 👎

response_operations_ui/common/mappers.py
```diff
@@ -1,5 +1,10 @@
 import calendar
 import iso8601
+import re
+
+
+def format_short_name(short_name):
+    return re.sub('(&)', r' \1 ', short_name)
```
AdamHawtin: This function could do with a comment explaining it's purpose.

PR #95, opened 2018-03-13
SM111 Remove Tabs, Correct CSS 

AdamHawtin:
The inbox/sent/draft tabs are no longer part of the messages pages, this update removes them.

Also changes many of the CSS classes in the messages table to the correct styles.


insacuri:
Can you check with Nic/Leah/<insert relevant person> about the nav.css file?  There are a lot of styles in there that are now no longer in use with this change.  I don't know if we're meant to just leave the css files alone or keep them as tidy as possible but having unused styles feels wrong to me.

PR #96, opened 2018-03-14
Bug fix: Missing string quotes in html

AdamHawtin:
The calling form for creating a message in the reporting unit page wasn't wrapping it's values in string quotes. If a string had a space in, only the first word up to the space was actually getting passed through the form.


JamesGardiner:
Does this handle strings with quotes in them?

PR #98, opened 2018-03-14
Functionality to only get 'FDI' as the short name for all FDI surveys

AdamHawtin:
The UI works off the survey short names from the survey service, the FDI surveys are actually several different surveys. The UI requires that only FDI is shown in some places for this whole group of surveys, this update adds a wrapper function to allow that.


response_operations_ui/controllers/survey_controllers.py
```diff
@@ -32,6 +34,17 @@ def get_survey(short_name):
...
+def fdi_survey_short_name_wrapper(get_survey_short_name_func):
+    """Function wrapper that converts the FDI specific survey names to all return as the string 'FDI'"""
+    @wraps(get_survey_short_name_func)
+    def convert_fdi_surveys(survey_id):
+        survey_short_name = get_survey_short_name_func(survey_id)
+        if survey_short_name in FDISurveys.__members__:
+            return "FDI"
+        return survey_short_name
+    return convert_fdi_surveys
+
```
warrenbailey: I don't see the need for a function wrapper here, i feel it over complicates it. It could just be part of the function.

Why not key all the FDI survey ids to "FDI" in the cache on initialisation rather than every call too?

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -32,6 +34,17 @@ def get_survey(short_name):
     return response.json()
 
 
+def fdi_survey_short_name_wrapper(get_survey_short_name_func):
+    """Function wrapper that converts the FDI specific survey names to all return as the string 'FDI'"""
+    @wraps(get_survey_short_name_func)
+    def convert_fdi_surveys(survey_id):
+        survey_short_name = get_survey_short_name_func(survey_id)
+        if survey_short_name in FDISurveys.__members__:
```
warrenbailey: enum is an iterable so access to __members__ isn't need you can just do:

for survey in FDISurveys:

PR #100, opened 2018-03-15
Polish sign in

insacuri:
Text and style now matches the prototype for both sign-in and sign-out functionality


insacuri:
There is still 1 more bit of styling that needs to be done, but the rest can be reviewed independent of this work

response_operations_ui/forms.py
```diff
@@ -9,8 +9,8 @@
 
 
 class LoginForm(FlaskForm):
-    username = StringField('Username', [InputRequired("Username is required")])
-    password = PasswordField('Password', [InputRequired("Password is required")])
+    username = StringField('Username', [InputRequired("Please enter a username")])
+    password = PasswordField('Password', [InputRequired("Please enter a password")])
```
gemmai95: You may need to have a look at the acceptance tests to aline with these changes.

response_operations_ui/error_handlers.py
```diff
@@ -19,8 +19,8 @@ def api_error(error):
 
 @app.errorhandler(401)
 def handle_authentication_error(error):
-    logger.error('Authentication failed')
-    flash('Authentication failed', category='failed_authentication')
+    logger.info('Authentication failed')
```
atrozzo: I would not use a info level for an failing authentication. The min required level is warning

response_operations_ui/templates/sign_in.html
```diff
@@ -4,72 +4,88 @@
...
-                            <h1 id="error-count" class="panel__title venus">There are {{ form.errors|length }} errors</h1>
-                            {% endif %}
+            <section>
+                <form method="post" class="form u-mt-m">
+                    {{ form.csrf_token }}
+                    <div>
+                        {% if form.errors %}
+                        <div class="panel panel--error u-mb-m">
+                            <div class="panel__header">
+                                {% if form.errors|length == 1 %}
```
atrozzo: do not you already check form.errors in the previous statment ?

response_operations_ui/error_handlers.py
insacuri: Why?  Nothing is actually wrong here.  Someone just put in the wrong username and password and noone cares (but we log it anyway just incase we want to track odd usage patterns).  We have alerts set up for log level errors because if we error then something really has gone wrong.

response_operations_ui/templates/sign_in.html
insacuri: yea, but I need to get the variable again to test the length

response_operations_ui/forms.py
insacuri: Have you got an example for me?  I ran the tests and nothing failed (related to this.... the suite was failing on other non-sign-in areas)

response_operations_ui/error_handlers.py
atrozzo: Thats why loggers provide different level of logs. When you do log anaylisys info are just information relveleant to the business. Warning and errors to trace not usual behavior , unexpected , errors , possible danger , failing operations , but DEF NOT use a info for authentication failed !!

response_operations_ui/forms.py
gemmai95: It may still work as it is looking for Username is required and Password is required.

response_operations_ui/templates/sign_in.html
atrozzo: you can have a function that return the right message in base of the errors lenght. So that you remove code duplication and IF statemnet that increase code complexity.

response_operations_ui/templates/sign_in.html
```diff
@@ -4,72 +4,88 @@
...
-                              {% for error in form.password.errors %}
-                                <li class="panel__title venus"><a href='#password'>{{ error }}</a></li>
-                              {% endfor %}
-                            </ol>
+                        {% endif %}
+                        <h1 class="saturn">Sign in</h1>
+                        {% if messages %}
+                            {% for category, message in messages %}
+                                {% if category == "failed_authentication" %}
+                                <div id="username-panel" class="field panel panel--simple panel--error">
```
atrozzo: Thsi is perfect candate for encapsulation , so the code will be more redable. You might have a function print_username_panel_error() where you actually do the same check 
 {% if messages %}
                            {% for category, message in messages %}
                                {% if category == "failed_authentication" %}
                                <div id="password-panel" class="field panel panel--simple panel--error u-mb-s">
                                {% endif %}
                            {% endfor %}
{% else %}

response_operations_ui/error_handlers.py
JamesGardiner: I'm going to jump in and say this is where the 'WARN' level is useful. Not something that would make me sit up and take notice, but useful enough to warrant distinguishing them from background noise of the INFO level.

insacuri: That actually seems reasonable.  Changed it to warn.  The only slight problem with it is there might be a splunk alert if there are a ton of these in 1 go, but that also seems like something we might want to know if we got a lot of warnings in a short time

response_operations_ui/templates/sign_in.html
insacuri: Unless I'm missing something, calling form.errors after the first time is identical to assigning it to a variable and calling that variable multiple times.
https://github.com/wtforms/wtforms/blob/master/wtforms/form.py#L149 shows that it caches the result after it gets called once.
Assuming that's true, then why bother assigning a variable or subclassing or doing anything, when form.errors|length is readable and cheap to do twice.  Or am I missing something?

response_operations_ui/forms.py
insacuri: I was being blind on thursday, the tests did need updating so I've got a PR in the tests repo for it

response_operations_ui/templates/sign_in.html
atrozzo: easy, it s called code reuse and apply DRY principle . Keeps your code cleaner and tidy :)

jcox-dev: If it's already being cached (as the WTForms code suggests) then there's no reason to over-engineer here and just stick with being explicit.

atrozzo: @jcox-dev is just because it s meant to be a clening task. Just because is a UI does not mean that is good to have few nested if and for condition with few code repetition. But mine is just a suggestion, but leave the code in the same state . Breaking DRY principle , Code complexity and generaly speaking we lack in UI strategy generating flask page that looks like the old JSP/ASP ;) BUt it s fine for me if you want leave like it is.

jcox-dev: I agree, we should probably have more partials that do one thing well or at least some improved documentation for presenting stuff on the front-end internally. Probably worth checking with tech leads to bring it up in their meetings.

PR #104, opened 2018-03-21
Us070

rjweeks:
https://trello.com/c/nnkdCIqd/157-us070-int-view-all-outstanding-responses-l-needed-for-release-18


response_operations_ui/views/reporting_units.py
```diff
@@ -129,6 +129,12 @@ def resent_verification(ru_ref, email, party_id):
                             info='Verification email re-sent'))
 
 
+@reporting_unit_bp.route('response_chasing/<ce_id>', methods=['GET'])
```
andrewmil: Should we have this link/the button in this PR at all?
Might be better saving it for the PR introducing reports so we don't have a broken button in master

response_operations_ui/views/reporting_units.py
```diff
@@ -129,6 +129,12 @@ def resent_verification(ru_ref, email, party_id):
...
+@login_required
+def response_chasing(ce_id):
+    logger.debug('Response chasing', ce_id=ce_id)
```
insacuri: Why is there an endpoint who's only job is to log out a message? I think this should be removed but I wasn't part of this card so there might be a good reason for it, though I can't think of one off the top of my head

rjweeks: This is incomplete at the moment, there will be a call here to the new Reporting Service.

atrozzo: @rjweeks you should lable with WIP label so that we know.

response_operations_ui/views/reporting_units.py
```diff
@@ -8,6 +8,7 @@
 from response_operations_ui.controllers import case_controller
 from response_operations_ui.controllers import contact_details_controller
 from response_operations_ui.controllers import reporting_units_controllers
+from response_operations_ui.controllers import collection_exercise_controllers
```
andrewmil: This import is unused

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -9,6 +9,14 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
+def download_report(ce_id):
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/collection-exercise/download-report/{ce_id}'
+    response = requests.get(url)
+    if response.status_code != 200:
+        raise ApiError(response)
+    return response
```
andrewmil: For consistency there'd be a debug level success log here

andrewmil:
:+1: Looks good but we do need tests for the new functionality

PR #105, opened 2018-03-21
SM114 Distinguish Unread Messages

AdamHawtin:
This adds different styling to unread messages in the messages view. Unread messages should now be highlighted in bold font with a circle icon. Currently the unread status is determined only from the last message in the thread since this is what the secure-message endpoints gives back. 
Also adds a call to remove the "UNREAD" label on the last message in the thread when the thread is opened.

**How to review:**
Run this branch locally, try sending messages in from a respondent. When the messages are unread they should be distinguished in bold with a circle icon, when the thread is opened they should automatically have the unread status removed and no longer be highlighted in the inbox. 

Trello Card:
[SM114 INT Identify Read/Unread Messages](https://trello.com/c/hY01M9Lv/63-sm114-int-identify-read-unread-messages-5)

Confluence Page:
[SM114: Internal User - Identify if Messages are Read/Unread in the inbox](https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/223314187/SM114+Internal+User+-+Identify+if+Messages+are+Read+Unread+in+the+inbox)


response_operations_ui/controllers/message_controllers.py
```diff
@@ -69,6 +69,20 @@ def send_message(message_json):
         raise ApiError(ex.response)
 
 
+def remove_unread_label(message_id):
+    url = f'{current_app.config["BACKSTAGE_API_URL"]}/v1/secure-message/update-label/{message_id}'
+    json = dumps({"label": "UNREAD", "action": "remove"})
```
JamesGardiner: Dumps just serializes this to a string - if the string never changes just format it correctly and cut out the call to dumps. Also, don't override the json module name with a var.

response_operations_ui/controllers/message_controllers.py
```diff
@@ -69,6 +69,20 @@ def send_message(message_json):
...
+
+    logger.debug("Removing message unread label", message_id=message_id)
+    response = requests.put(url, headers={"Authorization": _get_jwt(), "Content-Type": "application/json"}, data=json)
```
JamesGardiner: Will this work against the current implementation? I was under the impression that it only accepted GET and POST

response_operations_ui/views/messages.py
```diff
@@ -191,6 +191,8 @@ def view_conversation(thread_id):
             {"title": "Unavailable"}
         ]
 
+    if refined_thread[-1]['unread']:
```
JamesGardiner: Could this ever throw an IndexError?

AdamHawtin: No, at this point the thread is refined and potential missing values have already been handled

dcdarrell9:
Looking good. Not sure if it's important or not but if you've got an unread message and you click it to view the conversation, if you click the back button then it looks as if it stays unread (it does get classed as read)

PR #108, opened 2018-03-22
Upgrade jquery from 1.7.1 to 3.3.1

insacuri:
There seems to be only 2 files that actually uses jquery (static/js/validate-ci.js and static/js/read-csv.js), and all the functions it uses are in jquery 3.3.1 as well as 1.7.1 so there shouldn't be any downside.
I'd encourage people to test a bit of the site they know to make sure nothing is wrong, but I'm fairly confident that it should be fine as the site uses such a small amount of it.


insacuri:
Closing because another branch based off this branch, but with extra features got merged into master

PR #110, opened 2018-03-22
Fix multiple send

insacuri:
If this goes in before https://github.com/ONSdigital/response-operations-ui/pull/108 then that PR should be deleted (as it this branch was based off that, with the jquery upgrade).

This fixes a bug where the send button can be pressed multiple times, creating multiple messages while the original submission is sending

### DIFF
--- a/response_operations_ui/templates/create-message.html
+++ b/response_operations_ui/templates/create-message.html
@@ -4,8 +4,6 @@
 
 {% block main %}
 
-<script src="/static/js/disable-send.js"></script>
-
 {% if form.errors %}
     <div class="panel panel--error">
         <div class="panel__header">
--- a/response_operations_ui/templates/create-message.html
+++ b/response_operations_ui/templates/create-message.html
@@ -4,6 +4,8 @@
 
 {% block main %}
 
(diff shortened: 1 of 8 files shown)


response_operations_ui/templates/create-message.html
```diff
@@ -4,6 +4,8 @@
 
 {% block main %}
 
+<script src="/static/js/disable-send.js"></script>
```
jcox-dev: This isn't necessary as all JavaScript assets are minified into a single file and available to all templates that use the base template.

See: [minifier source](https://github.com/ONSdigital/response-operations-ui/blob/master/response_operations_ui/__init__.py#L22).
And: where the single file is included in [base.html](https://github.com/ONSdigital/response-operations-ui/blob/master/response_operations_ui/templates/layouts/base.html#L17).

insacuri: Removed it, forgot all the js got minified

insacuri:
@benjefferies yea, you're right.  But because it's internal it's a good enough fix as it's not really in any internal users interest to screw up their own inbox

PR #111, opened 2018-03-23
Us107 edit ce details

emilioward:
This PR adds functionality to allow an internal user to update the shown to respondent as date in a collection exercise.

Run this PR with the PR for ras-backstage to test changes

https://trello.com/c/a3s4Yy3x


emilioward:
[US107: INT Edit Collection Exercise details](https://trello.com/c/a3s4Yy3x/61-us107-int-edit-collection-exercise-details)

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -30,3 +30,20 @@ def execute_collection_exercise(short_name, period):
...
+    collection_exercise_details = {
+        "user_description": user_description
+    }
+
+    response = requests.put(url, json=collection_exercise_details)
+    if response.status_code != 200:
+        raise ApiError(response)
+
+    logger.debug('Successfully updated collection exercise details', collection_exercise_id=collection_exercise_id)
+    return 200
```
andrewmil: We don't need to return anything here. If there's a failure an exception will be raised

response_operations_ui/templates/edit-collection-exercise-details.html
```diff
@@ -0,0 +1,104 @@
...
+
+                  <br>
+
+                  <label class="label">
+                    Period as shown to respondent
+                    <span class="u-vh">Current value: </span>
+                    {{ user_description }}
+                    </span>
+                  </label>
+                  {{ form.user_description(class_="input input--text", id='user_description', required=required, maxlength=254) }}
```
andrewmil: The current `Period as shown to respondent` value should appear in the input box when the page loads

response_operations_ui/templates/edit-collection-exercise-details.html
```diff
@@ -0,0 +1,104 @@
+
+<html dir="ltr" lang="en-gb">
```
andrewmil: We have a spoke template layout which could be used to reduce this file. 
You can look at `new-enrolment-code.html` to see how to use it

emilioward:
@gemmai95 @andrewmil Hi, I saw the notification for this and thought I should reply. The internal period part was part of the original ticket, but this was removed after talking with Nic so it was only being able to edit the shown to respondent date. Nic designed all the other parts but I don't think they are getting implemented and more discussion needs to be done around them.

gemmai95:
@emilioward Tej spoke to me about this yesterday. I double checked with the BA and PO and the internal period does need to be changed up until ready for live. I am hoping to be able to look more into this tomorrow

response_operations_ui/views/collection_exercise.py
```diff
@@ -216,3 +217,28 @@ def _format_ci_file_name(collection_instruments, survey_details):
...
+@collection_exercise_bp.route('/<short_name>/<period>/edit-collection-exercise-details', methods=['POST'])
+@login_required
+def edit_collection_exercise_details(short_name, period):
+    form = request.form
+
+    collection_exercise_controllers.update_collection_exercise_details(form.get('collection_exercise_id'),
+                                                                       form.get('user_description'),
+                                                                       form.get('period'))
+
+    return redirect(url_for('surveys_bp.view_survey', short_name=short_name, period=period))
```
jcox-dev: Is `period=period` needed here for [view_survey](https://github.com/ONSdigital/response-operations-ui/blob/master/response_operations_ui/views/surveys.py#L26)?

jcox-dev: Is `period=period` needed here for [view_survey](https://github.com/ONSdigital/response-operations-ui/blob/master/response_operations_ui/views/surveys.py#L26)?

tests/views/test_collection_exercise.py
```diff
@@ -342,3 +350,46 @@ def test_failed_execution(self, mock_request):
...
+        mock_request.get(url_get_collection_exercise, json=edited_collection_exercise_details)
+        response = self.app.post(f"/surveys/test/000000/edit-collection-exercise-details",
+                                 data=changed_details, follow_redirects=True)
+        return response
+
+    @requests_mock.mock()
+    def test_update_collection_exercise_details_success(self, mock_request):
+        changed_ce_details = {
+            "collection_exercise_id": '14fb3e68-4dca-46db-bf49-04b84e07e77c',
+            "user_description": '16th June 2019'
```
jcox-dev: Do we need a separate test for POSTing a change of period here?

response_operations_ui/templates/edit-collection-exercise-details.html
```diff
@@ -59,12 +59,14 @@ <h1 class="jupiter">Edit collection exercise details</h1>
 
                   <br>
 
+                {% if ce_state != 'READY_FOR_LIVE' and ce_state != 'LIVE' %}
```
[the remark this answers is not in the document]
emilioward: I put this in flask jinja so that if the state was `ready for live` or `live` then the option to edit the period for the user disappeared. Does this need to be added in the view too?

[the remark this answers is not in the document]
jcox-dev: It's not needed in addition to the template code, but it would be better to have the logic related to states instead in the view to keep the business logic in one place.

e.g. [something here perhaps](https://github.com/ONSdigital/response-operations-ui/blob/e234d61bf651e660b78acabc6c616c7fdcaf1e49/response_operations_ui/views/collection_exercise.py#L228): 

```python
def view_collection_exercise_details(short_name, period):
...
    ce_state = ce_details['collection_exercise']['state']
    show_edit_period = ce_state not in ('READY_FOR_LIVE', 'LIVE')
...
    return render_template('edit-collection-exercise-details.html', show_edit_period=show_edit_period, ...
```

then in the template:

```html
...
{% if show_edit_period %}
...
{% endif %}
...
```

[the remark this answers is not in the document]
emilioward: Oh I understand yeah thats a lot cleaner way to implement that, thank you :)

response_operations_ui/forms.py
```diff
@@ -57,6 +57,18 @@ def __init__(self, form, default_values=None):
...
 
+class EditCollectionExerciseDetailsForm(FlaskForm):
+    user_description = StringField('user_description')
+    period = StringField('period')
+    collection_exercise_id = HiddenField('collection_exercise_id')
+
+    def __init__(self, form, default_values=None):
+        super().__init__(form)
+        if default_values:
+            self.user_description.data = default_values.get('user_description')
```
andrewmil: Apparently these lines are untested. Is there any case where we actually pass in default values?

response_operations_ui/templates/collection-exercise.html
```diff
@@ -327,6 +337,8 @@ <h2 class="venus u-pt-m">Reference information for this collection exercise</h2>
               <td class="table--cell tbl-ce-value" name="survey-info">
                 {{survey.surveyRef}} {{survey.shortName}}
               </td>
+              <td class="table--cell tbl-ce-">
```
andrewmil: Do we need the empty columns here?

emilioward: I added those in for the future if those columns ever needed to be updated, plus it visually looks better as some columns were uneven compared to the ones with pencils

response_operations_ui/forms.py
```diff
@@ -57,6 +57,18 @@ def __init__(self, form, default_values=None):
             self.telephone.data = default_values.get('telephone')
 
 
+class EditCollectionExerciseDetailsForm(FlaskForm):
+    user_description = StringField('user_description')
```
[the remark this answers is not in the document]
emilioward: There doesn't appear to be any form of validation for the period or user_description which is a free text edit field in the story. There probably will be for the future but for now there isn't.

response_operations_ui/forms.py
emilioward: ah no there isn't I added those in testing some stuff out, i'll remove them 👍

PR #112, opened 2018-03-26
Improve Messages Date Time Formatting

AdamHawtin:
The date time formatting with Maya didn't exactly match the UI design. This update removes dependency on Maya and does the formatting exactly as the UI designs specify.

Trello card:
[SM111-int-view-all-inbox-messages-from-threads-13](https://trello.com/c/gfaiP90r/60-sm111-int-view-all-inbox-messages-from-threads-13)
Note how the date times are displayed on the UI image on this card, with 'Today at ' or 'Yesterday at' where appropriate and 'DD MMM YYYY HH:MM' for dates any older than yesterday. The actually dates should now match that formatting, before this change older dates would sometimes not display the year where they should. The easiest way to test this is to have secure message running with a local database and manually change the timestamps on the events in the database to check that all dates appear as they should.


response_operations_ui/common/dates.py
```diff
@@ -0,0 +1,26 @@
...
+    if the given date is today or yesterday respectively otherwise returns the full date in the format '%b %d %Y %H:%M'
+    """
+    try:
+        datetime_parsed = datetime.strptime(datetime_string, string_format)
+    except (OverflowError, ValueError, AttributeError):
+        # Passed value wasn't date-ish or date arguments out of range
+        logger.exception("Failed to parse date from message", sent_date=datetime_string)
+        return datetime_string
+
+    delta = datetime.date(datetime_parsed) - date.today()
```
ryangrundy7: Why is it called delta?

PR #113, opened 2018-03-26
Sm108 reply to message

JamesGardiner:
This PR adds the ability to reply to a respondents message.


response_operations_ui/views/messages.py
```diff
@@ -11,6 +11,7 @@
 from response_operations_ui.controllers import message_controllers, survey_controllers
 from response_operations_ui.exceptions.exceptions import ApiError, InternalError, NoMessagesError
 from response_operations_ui.forms import SecureMessageForm
+from response_operations_ui.controllers.survey_controllers import get_survey_short_name_by_id, get_survey_ref_by_id
```
insacuri: 🎨 Alphabetise imports ;)

response_operations_ui/views/messages.py
```diff
@@ -48,69 +49,53 @@ def create_message():
...
-        'msg_from': current_user.id,
-        'msg_to': [form.hidden_to_uuid.data],
-        'subject': form.subject.data,
-        'body': form.body.data,
-        'thread_id': "",
-        'collection_case': "",
-        # TODO Make this UUID for v2 api
-        'survey': form.hidden_survey_id.data,
-        'ru_id': form.hidden_to_ru_id.data})
+    if refined_thread[-1]['unread']:
```
insacuri: Should there be some error handling here?  It'll throw a keyError if unread isn't defined.  Or can we basically guarantee that we'll have what we want here?

dcdarrell9: From what I remember, the `_refine` function catches any errors?

insacuri: just took a looksee, it does indeed.

PR #114, opened 2018-03-27
US091 - Find respondent by email

gemmai95:
Find respondent by email. This works with party and backstage changes. Currently adding more unit tests to up code coverage.


gemmai95:
[US091: Internal user to search respondent via their email address](https://trello.com/c/fpJMbZOk/62-us091-internal-user-to-search-respondent-via-their-email-address)

response_operations_ui/controllers/respondent_controllers.py
```diff
@@ -0,0 +1,31 @@
...
+def search_respondent_by_email(email):
+    logger.debug('Search respondent via email')
+
+    request_json = {
+        'email': email
+    }
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/party/get-respondent-by-email'
+    response = requests.get(url, json=request_json)
+
+    if 'No respondent found' in response:
```
andrewmil: Will the response also be a 404? It's probably better to check on the status code rather than text in the response

response_operations_ui/templates/search-respondent.html
```diff
@@ -0,0 +1,32 @@
+{% extends "layouts/base.html" %}
+{% block page_title %}Respondents | Survey Data Collection{% endblock %}
+
+{% block main %}
+    <div role="main" id="main" class="page__main">
```
andrewmil: Could we indent the html inside this div?

response_operations_ui/views/respondents.py
```diff
@@ -0,0 +1,42 @@
...
+    return render_template('search-respondent.html', response=response, form=form, breadcrumbs=breadcrumbs)
+
+
+@respondent_bp.route('/respondent-details/<respondent_id>', methods=['GET'])
+@login_required
+def respondent_details(respondent_id):
+
+    respondent = contact_details_controller.get_contact_details(respondent_id)
+
+    return render_template('respondent.html', respondent=respondent)
```
andrewmil: Missing newline

tests/views/test_respondents.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+
+from config import TestingConfig
+from response_operations_ui import app
+
+
+class TestRespondents(unittest.TestCase):
```
andrewmil: We should have tests for the failing cases, 404/500 status codes

response_operations_ui/views/respondents.py
```diff
@@ -0,0 +1,42 @@
...
+
+        if respondent.get('id'):
+            return redirect(url_for('respondent_bp.respondent_details', respondent_id=respondent['id']))
+        else:
+            response = respondent['Response']
+
+    return render_template('search-respondent.html', response=response, form=form, breadcrumbs=breadcrumbs)
+
+
+@respondent_bp.route('/respondent-details/<respondent_id>', methods=['GET'])
```
andrewmil: We're missing breadcrumbs for this endpoint

response_operations_ui/controllers/respondent_controllers.py
gemmai95: This is being returned from the API as a 200. The response text is the only difference. I don't know why but we only seem to handle 200s and 500s in this repo and backstage.

response_operations_ui/views/respondents.py
dcdarrell9: @andrewmil @gemmai95 In some places we've added a _build_breadcrumbs function to split it out, what's your opinion on this? Because at the moment we're doing 2 things, would be nice to have it all the same

gemmai95: I seen this in the messaging stuff it does look a lot cleaner but atm I just added it into the function.

andrewmil: If we can easily generalise it to work across the repo I'm a fan of it @dcdarrell9.
Agree we should have consistency though for sure

response_operations_ui/templates/search-respondent.html
```diff
@@ -0,0 +1,32 @@
...
+        <h1 class="jupiter">Respondents</h1>
+        <div class="grid">
+            <div class="grid__col col-12@l">
+
+                <div class="ru-search">
+
+                    <form action="" method="post">
```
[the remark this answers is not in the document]
gemmai95: This is more consistency I guess. I followed the reporting unit search method of doing this.

PR #116, opened 2018-04-03
Store messages survey in redis session

ryangrundy7:
This pull request stores the survey in a redis session meaning that when the user goes back to their messages, it would be on the survey they have chosen instead of going back to the select survey screen.

How to test:

- Check the functionality by selecting a survey and making sure it returns back to the correct messages screen.
- Unit tests run

### DIFF
--- a/response_operations_ui/views/logout.py
+++ b/response_operations_ui/views/logout.py
@@ -1,5 +1,3 @@
-from contextlib import suppress
-
 from flask import Blueprint, flash, redirect, url_for, session
 from flask_login import logout_user
 
@@ -10,7 +8,7 @@ logout_bp = Blueprint('logout_bp', __name__, static_folder='static', template_fo
 @logout_bp.route('/')
 def logout():
     logout_user()
-    with suppress(KeyError):
-        del session['messages_survey_selection']
+    if 'messages_survey_selection' in session:
+        session.pop('messages_survey_selection')
     flash("You are now signed out", category='successful_signout')
     return redirect(url_for('sign_in_bp.sign_in'))
--- a/response_operations_ui/views/logout.py
+++ b/response_operations_ui/views/logout.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, flash, redirect, url_for, session
+from flask import Blueprint, flash, redirect, session, url_for
 from flask_login import logout_user
 
 
--- a/response_operations_ui/views/logout.py
+++ b/response_operations_ui/views/logout.py
@@ -11,6 +11,6 @@ logout_bp = Blueprint('logout_bp', __name__, static_folder='static', template_fo
 def logout():
     logout_user()
     with suppress(KeyError):
-        del session["messages_survey"]
+        del session['messages_survey_selection']
     flash("You are now signed out", category='successful_signout')
     return redirect(url_for('sign_in_bp.sign_in'))
--- a/response_operations_ui/views/logout.py
+++ b/response_operations_ui/views/logout.py
@@ -1,4 +1,6 @@
-from flask import Blueprint, flash, redirect, url_for
+from contextlib import suppress
+
+from flask import Blueprint, flash, redirect, url_for, session
 from flask_login import logout_user
 
 
@@ -8,5 +10,7 @@ logout_bp = Blueprint('logout_bp', __name__, static_folder='static', template_fo
 @logout_bp.route('/')
 def logout():
     logout_user()
+    with suppress(KeyError):
+        del session["message_survey"]
     flash("You are now signed out", category='successful_signout')
     return redirect(url_for('sign_in_bp.sign_in'))
--- a/response_operations_ui/views/logout.py
+++ b/response_operations_ui/views/logout.py
@@ -11,6 +11,6 @@ logout_bp = Blueprint('logout_bp', __name__, static_folder='static', template_fo
 def logout():
     logout_user()
     with suppress(KeyError):
-        del session["message_survey"]
+        del session["messages_survey"]
     flash("You are now signed out", category='successful_signout')
     return redirect(url_for('sign_in_bp.sign_in'))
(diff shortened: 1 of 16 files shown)


response_operations_ui/views/logout.py
```diff
@@ -8,5 +10,7 @@
 @logout_bp.route('/')
 def logout():
     logout_user()
+    with suppress(KeyError):
```
insacuri: Can you put a comment on this saying that you're trying to delete something and if it's not there, we don't care

response_operations_ui/views/logout.py
```diff
@@ -8,5 +10,7 @@
...
+        del session["messages_survey"]
```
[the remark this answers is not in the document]
ajmaddaford: I think it should be messages rather than inbox as inbox/sent has been dropped, maybe `messages_survey_selection`?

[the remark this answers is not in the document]
ryangrundy7: Changed it to messages_survey_selection

response_operations_ui/views/logout.py
ajmaddaford: If we know it may not be in the session, it would be clearer to check for it's existence first:
```
if 'messages_survey' in session:
    session.pop('messages_survey')
```
It can be done on one line but it's not as clear:
```
session.pop("messages_survey", None)
```

tests/views/test_message.py
```diff
@@ -410,3 +410,27 @@ def test_get_messages_survey_does_not_exist(self, mock_request):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("Filter messages by survey".encode(), response.data)
+
+    @requests_mock.mock()
+    def test_get_messages_page_with_survey(self, mock_request):
+        mock_request.get(url_get_threads_list, json=thread_list)
+        mock_request.get(url_get_surveys_list, json=self.surveys_list_json)
+        mock_request.get(shortname_url + "/ASHE", json=ashe_info)
+
+        posts_survey_json = {"radio-answer": "ASHE"}
```
ajmaddaford: Minor, this is a dict rather than json.

response_operations_ui/views/messages.py
```diff
@@ -98,36 +99,48 @@ def view_conversation(thread_id):
...
     if request.method == 'POST':
         selected_survey = request.form.get('radio-answer')
         if selected_survey:
             return redirect(url_for("messages_bp.view_selected_survey",
                                     selected_survey=selected_survey))
         else:
-            response_error = True
-    else:
-        response_error = False
+            return render_template("message_select_survey.html",
```
JamesGardiner: I prefer the previous way of doing this, there was less repetition of code.

response_operations_ui/views/messages.py
```diff
@@ -98,36 +99,48 @@ def view_conversation(thread_id):
...
-                           response_error=response_error,
+                           response_error=False,
                            survey_list=survey_list)
 
 
 @messages_bp.route('/<selected_survey>', methods=['GET'])
 @login_required
 def view_selected_survey(selected_survey):
     formatted_survey = format_short_name(selected_survey)
+    session['messages_survey_selection'] = selected_survey
```
JamesGardiner: Could this ever not exist in the session at this point?

response_operations_ui/views/logout.py
```diff
@@ -8,5 +8,7 @@
 @logout_bp.route('/')
 def logout():
     logout_user()
+    if 'messages_survey_selection' in session:
+        session.pop('messages_survey_selection')
```
JamesGardiner: I think a try except block here is still more pythonic. I know @ajmaddaford thinks differently, so won't block the PR on this, but I'd do something like:

```Python
try:
    del(session['messages_survey_selection'])
except KeyError:
    logger.info('No survey to remove from session')
finally:
    flash("You are now signed out", category='successful_signout')
    return redirect(url_for('sign_in_bp.sign_in'))
```

response_operations_ui/views/messages.py
```diff
@@ -98,11 +99,22 @@ def view_conversation(thread_id):
                            form=form)
 
 
-@messages_bp.route('/', methods=['GET', 'POST'])
+@messages_bp.route('/', methods=['GET'])
 @login_required
 def view_select_survey():
+    if "messages_survey_selection" in session:
```
JamesGardiner: I think this would read better:

```Python
try:
    selected_survey = session["messages_survey_selection"]
except KeyError:
    return redirect(url_for("messages_bp.select_survey"))
else:
    return redirect(url_for("messages_bp.view_selected_survey",
                             selected_survey=selected_survey))
```

ajmaddaford: I think that may fail the pylint no-else-return rule. Without the else it would be:
```Python
try:
    selected_survey = session["messages_survey_selection"]
except KeyError:
    return redirect(url_for("messages_bp.select_survey"))

return redirect(url_for("messages_bp.view_selected_survey",
                         selected_survey=selected_survey))
```
vs
```Python
if "messages_survey_selection" in session:
    return redirect(url_for("messages_bp.view_selected_survey",
                             selected_survey=session["messages_survey_selection"]))
 
return redirect(url_for("messages_bp.select_survey"))
```
I'm torn as to which one reads better 🤷‍♂️

response_operations_ui/views/logout.py
ajmaddaford: - I don't see the value in logging that there is no survey to remove from the session, we're not logging when it is in the session and is removed.
- Having the flash and return in the finally block implies it is in some way linked to the deletion of the survey selection key, which it's not.
- The `session.pop` should be changed to `del(session['messages_survey_selection'])` as the existence of the key has already been checked.

Taking these things into account, I think the options are:
```Python
try:
    del(session['messages_survey_selection'])
except KeyError:
    pass

flash("You are now signed out", category='successful_signout')
return redirect(url_for('sign_in_bp.sign_in'))
```
vs
```Python
if 'messages_survey_selection' in session:
    del(session['messages_survey_selection'])

flash("You are now signed out", category='successful_signout')
return redirect(url_for('sign_in_bp.sign_in'))
```

I'd be happy to go with the first option, I can see there is some clarity from not checking for the existence of the key before deletion.

PR #129, opened 2018-04-18
Stop caching messages page

ryangrundy7:
This PR stops the messages page from being cached by the browser. 

Testing:
Make sure all tests pass and unread messages become read when navigating back to the messages page using the back button.

### DIFF
--- a/response_operations_ui/views/messages.py
+++ b/response_operations_ui/views/messages.py
@@ -1,8 +1,7 @@
 import json
 import logging
 
-from flask import Blueprint, flash, g, Markup, render_template, request, redirect, session, url_for, make_response
-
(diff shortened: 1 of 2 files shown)


response_operations_ui/views/messages.py
```diff
@@ -330,3 +334,10 @@ def _get_human_readable_date(sent_date):
 
 def _get_unread_status(message):
     return 'UNREAD' in message.get('labels', [])
+
+
+@messages_bp.after_request
+def disabling_caching_headers(response):
```
ajmaddaford: This method name is a little misleading, as it could be read as disabling the cache headers i.e. not setting them. I think it would be clearer as `disable_caching`

PR #140, opened 2018-04-27
Us102 create collection exercise

emilioward:
This PR allows for a user to create a new collection exercise with validation to ensure that a period cannot be entered if it already exists for a collection exercise for that survey.

Also added in missing validation for editing a collection exercise so that a user cannot change the period to one that already exists


emilioward:
[US102: INT User to Create a CE - After R17/18 (M)](https://trello.com/c/9VgewZXM/77-us102-int-user-to-create-a-ce-after-r17-18-m)

config.py
```diff
@@ -6,10 +6,24 @@ class Config(object):
...
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
+    SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME')
+    SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD')
+    BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
     USE_SESSION_FOR_NEXT = True
     RESPONSE_OPERATIONS_UI_SECRET = os.getenv('RESPONSE_OPERATIONS_UI_SECRET', "secret")
     SECURE_MESSAGE_URL = os.getenv('SECURE_MESSAGE_URL')
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL')
+
+    RM_COLLECTION_EXERCISE_SERVICE_HOST = os.getenv('RM_COLLECTION_EXERCISE_SERVICE_HOST',
```
rhiGriff: These shouldn't be split out now as we are now using the uri and are removing the RAS/RM prefix in the variable name.

tests/views/test_collection_exercise.py
```diff
@@ -354,22 +365,29 @@ def test_failed_execution(self, mock_request):
...
         self.assertIn('Check collection instruments'.encode(), response.data)
 
-    def mock_for_update_ce_details(self, changed_details, mock_request):
-        mock_request.get(url_get_collection_exercise, json=collection_exercise_details)
-        mock_request.put(url_update_ce)
-        mock_request.get(url_get_survey_by_short_name, json=updated_survey_info)
-        response = self.app.post(f"/surveys/test/000000/edit-collection-exercise-details",
-                                 data=changed_details, follow_redirects=True)
-        return response
+    # def mock_for_update_ce_details(self, changed_details, mock_request):
```
rhiGriff: Is there a reason this has been left in commented out and not removed?

config.py
```diff
@@ -26,9 +30,13 @@ class DevelopmentConfig(Config):
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'DEBUG')
     BACKSTAGE_API_URL = os.getenv('BACKSTAGE_API_URL', 'http://localhost:8001/backstage-api')
     SECURE_MESSAGE_URL = os.getenv('SECURE_MESSAGE_URL', 'http://localhost:5050')
+    COLLECTION_EXERCISE_SERVICE = os.getenv('COLLECTION_EXERCISE_SERVICE', 'http://localhost:8145')
```
tej99: Should we keep the env var naming convention consistent with ones that are already there? COLLECTION_EXERCISE_URL?

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -63,3 +64,37 @@ def update_collection_exercise_details(collection_exercise_id, user_description,
         raise ApiError(response)
 
     logger.debug('Successfully updated collection exercise details', collection_exercise_id=collection_exercise_id)
+
+
+def create_collection_exercise(survey_id, survey_name, user_description, period):
+    logger.debug('Creating a new collection exercise for', survey_id=survey_id, survey_name=survey_name)
+    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/collection-exercise/create-collection-exercise'
```
tej99: Could just go directly to collection exercise service now, the back stage endpoint provide no additional functionality here.

emilioward: Yeah I finished all this before people started talking about not using backstage 😢

gemmai95: Probably best to close the backstage PR now.

response_operations_ui/views/collection_exercise.py
```diff
@@ -277,23 +276,70 @@ def _get_form_type(file_name):
 def view_collection_exercise_details(short_name, period):
     ce_details = collection_exercise_controllers.get_collection_exercise(short_name, period)
```
gemmai95: Can we add some logging in here?

response_operations_ui/views/collection_exercise.py
```diff
@@ -277,23 +276,70 @@ def _get_form_type(file_name):
...
-                           collection_exercise_id=ce_details['collection_exercise']['id'])
+                           collection_exercise_id=ce_details['collection_exercise']['id'],
+                           survey_id=survey_details['survey']['id'])
 
 
 @collection_exercise_bp.route('/<short_name>/<period>/edit-collection-exercise-details', methods=['POST'])
 @login_required
 def edit_collection_exercise_details(short_name, period):
-    form = request.form
+    form = EditCollectionExerciseDetailsForm(form=request.form)
```
gemmai95: Same as above should probably have some info level logging

response_operations_ui/templates/create-collection-exercise.html
```diff
@@ -0,0 +1,60 @@
...
+          <h2 id='error-creating-ce-message' class="venus">Error creating collection exercise</h2>
+          <p class="mars">Please use a period that is not in use by any collection exercise for this survey</p>
+        </div>
+      </div>
+      {% endif %}
+        <div class="field u-mb-l">
+          <label class="label" for="period">
+            Period
+          </label>
+          {{ form.period(class_="input input--text", id='period', required=required, maxlength=6, value=period) }}
```
dcdarrell9: May want to add some more validation, when I type a letter and save I get sent to error 500 page

dcdarrell9: Same kind of error with the 'shown to respondent' if I copy and paste over a certain amount of characters I get a 500

emilioward: yeah my bad i set the field limit to be 100 when in the database it has a max of 50, ill change that

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -63,3 +64,41 @@ def update_collection_exercise_details(collection_exercise_id, user_description,
...
+
+    collection_exercise_details = {
+        "surveyId": survey_id,
+        "name": survey_name,
+        "userDescription": user_description,
+        "exerciseRef": period
+    }
+
+    response = requests.post(url, json=collection_exercise_details, headers=header, auth=app.config['BASIC_AUTH'])
+    if response.status_code == 404:
```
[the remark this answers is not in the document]
andrewmil: I'd imagine it'd be one or the other
Also we probably don't need an if statement here for the 404 case if we're just raising the ApiError exception anyway?

response_operations_ui/forms.py
```diff
@@ -59,15 +61,45 @@ def __init__(self, form, default_values=None):
...
+    period = IntegerField('period')
     collection_exercise_id = HiddenField('collection_exercise_id')
+    hidden_survey_id = HiddenField('hidden_survey_id')
+
+    @staticmethod
+    def validate_period(form, field):
+        hidden_survey_id = form.hidden_survey_id.data
+        ce_details = collection_exercise_controllers.get_collection_exercises_by_survey(hidden_survey_id)
+        inputted_period = field.data
+        for key in ce_details:
```
andrewmil: `key` is a bit vague, could replace with `collection_exercise` or `ce`

response_operations_ui/controllers/collection_exercise_controllers.py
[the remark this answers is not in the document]
JamesGardiner: raise_for_status is cleaner

response_operations_ui/templates/create-collection-exercise.html
andrewmil: Should we have validation in the html like this if we're using flaskform for it?

andrewmil: :+1: Having both may the best way to go, the problem with the validation inline like this is that users can edit the html and submit stuff they shouldn't be able to, so we should always have the server side validation at least

config.py
```diff
@@ -6,10 +6,14 @@ class Config(object):
     TESTING = False
     PORT = os.getenv('PORT', 8085)
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
+    SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME')
```
gemmai95: These are now separate per service so should be COLLECTION_EXERCISE_USERNAME etc?

PR #147, opened 2018-05-09
Removing next from session on sign out

ryangrundy7:
At the moment, the session keeps a key that takes you to a page stored in the session once you signed in. This pull request removes it on sign out.

To test:

- Before logging in, go to /messages/bricks and log in

- Log back out

- When you log back in, it should take you to the homepage instead of /messages/bricks


response_operations_ui/views/logout.py
```diff
@@ -14,5 +14,10 @@ def logout():
     except KeyError:
         pass
 
+    try:
+        del (session['next'])
```
dcdarrell9: Could we add this into the other try?

ryangrundy7: I originally thought that but was thinking that it could just ignore the del(next) if for some reason survey wasn't there. it would just cause an exception and pass through

ryangrundy7: Okay i'll add that change now

warrenbailey:
Though just read all the comments and discussion on: https://github.com/ONSdigital/response-operations-ui/pull/116/.

I don't see the problem with session.pop but I see @ajmaddaford and @JamesGardiner disagree with me.

tests/views/test_sign_in.py
```diff
@@ -101,3 +101,13 @@ def test_sign_in_next_url(self, mock_request):
...
+
+    def test_sign_out_deleting_session_variables(self):
+        with self.app.session_transaction() as session:
+            session['next'] = '/messages/bricks'
+        response = self.app.get('/logout', follow_redirects=True)
+        self.assertIn(b'You are now signed out', response.data)
+        self.assertEqual(response.status_code, 200)
+        self.assertNotIn(b"Sign out", response.data)
+        with self.app.session_transaction() as session:
+            self.assertEqual(session.get("next"), None)
```
dcdarrell9: Would it be worth asserting that next is in the session before logging out?

ajmaddaford:
[Remove next from session on sign out  in response operations](https://trello.com/c/1OSrwwb9/211-remove-next-from-session-on-sign-out-in-response-operations)

PR #153, opened 2018-05-15
Allow creation of first collection exercise for survey

gemmai95:
If you try to create a collection exercise for a survey that does not already have collection exercises this errors and gives misleading errors to the user. This PR addresses this by removing the dependencies on previous collection exercises for the survey.


gemmai95:
[When creating a collection exercise another collection exercise is expected to be there otherwise this fails.](https://trello.com/c/1YFHZO9J/89-when-creating-a-collection-exercise-another-collection-exercise-is-expected-to-be-there-otherwise-this-fails)

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -143,6 +143,9 @@ def get_collection_exercises_by_survey(survey_id):
 
     response = requests.get(url, auth=app.config['COLLECTION_EXERCISE_AUTH'])
 
+    if response.status_code == 204:
+        return
```
tej99: Can we return an empty list instead?

response_operations_ui/views/collection_exercise.py
```diff
@@ -326,23 +326,40 @@ def get_create_collection_exercise_form(survey_ref, short_name):
...
+                               survey_name=survey_details['survey']['shortName'])
 
     else:
         logger.info("Creating collection exercise for survey", short_name=short_name, survey_ref=survey_ref)
         form = request.form
+
+        created_period = form.get('period')
+        ce_details = collection_exercise_controllers.get_collection_exercises_by_survey(survey_details['survey']['id'])
+
+        if ce_details:
```
tej99: Will not need this check if get_collection_exercises_by_survey returns an empty list.

response_operations_ui/views/collection_exercise.py
```diff
@@ -326,23 +326,40 @@ def get_create_collection_exercise_form(survey_ref, short_name):
...
 
 
 @collection_exercise_bp.route('/<survey_ref>-<short_name>/create-collection-exercise', methods=['POST'])
 @login_required
 def create_collection_exercise(survey_ref, short_name):
-    form = CreateCollectionExerciseDetailsForm(form=request.form)
-    if not form.validate():
+    ce_form = CreateCollectionExerciseDetailsForm(form=request.form)
+
+    survey_details = survey_controllers.get_survey(short_name)
```
tej99: Do we need this call? We have short_name and the survey id as a hidden field in the form.

PR #155, opened 2018-05-15
Directly change case group status

andrewmil:
Remove calls to backstage when changing case group response status

^Branch name is wrong


response_operations_ui/controllers/case_controller.py
```diff
@@ -9,25 +9,21 @@
...
-    if response.status_code != 200:
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        logger.exception('Error updating status', collection_exercise_id=collection_exercise_id, ru_ref=ru_ref,
+                         case_group_event=case_group_event)
         raise ApiError(response)
 
-    logger.debug('Successfully updated case group status', short_name=short_name, period=period, ru_ref=ru_ref)
+    logger.debug('Successfully updated status', collection_exercise_id=collection_exercise_id, ru_ref=ru_ref,
```
rhiGriff: Is this log line talking about a specific case or a case group?

andrewmil: It is but we don't have access to a case group id here
A case group is in a one to one relationship with a combination of ru_ref and collection exercise so we would have sight the case group from these details anyway

andrewmil: Will change the wording to `Successfully updated case group status` though

rhiGriff: That's all I wanted to clarify was the logging wording 👍

andrewmil:
[US050d: INT Manually change the response status for a given RU, Survey, CE to No Longer Required (L)](https://trello.com/c/2a1r2ZJM/105-us050d-int-manually-change-the-response-status-for-a-given-ru-survey-ce-to-no-longer-required-l)

jcox-dev:
[response-operations-ui should directly call case service (not via backstage)](https://trello.com/c/7WefKOmo/94-response-operations-ui-should-directly-call-case-service-not-via-backstage)

PR #159, opened 2018-05-17
Us053 remove loaded sample

emilioward:
Works alongside:

https://github.com/ONSdigital/ras-party/pull/131
https://github.com/ONSdigital/rm-collection-exercise-service/pull/60

This PR enables an internal user to unlink a sample from a collection exercise
https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/248611078/US053+INT-+Remove+a+loaded+sample

**How to test:**

Remove a loaded sample from a collection exercise that is not in `ready for live` or `live` state then check to see if the sample is removed.


emilioward:
[US053: INT- Remove a loaded sample (13)](https://trello.com/c/IQ8agznp/86-us053-int-remove-a-loaded-sample-13)

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -161,6 +161,26 @@ def get_case_group_status_by_collection_exercise(case_groups, collection_exercis
                 if case_group['collectionExerciseId'] == collection_exercise_id)
 
 
+def unlink_sample_summary(collection_exercise_id, sample_summary_id):
+    logger.debug('un-linking sample summary from collection exercise', collection_exercise_id=collection_exercise_id,
+                 sample_summary_id=sample_summary_id)
+    url = f'{app.config["COLLECTION_EXERCISE_URL"]}/collectionexercises/unlink/{collection_exercise_id}' \
+          f'/sample/{sample_summary_id}'
+
+    response = requests.put(url, auth=app.config['COLLECTION_EXERCISE_AUTH'])
```
tej99: The put call has now changed to delete `requests.delete(url, auth=app.config['COLLECTION_EXERCISE_AUTH'])`

response_operations_ui/templates/collection-exercise.html
```diff
@@ -231,7 +245,11 @@ <h2 class="venus u-pt-m">Sample</h2>
               </tr>
             </tbody>
           </table>
-        <p>Sample loaded: {{sample.ingestDateTime}}</p>
+        <p>Sample loaded: {{sample.ingestDateTime}}
+          {% if show_edit_period %}
```
tej99: Can we replace all uses of show_edit_period and `if not ce.state == 'Live' and not ce.state == 'Ready for live'` checks with `if not locked` as they do not take 'EXECUTION_STARTED', 'VALIDATED', 'EXECUTED' states under consideration.

response_operations_ui/views/collection_exercise.py
```diff
@@ -37,6 +46,12 @@ def view_collection_exercise(short_name, period, error=None, success_panel=None)
         }
     ]
 
+    message_key = request.args.get('message_key')
+    if message_key:
+        info_message = get_info_message(message_key)
```
tej99: All messages (even error messages) show up as green info success panel

response_operations_ui/views/collection_exercise.py
```diff
@@ -57,8 +72,10 @@ def view_collection_exercise(short_name, period, error=None, success_panel=None)
                            collection_instruments=ce_details['collection_instruments'],
                            eq_ci_selectors=ce_details['eq_ci_selectors'],
                            error=error,
+                           error_panel=error_panel,
```
tej99: Doesn't look like error_panel is being used?

emilioward: I'm refactoring this in my PR so this will change 👍

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -161,6 +161,26 @@ def get_case_group_status_by_collection_exercise(case_groups, collection_exercis
...
+    try:
+        response.raise_for_status()
+    except HTTPError:
+        logger.exception('Failed to unlink sample summary from collection exercise',
+                         collection_exercise_id=collection_exercise_id, sample_summary_id=sample_summary_id)
+        return False
+
+    logger.debug('Successfully unlinked sample summary from a collection exercise',
+                 collection_exercise_id=collection_exercise_id, sample_summary_id=sample_summary_id)
+    return response.json
```
jcox-dev: Should this be `response.json()`?

response_operations_ui/views/collection_exercise.py
```diff
@@ -8,21 +8,35 @@
...
 
 collection_exercise_bp = Blueprint('collection_exercise_bp', __name__,
                                    static_folder='static', template_folder='templates')
 
 
+def get_success_message(success_key):
+    return {
+        'sample_removed_success': "Sample removed",
+        'sample_loaded_success': "Sample successfully loaded"
+    }[success_key]
```
tej99: This will throw a keyError if an incorrect success key is passed through the query string. Using `{}.get(success_key, None)` might be better?

response_operations_ui/views/collection_exercise.py
```diff
@@ -8,21 +8,35 @@
...
+
+
+def get_error_message(error_key):
+    return {
+        'sample_removed_error': "Error failed to remove sample"
+    }[error_key]
```
tej99: Using `{}.get(error_key, None)` might be better.

response_operations_ui/views/collection_exercise.py
```diff
@@ -37,12 +51,21 @@ def view_collection_exercise(short_name, period, error=None, success_panel=None,
         }
     ]
 
+    success_key = request.args.get('success_key')
+    error_key = request.args.get('error_key')
+    if success_key:
```
tej99: If you use .get() in get_success_message and get_error_message you won't need this if else statement.
```
success_message = get_success_message(success_key)
error_message = get_error_message(error_key)
```

response_operations_ui/controllers/collection_exercise_controllers.py
tej99: Request to delete sample link does not return a JSON response. Maybe just return True?

response_operations_ui/templates/collection-exercise.html
```diff
@@ -243,6 +257,21 @@ <h2 class="venus u-pt-m">Sample</h2>
...
+                </td>
+              </tr>
+            </tbody>
+          </table>
+          <p>Sample loaded: {{sample.ingestDateTime}}
+            {% if not locked %}
+            <a class="u-ml-xs" id="remove_sample_btn" href="{{ url_for('collection_exercise_bp.get_confirm_remove_sample', short_name=survey.shortName, period=ce.exerciseRef) }}">Remove</a>
+            {% endif %}
+          </p>
                   <tr class="table--row">
```
tej99: Looks like this is duplicated 
```
              <tr class="table--row">
                     <td class="table--cell tbl-ce-">
                       Collection instruments
                     </td>
                     <td class="table--cell tbl-ce-" name="total-ci">
                       {{sample.expectedCollectionInstruments}}
                     </td>
                   </tr>
                 </tbody>
               </table>
               <p>Sample loaded: {{sample.ingestDateTime}}</p>
```

response_operations_ui/controllers/collection_exercise_controllers.py
emilioward: No it doesn't return any response probably best just to return true

response_operations_ui/views/collection_exercise.py
emilioward: ah ok I just mimic'd something that was used in the survey.py file, i'll change that

response_operations_ui/views/collection_exercise.py
```diff
@@ -21,13 +21,13 @@ def get_success_message(success_key):
     return {
         'sample_removed_success': "Sample removed",
         'sample_loaded_success': "Sample successfully loaded"
-    }[success_key]
+    }.get(success_key, None)
```
jcox-dev: `None` is already the default for `dict.get()`

PR #161, opened 2018-05-21
add trading as in ru search result

tej99:
The reporting unit trading as name should be shown as well the business name. This pr adds a trading as column to the reporting units search results page.

You will need the ras party changes for the trading as names to show: https://github.com/ONSdigital/ras-party/pull/131

### DIFF
--- a/response_operations_ui/templates/reporting-units.html
+++ b/response_operations_ui/templates/reporting-units.html
@@ -40,6 +40,9 @@
                         <th scope="col" class="table--header--cell">
                             Name
                         </th>
+                        <th scope="col" class="table--header--cell">
+                            Trading as
+                        </th>
                     </tr>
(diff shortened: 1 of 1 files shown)


response_operations_ui/templates/reporting-units.html
```diff
@@ -40,18 +40,24 @@ <h1 class="jupiter">Reporting units</h1>
                         <th scope="col" class="table--header--cell">
                             Name
                         </th>
+                        <th scope="col" class="table--header--cell">
+                            Trading as
```
gemmai95: Do we still want this to appear if there is no trading as?

tej99: Yeah should just show up empty in the list of business returned.

PR #166, opened 2018-06-04
P2 load sample timeout final

madeye-matt:
Changed the behaviour of the sample upload such that a canned error is displayed in the sample section when an error occurs during upload.  A link may be clicked that will open a plain text error report in a new tab


madeye-matt:
[P2 Load sample timesout and doesn't generate samplelink record in collectionexercise](https://trello.com/c/GE1Rcz6Q/82-p2-load-sample-timesout-and-doesnt-generate-samplelink-record-in-collectionexercise)

response_operations_ui/controllers/sample_controllers.py
```diff
@@ -16,7 +16,7 @@ def upload_sample(short_name, period, file):
 
     survey_type = 'B'
     url = f'{app.config["SAMPLE_URL"]}/samples/{survey_type}/fileupload'
-    response = requests.post(url=url, auth=app.config['SAMPLE_AUTH'], files={'file': file})
+    response = requests.post(url=url, auth=app.config['SAMPLE_AUTH'], files={'file': (file.filename, file)})
```
JamesGardiner: This isn't actually posting a file, it's posting a FileStorage object. You may want to alter the part where the FileStorage object is accessed and us it's `.read()` method to read it as binary data. This will play nicely with requests content-header settings.

From the requests docs:
`Warning
It is strongly recommended that you open files in binary mode. This is because Requests may attempt to provide the Content-Length header for you, and if it does this value will be set to the number of bytes in the file. Errors may occur if you open the file in text mode.`

response_operations_ui/controllers/sample_controllers.py
```diff
@@ -33,3 +33,20 @@ def upload_sample(short_name, period, file):
                  sample_id=sample_summary['id'])
 
     return sample_summary
+
+
+def get_sample_summary(summary_uuid):
+    logger.debug('Getting sample summary', summary_uuid=summary_uuid)
```
JamesGardiner: Maybe use structlog's wrapped logger here? Not sure if this is a pattern in RAS or not... but it's used in SDX and EQ.

response_operations_ui/controllers/sample_controllers.py
```diff
@@ -33,3 +33,20 @@ def upload_sample(short_name, period, file):
...
+
+    url = f'{app.config["SAMPLE_URL"]}/samples/samplesummary/{summary_uuid}'
+    response = requests.get(url, auth=app.config['SAMPLE_AUTH'])
+
+    try:
+        response.raise_for_status()
+    except (HTTPError, RequestException):
```
JamesGardiner: We should handle 4xx and 5xx errors differently.

response_operations_ui/views/samples.py
```diff
@@ -0,0 +1,17 @@
+from flask import Blueprint, render_template, make_response
+from flask_login import login_required
+from response_operations_ui.controllers import sample_controllers
```
JamesGardiner: There should be a space between the 2nd and 3rd imports (one is a package, the other is a local module)

response_operations_ui/controllers/sample_controllers.py
jcox-dev: Frontstage uses it in most of its modules.

andrewmil: This is using the wrapped logger?
line 11 of this file `logger = wrap_logger(logging.getLogger(__name__))`

response_operations_ui/controllers/sample_controllers.py
andrewmil: If we're wanting to do this it might be a wider piece of work across the repo/multiple repos? We probably want to be consistent and there's a lot of places where we're not doing this

response_operations_ui/controllers/sample_controllers.py
JamesGardiner: Ah that makes it an easier change then 👍 cheers @andrewmil

response_operations_ui/controllers/sample_controllers.py
JamesGardiner: Yep, i've asked @nicprice for RAD's views on this

jcox-dev: @JamesGardiner worth raising a TBL for this?

response_operations_ui/views/collection_exercise.py
```diff
@@ -278,7 +279,7 @@ def _validate_sample():
         file = request.files['sampleFile']
         if not str.endswith(file.filename, '.csv'):
             logger.debug('Invalid file format uploaded', filename=file.filename)
-            error = 'Invalid file format'
+            error = 'invalid_file_format'
     else:
         logger.debug('No file uploaded')
         error = 'File not uploaded'
```
jcox-dev: @madeye-matt Does this need updating to match the new underscore format?

benjefferies:
@madeye-matt is this also going to be closed?

benjefferies:
Needs to be revisited after a review from designers

PR #167, opened 2018-06-06
SM123 close conversation

dcdarrell9:
## What has been changed
- Backend functionality has been added to [this pr in secure-message.](https://github.com/ONSdigital/ras-secure-message/pull/207)

- Added functionality for a thread to be closed and re-opened.

- When viewing a conversation there should be the option to close a conversation. This should take you to the close-conversation speedbump page which after confirming will take you back to the conversation list page with a flash message to go back to the message.

- When viewing a closed conversation you will have the option to re-open, after doing so should take you back to the conversation list page again with a flash message.

## Review
- Check that tests pass and functionality works. 

## Future
- Open/Closed tab with list of threads for each status.
- R-Ops/Frontstage to handle any errors on edge cases.


response_operations_ui/controllers/message_controllers.py
```diff
@@ -106,6 +106,34 @@ def remove_unread_label(message_id):
...
+    url = f"{current_app.config['SECURE_MESSAGE_URL']}/v2/threads/{thread_id}"
+    data = {"is_closed": True}
+
+    logger.debug("Adding closed conversation label", thread_id=thread_id)
+    response = requests.patch(url, headers={"Authorization": _get_jwt(), "Content-Type": "application/json"}, json=data)
+
+    try:
+        response.raise_for_status()
+        logger.debug("Successfully added closed conversation label", thread_id=thread_id)
+    except HTTPError:
```
JamesGardiner: I've raised an issue on other PRs around how we should be treating error codes differently (i.e. 4xx vs 5xx). I won't stop the PR on this but it's something to be aware of.

response_operations_ui/views/messages.py
```diff
@@ -195,6 +199,38 @@ def view_selected_survey(selected_survey):
...
 
 
+@messages_bp.route('/close-conversation', methods=['GET', 'POST'])
+@login_required
+def close_conversation():
+    try:
+        messages = session['messages']
+        selected_survey = session['messages_survey_selection']
+        thread_id = messages[0]['thread_id']
+    except KeyError:
```
JamesGardiner: Should these be in different try/except blocks? I'm guessing that it doesn't matter which of the three dict access calls throw KeyError which is why it is written as one block?

tests/views/test_message.py
```diff
@@ -577,3 +577,62 @@ def test_get_to_id_index_error(self):
...
+        mock_request.get(shortname_url + "/ASHE", json=ashe_info)
+        mock_request.get(url_send_message + '/count', json={"total": 1}, status_code=200)
+        mock_request.get(url_get_threads_list, json=thread_list)
+
+        response = self.app.post("/messages/close-conversation?reopen_conversation=True",
+                                 follow_redirects=True)
+
+        self.assertEqual(200, response.status_code)
+        self.assertIn("Conversation re-opened.".encode(), response.data)
+        self.assertIn("Ashe Messages".encode(), response.data)
```
JamesGardiner: New unit tests for new functions in response_operations_ui/controllers/message_controllers.py need to be added. The integration tests above are fine for the routes.

response_operations_ui/controllers/message_controllers.py
dcdarrell9: Fair point, how should we be doing this? I'm guessing it's doing the same thing everywhere?

JamesGardiner: Will be picked up in another piece of work by RAD team.

response_operations_ui/views/messages.py
JamesGardiner: Happy with the explanation for this 👍

tests/views/test_message.py
JamesGardiner: Done in e8231df 👍

response_operations_ui/views/messages.py
```diff
@@ -195,6 +199,38 @@ def view_selected_survey(selected_survey):
                                response_error=True)
 
 
+@messages_bp.route('/close-conversation', methods=['GET', 'POST'])
+@login_required
+def close_conversation():
+    try:
+        messages = session['messages']
+        selected_survey = session['messages_survey_selection']
+        thread_id = messages[0]['thread_id']
```
ajmaddaford: We shouldn't be using the session to store state across sequential requests like this. If you put `close-conversation` on `/threads/<thread_id>` - `/threads/<thread_id>/close-conversation` you'd have access to the thread id for the conversation being closed.

response_operations_ui/templates/conversation-view.html
```diff
@@ -97,11 +97,22 @@ <h1 class="saturn" name="page-messages-title">{{ subject }}</h1>
...
             </form>
-          <a href="{{ url_for('messages_bp.view_selected_survey', selected_survey=selected_survey, page=page) }}">< Back</a>
         </div>
+        {% else %}
+        <div class="panel panel--simple panel--info">
+            <div class="panel__body">
+                <p class="mars">This conversation was closed on {{ thread_data.closed_at }} by {{ thread_data.closed_by }}</p>
+            </div>
+        </div>
+        <form action="{{ url_for('messages_bp.close_conversation', thread_id=messages[0].thread_id, selected_survey=selected_survey, reopen_conversation=True) }}" method="post">
```
ajmaddaford: This form action should be on the view conversation route, rather than the close conversation route, as that is where the action takes place.

Also, selected_survey shouldn't be needed as it's handled by the conversation listing view.

response_operations_ui/views/messages.py
```diff
@@ -195,6 +199,39 @@ def view_selected_survey(selected_survey):
...
+    refined_thread = [_refine(message) for message in reversed(thread_conversation['messages'])]
+    try:
+        selected_survey = session['messages_survey_selection']
+    except KeyError:
+        logger.exception('Session unavailable')
+        selected_survey = None
+
+    if request.method == 'POST':
+        if request.args.get('reopen_conversation'):
+            message_controllers.update_close_conversation_label(thread_id=thread_id, status=False)
```
ajmaddaford: If we're re-opening a conversation we don't need to have loaded the full conversation, so we should try and do this before loading all of the messages.

response_operations_ui/templates/close-conversation.html
```diff
@@ -0,0 +1,46 @@
...
+        <dd class="">
+          {{respondent}}
+        </dd>
+      </dl>
+
+      <form action="{{ url_for('messages_bp.close_conversation', thread_id=thread_id, page=page) }}" method="post">
+        <input type="hidden" name="closed" value="True" />
+        <button class="btn btn--primary u-mt-m">Close conversation</button>
+      </form>
+      <a href="{{ url_for('messages_bp.view_conversation', thread_id=thread_id, page=page) }}#latest-message">Cancel</a>
```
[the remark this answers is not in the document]
dcdarrell9: I think we do need page in this view, as the '< Back' link in the thread view is reliant on knowing which page the user was on, otherwise they'd go back to the first page of pagination, just tried without it being set and it didn't work, added in and worked 👍

tests/views/test_message.py
```diff
@@ -577,3 +577,82 @@ def test_get_to_id_index_error(self):
...
+
+        self.assertEqual(200, response.status_code)
+        self.assertIn("Subject".encode(), response.data)
+        self.assertIn("Business".encode(), response.data)
+        self.assertIn("Reference".encode(), response.data)
+        self.assertIn("Respondent".encode(), response.data)
+
+    @requests_mock.mock()
+    @patch('response_operations_ui.controllers.message_controllers._get_jwt')
+    def test_post_close_conversation_add_closed_status(self, mock_request, mock_get_jwt):
```
ajmaddaford: I think we should rename this to `test_close_conversation`. The closed status isn't part of response ops - it's part of the api implementation - so we should avoid it becoming part of the repsonse ops tests.

tests/views/test_message.py
```diff
@@ -577,3 +577,82 @@ def test_get_to_id_index_error(self):
...
+            session['messages_survey_selection'] = 'Ashe'
+        mock_get_jwt.return_value = "blah"
+        mock_request.get(url_get_thread, json=thread_json)
+        mock_request.get(url_get_surveys_list, json=survey_list)
+        mock_request.patch(url_get_thread, json=thread_json, status_code=500)
+        mock_request.get(shortname_url + "/ASHE", json=ashe_info)
+        mock_request.get(url_send_message + '/count', json={"total": 1}, status_code=200)
+        mock_request.get(url_get_threads_list, json=thread_list)
+
+        response = self.app.post("/messages/threads/fb0e79bd-e132-4f4f-a7fd-5e8c6b41b9af/close-conversation",
```
ajmaddaford: Where does `url_get_thread` get called when doing the post?

tests/views/test_message.py
```diff
@@ -577,3 +577,82 @@ def test_get_to_id_index_error(self):
...
+                                 follow_redirects=True)
+
+        self.assertEqual(500, response.status_code)
+        self.assertIn("Error 500 - Server error".encode(), response.data)
+
+    @requests_mock.mock()
+    @patch('response_operations_ui.controllers.message_controllers._get_jwt')
+    def test_post_close_conversation_remove_closed_status(self, mock_request, mock_get_jwt):
```
ajmaddaford: This would be better named `test_reopen_conversation`

tests/views/test_message.py
dcdarrell9: Doesn't look like it is, removed it and it's fine 👍

response_operations_ui/views/messages.py
```diff
@@ -195,6 +213,28 @@ def view_selected_survey(selected_survey):
                                response_error=True)
 
 
+@messages_bp.route('/threads/<thread_id>/close-conversation', methods=['GET', 'POST'])
```
atrozzo: I m not sure I agree with the use of any kind of verb in a uri . the resource should resource based an dnot action based. Please check this nice example : https://stackoverflow.com/questions/18233632/rest-uri-design-for-changing-the-status-for-resource

dcdarrell9: So, this route brings up the close-conversation speedbump page which lets the user confirm that they want to close and it contains some message data (ref, id etc) so not sure what this could be changed to

atrozzo: ok just realised it already wrongly set up like this elsewhere. Let s keep consistent and leave it like it is then.

PR #173, opened 2018-06-13
Reload templates without restarting

JamesGardiner:
# Motivation and Context
Autoreloads templates when running debug mode. Developers will see changes to html templates without having to restart a running server.

# What has changed
Added new config to config.py to set auto reloading to true and to enable more verbose output of template loading information.

# How to test?
- Run the server either under ras-rm-docker-dev or locally.
- Make sure the app is running in debug mode.
- Navigate to the sign in page.
- Alter the source code for one of the relevant templates.
- Reload the page, check the change is visible.

### DIFF
--- a/config.py
+++ b/config.py
@@ -104,6 +104,9 @@ class DevelopmentConfig(Config):
     SURVEY_PASSWORD = os.getenv('SURVEY_PASSWORD', 'secret')
     SURVEY_AUTH = (SURVEY_USERNAME, SURVEY_PASSWORD)
 
+    TEMPLATES_AUTO_RELOAD = True
+    EXPLAIN_TEMPLATE_LOADING = True
+
     UAA_SERVICE_URL = os.getenv('UAA_SERVICE_URL', 'http://localhost:9080')
     UAA_CLIENT_ID = os.getenv('UAA_CLIENT_ID', 'response_operations')
     UAA_CLIENT_SECRET = os.getenv('UAA_CLIENT_SECRET', 'password')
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -104,6 +104,9 @@ class DevelopmentConfig(Config):
     SURVEY_PASSWORD = os.getenv('SURVEY_PASSWORD', 'secret')
     SURVEY_AUTH = (SURVEY_USERNAME, SURVEY_PASSWORD)
 
+    TEMPLATES_AUTO_RELOAD = True
+    EXPLAIN_TEMPLATE_LOADING = True
```
jcox-dev: Do we want EXPLAIN_TEMPLATE_LOADING on by default? It does tend to flood logs.

JamesGardiner: Hmm, good point. It can be useful if there are issues around template loading. I'll conduct a straw poll tomorrow and see what people prefer.

jcox-dev:
[Mount Python services using Volumes in docker containers (8)](https://trello.com/c/4O9axOVO/155-mount-python-services-using-volumes-in-docker-containers-8)

config.py
ajmaddaford: Is `TEMPLATES_AUTO_RELOAD` needed? See https://github.com/pallets/flask/issues/1907 and https://github.com/pyvec/elsa/issues/44

Pipfile
```diff
@@ -7,7 +7,7 @@ url = "https://pypi.python.org/simple"
 python_version = "3.6"
 
 [packages]
-Flask = "==0.12.2"
+Flask = "==0.12"
```
jcox-dev: 👍

PR #178, opened 2018-06-19
[TBL] Calls party direct when searching reporting units

jcox-dev:
# Motivation and Context

Removes the call to backstage for searching for reporting units.

# What has changed

Changes the `search_reporting_units` method in reporting_units_controllers.py to hit the party service directly.

# How to test?

Tail the logs for backstage, party service. Hit the `/reporting-units/` page and search. Make sure backstage doesn't get the call, but party service does.


jcox-dev:

PR #179, opened 2018-06-19
[TBL] Call party service direct to change enrolment status

jcox-dev:
# Motivation and Context

Removes the call to backstage for changing enrolment status.

# What has changed

Changes the `change_enrolment_status` method in reporting_units_controllers.py to hit the party service directly.

# How to test?

Tail the logs for backstage, party service. Hit the `/reporting-units/<ru_ref>/change-enrolment-status` page and POST. Make sure backstage doesn't get the call, but party service does.


jcox-dev:

response_operations_ui/controllers/reporting_units_controllers.py
```diff
@@ -23,18 +23,20 @@ def search_reporting_units(query):
...
     enrolment_json = {
         'respondent_id': respondent_id,
         'business_id': business_id,
         'survey_id': survey_id,
         'change_flag': change_flag
     }
-    response = requests.put(url, json=enrolment_json)
+    response = requests.put(url, json=enrolment_json, auth=app.config['PARTY_AUTH'])
 
     if response.status_code != 200:
```
tej99: raise_for_status?

PR #181, opened 2018-06-19
[TBL] Call party direct when resending verification email

jcox-dev:
# Motivation and Context

Removes the call to backstage for resending verification email.

# What has changed

Changes the `resend_verification_email` method in reporting_units_controllers.py to hit the party service directly.

# How to test?

Tail the logs for backstage, party service. Hit the `/reporting-units/resend_verification/<ru_ref>/<party_id>` page. Make sure backstage doesn't get the call, but party service does.


jcox-dev:

response_operations_ui/controllers/reporting_units_controllers.py
```diff
@@ -56,8 +56,8 @@ def generate_new_enrolment_code(collection_exercise_id, ru_ref):
 
 def resend_verification_email(party_id):
     logger.debug('Re-sending verification email', party_id=party_id)
-    url = f'{app.config["BACKSTAGE_API_URL"]}/v1/reporting-unit/resend-verification-email/{party_id}'
-    response = requests.post(url)
+    url = f'{app.config["PARTY_URL"]}/party-api/v1/resend-verification-email/{party_id}'
+    response = requests.get(url, auth=app.config['PARTY_AUTH'])
```
tej99: Out of scope for this change but a get action feels wrong for this api in party. Possibly need to raise a TBL to change this?

response_operations_ui/controllers/reporting_units_controllers.py
```diff
@@ -56,8 +56,8 @@ def generate_new_enrolment_code(collection_exercise_id, ru_ref):
...
 
     if response.status_code != 200:
```
tej99: Maybe change to raise_for_status if that's the pattern we're following in front stage

response_operations_ui/controllers/reporting_units_controllers.py
jcox-dev: @tej99 tech debt item created for this to be considered 👍

PR #184, opened 2018-06-21
Remove unused test URLs and use raise for status pattern

jcox-dev:
# Motivation and Context
Remove unused code and ensure all of the reporting unit controller uses the same raise_for_status pattern for error handling. 

# How to test?
Run the unit and acceptance tests for 🍰


jcox-dev:
@insacuri perhaps the test/ dir is excluded?

PR #185, opened 2018-06-21
[TBL] Link and unlink CI's and CE's directly

jcox-dev:
# Motivation and Context

Removes the call to backstage for linking and unlinking collection instruments.

# What has changed

Changes the links/unlink method in collection_exercise_controllers.py to hit the collection instrument service directly.

This also tidies up the test module to use predictable variables to use across the test methods.

# How to test?

Tail the logs for backstage, collection instrument service. Go to a collection exercise's detail page and select / deselect a collection instrument. Make sure backstage doesn't get the call, but collection instrument service does.


jcox-dev:

PR #187, opened 2018-06-21
Changing env variable from RM_REPORT_SERVICE to REPORT_URL and adding…

pricem14pc:
… fwd slash

# Motivation and Context
Bug

# What has changed
See commit

# How to test?
Download response status report

# Links
n/a

n/a

### DIFF
--- a/response_operations_ui/controllers/collection_exercise_controllers.py
+++ b/response_operations_ui/controllers/collection_exercise_controllers.py
@@ -18,8 +18,8 @@ def download_report(collection_exercise_id, survey_id):
     )
 
     url = (
-        f"{app.config['RM_REPORT_SERVICE']}"
-        f"reporting-api/v1/response-chasing/download-report/"
+        f"{app.config['REPORT_URL']}"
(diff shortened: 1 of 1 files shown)


response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -18,8 +18,8 @@ def download_report(collection_exercise_id, survey_id):
     )
 
     url = (
-        f"{app.config['RM_REPORT_SERVICE']}"
-        f"reporting-api/v1/response-chasing/download-report/"
+        f"{app.config['REPORT_URL']}"
+        f"/reporting-api/v1/response-chasing/download-report/"
```
benjefferies: We should write an acceptance test for this so it won't slip through again

tej99: I agree but that will need a little more work, I don't think the reporting service is part of our CI pipeline or part of docker dev. I feel like this bug fix PR should still go in without it, providing the TBL tickets have been raised and prioritised.

PR #188, opened 2018-06-22
remove email from url

tej99:
# Motivation and Context

User emails should not be passed in url as they will be exposed in service logs

# What has changed

Removed unnecessarily passing user email in url

# How to test?

Check you can still re send verification email for pending account.

There is a known bug for resending verification email for email changes, the changes in this PR did not create and will not fix this bug. 

# Links


tej99:
[Party service email verification (enrolment) spike](https://trello.com/c/mbpC8vqa/89-party-service-email-verification-enrolment-spike)

PR #189, opened 2018-06-22
[TBL] Get survey details direct from survey service

jcox-dev:
# Motivation and Context

Removes the proxying call to backstage and ports the logic which builds a survey's details to this service.

# What has changed

Porting the get_survey from backstage had a knock-on effect for the various views (mostly at a test level) that rely on building and formatting everything related to a survey and its details. In other words there was more than just the survey controller that needed updating :)

# How to test?


jcox-dev:

PR #191, opened 2018-06-25
[TBL] Update details direct with survey service

jcox-dev:
# Motivation and Context

# What has changed

Change to update a survey's details direct through the survey service rather than proxying through the backstage service.

# How to test?


jcox-dev:

PR #192, opened 2018-06-25
Output time as BST for secure message

shahi645:
# Motivation and Context
The Business have been informed that the time stamp for secure messaging doesn't reflect BST.

# What has changed
Time displayed for users in secure message is BST

# How to test?
Run RAS/RM, send a message to an internal user, the message should display the correct time

### DIFF
--- a/tests/test_dates.py
+++ b/tests/test_dates.py
@@ -1,31 +1,30 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timezone, date
 import unittest
+from unittest.mock import patch
 
-from response_operations_ui.common.dates import get_formatted_date
+from response_operations_ui.common.dates import get_formatted_date, convert_to_bst
 
 
 class TestDates(unittest.TestCase):
 
-    def test_get_formatted_date_today(self):
-        today = datetime.now()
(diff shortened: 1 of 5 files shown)


shahi645:
@gemmai95 the functionality will work, but tests will need updating

gemmai95:
@shahi645 Can we note this somewhere?

tests/test_dates.py
```diff
@@ -37,3 +39,10 @@ def test_get_formatted_date_returns_malformed_dates(self):
         # Strings not in the correct format or invalid dates should be returned as given and a exception logged
         self.assertEqual(get_formatted_date(''), '')
         self.assertEqual(get_formatted_date('1999-12-32 23:59:59'), '1999-12-32 23:59:59')
+
+    def test_get_current_date_bst(self):
+        # Test must be updated for last Sunday in March/October to reflect BST
```
gavinleeedwards: Would it be better to work out whether we're in GMT or BST based on the current date (i.e. BST if the current date is between the last Sunday in March and the last Sun in Oct, otherwise GMT)?  It would save the tests having to be changed every 6 months.

shahi645: changes coming, sorry should hve put a label on this

PR #193, opened 2018-06-25
[TBL] Get collection exercise details direct

jcox-dev:
# Motivation and Context

# What has changed

Change to get a collection exercise's details direct through the various services rather than proxying through the backstage service.

Big changes to the tests to accommodate the changes to how a ce details will be built.

# How to test?


jcox-dev:

PR #196, opened 2018-06-26
Mock the ApiError handler to prevent false positives

jcox-dev:
# Motivation and Context

This mocks the Flask error handler so we can validate that an ApiError was raised rather than asserting false positives from other 500s.

# What has changed

Creates a new ViewTestCase class that the other test cases are then derived from which mocks the error handler for ApiError at setUp time.

# How to test?

Run the unit tests and check that we're testing everything accurately ✨🍰 ✨

### DIFF
--- a/config.py
+++ b/config.py
@@ -134,3 +134,5 @@ class TestingConfig(DevelopmentConfig):
     WTF_CSRF_ENABLED = False
     SESSION_TYPE = "filesystem"
     SESSION_PERMANENT = False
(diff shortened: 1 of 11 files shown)


jcox-dev:
The collection exercise tests will be updated once #193 is merged in (which this work is extended from).

gemmai95:
@jcox-dev Looks fine but I can't approve it until I can run the acceptance tests against it.

PR #198, opened 2018-06-27
[TBL] Update a collection exercise direct

jcox-dev:
# Motivation and Context

# What has changed

Change to update a collection exercise's details direct through the collection exercise service rather than proxying through the backstage service.

This also updates the form logic to fix an issue where validation would fail unless the user entered a new period.

# How to test?

### DIFF
--- a/response_operations_ui/forms.py
+++ b/response_operations_ui/forms.py
@@ -75,11 +75,14 @@ class EditCollectionExerciseDetailsForm(FlaskForm):
     @staticmethod
     def validate_period(form, field):
         hidden_survey_id = form.hidden_survey_id.data
+        hidden_ce_id = form.collection_exercise_id.data
         ce_details = collection_exercise_controllers.get_collection_exercises_by_survey(hidden_survey_id)
         inputted_period = field.data
         if inputted_period is None:
             raise ValidationError('Please enter numbers only for the period')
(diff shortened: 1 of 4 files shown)


jcox-dev:
Coverage should be able to be increased once #196 is merged in.

jcox-dev:

response_operations_ui/forms.py
```diff
@@ -75,11 +75,14 @@ class EditCollectionExerciseDetailsForm(FlaskForm):
     @staticmethod
     def validate_period(form, field):
         hidden_survey_id = form.hidden_survey_id.data
+        hidden_ce_id = form.collection_exercise_id.data
         ce_details = collection_exercise_controllers.get_collection_exercises_by_survey(hidden_survey_id)
         inputted_period = field.data
         if inputted_period is None:
             raise ValidationError('Please enter numbers only for the period')
         for ce in ce_details:
+            if ce['id'] == str(hidden_ce_id):
```
benjefferies: Not sure what value this check adds but I may be being stupid

jcox-dev: @benjefferies ah, this is to skip over checking against its own details. Prior to this change, you would be unable to update the user description of a CE if the period remained the same.

PR #207, opened 2018-08-01
Fix verification email link

gavinleeedwards:
# Motivation and Context
Fixes a bug introduced in the latest release where a 500 error was generated if an internal user attempted to resend a verification email to a respondent that had not yet verified their account.

# What has changed
- Slight code change

# How to test?
- Test all different scenarios for resending a verification email:

- For a new respondent that has not yet verified
- If a respondent has verified but wants to change their email address
- For a new respondent that has not yet verified AND wants to change their email address from the one they created the account with


gavinleeedwards:
Rather than there being a missing test, the test data was incorrect and wasn't reflecting how it appears in the system.  Removed the 'pendingEmailAddress' key from respondent data and all tests pass.

PR #209, opened 2018-08-03
Stop redirecting to error pages in ROps

MIslam92:
Redirecting to an error page caused us issues previously when it came to tracking/diagnosing problems.
This should now allow us to receive an URL to know exactly where an issue has occurred rather then just the error page.

Card: https://trello.com/c/odV0qjxk/272-stop-redirecting-to-error-pages-in-rops-s

### How to Review
Ensure it now renders the correct error pages and unnecessary redirects are removed.
Just `raise exception` somewhere in the code to test it displays the error pages.
Check that all functionality works as per norm and tests pass.

### DIFF
--- a/response_operations_ui/views/errors.py
+++ b/response_operations_ui/views/errors.py
@@ -1,6 +1,6 @@
 import logging
 
-from flask import Blueprint, flash, redirect, render_template, url_for
+from flask import Blueprint, flash, redirect, render_template, url_for, request
 from structlog import wrap_logger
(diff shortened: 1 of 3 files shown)


response_operations_ui/views/errors.py
```diff
@@ -10,7 +10,6 @@
 error_bp = Blueprint('error_bp', __name__, template_folder='templates/errors')
 
 
-@error_bp.route('/500', methods=['GET'])
 def server_error_page():
```
jcox-dev: This function can probably be deleted if it's not being used.

PR #211, opened 2018-08-07
Sus003 view case details

emilioward:
# Motivation and Context

This change adds the page to be able to view the social case details along with unit tests

# What has changed

Added new UI for page along with CSS and banner
New controller to access sample attributes
Unit tests

# How to test?

Run unit tests against the code, wont able to view the page as the story intended until https://github.com/ONSdigital/response-operations-ui/tree/find_case_by_postcode is merged in 

You can test this page by going to RM-tools/social-test-setup and running `make setup-and-execute` then once its completed go into the case table and grab a case_id then put that into the URL for response ops (http://localhost:8085/social/case/<CASE_ID>) the page will render with the relevant details

# Links


emilioward:
[SUS003: View Case Details (8)](https://trello.com/c/oLmDvgmZ/188-sus003-view-case-details-8)

response_operations_ui/static/css/override.css
```diff
@@ -374,3 +373,63 @@ button{
...
+
+.results-list-ref {
+    margin-right: 1rem;
+}
+
+.results-list-postcode {
+    margin-right: 1rem;
+}
+
+.results-list-address {
```
benjefferies: Shall we remove classes we don't define anything for?

response_operations_ui/static/css/override.css
```diff
@@ -374,3 +373,63 @@ button{
...
+.case-search-input2 {
+    width: 8rem;
+}
+
+.case-search-submit2 {
+    padding: 0.5rem 1rem;
+    margin-left: 0.5rem;
+}
+
+.social-case-search-results {
```
benjefferies: Shall we remove classes we don't define anything for?

response_operations_ui/views/social_view_case_details.py
```diff
@@ -0,0 +1,25 @@
...
+social_bp = Blueprint('social_bp', __name__,
+                      static_folder='static', template_folder='templates')
+
+
+@social_bp.route('/case/<case_id>', methods=['GET'])
+@login_required
+def view_social_case_details(case_id):
+    form = SearchForm()  # TODO - This needs to be edited and expanded on the work done for 'Find case by postcode'
+
+    social_case = case_controller.get_case_by_id(case_id)
```
damien-lloyd: How are we validating that this is case is indeed a social case? What controls do we have in place to ensure the case retrieved is a social case?

liamtoozer: That's a good point, however there's currently no design in place to handle a 'case not found' scenario directly in the UI. We're using get_case_by_id: https://github.com/ONSdigital/response-operations-ui/blob/0a3fb6016e21672993821e2a1c9fca79a4cfa2f7/response_operations_ui/controllers/case_controller.py#L13
This should raise an exception if the case is not found by UUID, which is provided directly in the URL or from the post code search result.

emilioward: Currently if you put a case_id for a business into the url the page will render but no data will be displayed as the sample attributes are different

damien-lloyd: @emilioward @liamtoozer Maybe we should only enable this functionality for social cases.

response_operations_ui/views/social_view_case_details.py
```diff
@@ -0,0 +1,25 @@
...
+    sample_attributes = sample_controllers.get_sample_attributes(social_case['sampleUnitId'])
+
+    social_case['caseGroup']['caseGroupStatus'] = map_social_case_status(social_case['caseGroup']['caseGroupStatus'])
```
jcox-dev: This could be `mapped_status = ...` or similar as it's only being passed in as a parameter of the template.

Then in the template `{{ status }}`.

response_operations_ui/views/social_view_case_details.py
emilioward: It depends on what we want to present back to the user

jcox-dev:
Looks good when tested locally against the social setup scripts.

image

tests/views/test_social.py
```diff
@@ -4,7 +4,7 @@
 import requests_mock
 
 from config import TestingConfig
-from response_operations_ui.views.social import format_address_for_results
+from response_operations_ui.views.social.social_case_search import format_address_for_results
 from tests.views import ViewTestCase
```
jcox-dev: Second newline missing after imports. 🎨

response_operations_ui/views/social/social_view_case_details.py
```diff
@@ -0,0 +1,19 @@
+import logging
+from flask import render_template
+from flask_login import login_required
+from structlog import wrap_logger
+from response_operations_ui.controllers import case_controller, sample_controllers
+from response_operations_ui.common.mappers import map_social_case_status
+
```
jcox-dev: Second newline missing after imports. 🎨

response_operations_ui/views/social/social_view_case_details.py
```diff
@@ -0,0 +1,19 @@
+import logging
+from flask import render_template
```
jcox-dev: Linebreak between import groups.

```
# stdlib

# third party

# project
```

response_operations_ui/views/social/social_case_search.py
```diff
@@ -1,19 +1,16 @@
 import logging
 
-from flask import Blueprint, render_template, request
+from flask import render_template, request
 from flask_login import login_required
 from structlog import wrap_logger
 
 from response_operations_ui.controllers.case_controller import get_cases_by_sample_unit_id
 from response_operations_ui.controllers.sample_controllers import search_samples_by_postcode
```
jcox-dev: Second newline missing after imports. 🎨

response_operations_ui/views/social/__init__.py
```diff
@@ -0,0 +1,10 @@
+from flask import Blueprint
+
+from response_operations_ui.views.social.social_case_search import social_case_search
+from response_operations_ui.views.social.social_view_case_details import view_social_case_details
+
```
jcox-dev: Second newline missing after imports. 🎨

PR #213, opened 2018-08-10
Added test to cover CloudFoundry set up

MIslam92:
We were missing a test for when we deploy to a cloud foundry env and using its redis instance rather then the default one. 

Card: https://trello.com/c/kkO8qp4j/308-improve-rops-test-coverage-s 

### How to review
Just check all tests pass and the previous untested code has been covered.

### DIFF
--- a/tests/views/test_create_app.py
+++ b/tests/views/test_create_app.py
@@ -16,11 +16,11 @@ class TestCreateApp(unittest.TestCase):
         mock_cf.redis.return_value = 'REDIS_SERVICE'
 
         mock_cf.redis_service = {
-            'host': 'test',
-            'port': 'test'
+            'host': 'test_host',
+            'port': 'test_port'
         }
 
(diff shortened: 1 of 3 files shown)


tests/views/test_create_app.py
```diff
@@ -0,0 +1,26 @@
...
+        test_app = create_app('TestingConfig')
+
+        self.assertEqual(test_app.config['REDIS_HOST'], 'localhost')
+
+    @patch('response_operations_ui.cf')
+    def test_create_app_with_cf(self, mock_cf):
+        mock_cf.redis.return_value = 'REDIS_SERVICE'
+
+        mock_cf.redis_service = {
+            'host': 'test',
```
jcox-dev: Could be explicit to avoid false positives: 

```
    'host': 'test_host',
    'port': 'test_port'

...
```

PR #215, opened 2018-08-13
Improve ROps test coverage

MIslam92:
This is a resurrected PR where the initial one passed locally/travis etc. but failed when it came to concourse somehow

We were missing a test for when we deploy to a cloud foundry env and using its Redis instance rather then the default one.

Card: https://trello.com/c/kkO8qp4j/308-improve-rops-test-coverage-s

### How to review
Just check all tests pass and the previous untested code has been covered.

If you have access to the `dev` space in Cloud Foundry you can push this branch up and check that the app builds successfully on there.

### DIFF
--- a/response_operations_ui/__init__.py
+++ b/response_operations_ui/__init__.py
@@ -13,6 +13,8 @@ from response_operations_ui.logger_config import logger_initial_config
 from response_operations_ui.user import User
 from response_operations_ui.views import setup_blueprints
 
+cf = ONSCloudFoundry()
+
 
 def create_app(config_name=None):
     app = Flask(__name__)
@@ -26,8 +28,7 @@ def create_app(config_name=None):
     js_min = Bundle('js/*', filters='jsmin', output='minimised/all.min.js')
     assets.register('js_all', js_min)
 
-    config_name = config_name or os.environ.get("APP_SETTINGS", "Config")
-    app_config = f'config.{config_name}'
+    app_config = f'config.{config_name or os.environ.get("APP_SETTINGS", "Config")}'
     app.config.from_object(app_config)
 
     app.url_map.strict_slashes = False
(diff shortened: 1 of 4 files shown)


response_operations_ui/cloud/cloudfoundry.py
```diff
@@ -6,8 +6,10 @@ class ONSCloudFoundry(object):
...
-        return True if self._cf_env.app else False
+    @property
+    def detected(self):
+        return self._cf_env.app
 
-    def redis(self, service_name):
-        return self._cf_env.get_service(name=service_name)
+    @property
+    def redis(self):
+        return self._cf_env.get_service(name='ras-redis')
```
andrewmil: We should probably be able to set the redis service name with an environment variable

response_operations_ui/__init__.py
```diff
@@ -26,16 +27,14 @@ def create_app(config_name=None):
...
+    app_config = f'config.{config_name or os.environ.get("APP_SETTINGS", "Config")}'
     app.config.from_object(app_config)
 
     app.url_map.strict_slashes = False
     app.secret_key = app.config['RESPONSE_OPERATIONS_UI_SECRET']
 
     logger_initial_config(service_name='response-operations-ui', log_level=app.config['LOGGING_LEVEL'])
-    logger = wrap_logger(logging.getLogger(__name__))
-    logger.info('Logger created', log_level=app.config['LOGGING_LEVEL'])
+    logger = logging.getLogger(__name__)
```
andrewmil: why did we remove `wrap_logger` here?

response_operations_ui/__init__.py
```diff
@@ -45,20 +44,19 @@ def create_app(config_name=None):
     def user_loader(user_id):
         return User(user_id)
 
-    if app.config['SESSION_TYPE'] == 'redis':
-        cf = ONSCloudFoundry()
-        # If deploying in cloudfoundry set config to use cf redis instance
-        if cf:
-            redis_service_name = app.config['REDIS_SERVICE']
+    if cf.detected:
+        with app.app_context():
```
andrewmil: What's this? Feels strange that we'd have to use a context here

MIslam92: ah so it works locally but in the cf it kept throwing `RuntimeError: Working outside of application context.`
To overcome this I had to use `app_context`

andrewmil: :+1: Yeah this is a sensible place to use it

Potentially you could just pass the app object into `ONSCloudfoundry(...)` instead but this solution works fine

PR #216, opened 2018-08-13
Log config used and logger level on startup

jcox-dev:


response_operations_ui/__init__.py
```diff
@@ -64,4 +65,6 @@ def user_loader(user_id):
 
     setup_blueprints(app)
 
+    logger.info("App setup complete", config=app_config)
```
insacuri: I don't think we can do this.  Might be worth talking to Dan O'Brian about it.  We're not allowed to log sensitive information and this would log unames/passwords that we use to connect to databases, etc

jcox-dev: No this just logs the *name* of the config object.

insacuri: got an example for me of what it looks like?  I don't see how that'd be useful to anyone.

jcox-dev:
@AdamHawtin oops! Level is a reserved one.

AdamHawtin:
@jcox-dev Now I'm getting `TypeError: _log() got an unexpected keyword argument 'log_level'`

jcox-dev:
@AdamHawtin that should do it. We weren't using structlog so no wonder it complained!

jcox-dev:
[Log out config used in Python services](https://trello.com/c/Tctoorqs/251-log-out-config-used-in-python-services)

PR #220, opened 2018-08-20
Social sample file pilot changes

emilioward:
# Motivation and Context

Due to the format of the social sample file changing we need to update the data being displayed back to the user from searching for a social postcode and viewing the data

# What has changed

Updated templates and unit tests associated with the changes and added new updated social sample test data

# How to test?

Run unit tests and ensure they pass, spin up manually and check info is all correct too 

# Links

### DIFF
--- a/response_operations_ui/templates/social-view-case-details.html
+++ b/response_operations_ui/templates/social-view-case-details.html
@@ -3,7 +3,7 @@
 {% block main %}
 
 <div class="container page__container">
-    <h1 class="jupiter u-mt-m" id="case_ref">{{attributes.Reference}}</h1>
+    <h1 class="jupiter u-mt-m" id="case_ref">{{attributes.TLA}}{{attributes.REFERENCE}}</h1>
     <div class="case-details u-mt-m">
         <dl class="description-list description-list-4" id="case-status">
             <dt class="description-list__name">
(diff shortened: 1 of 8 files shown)


response_operations_ui/templates/social-view-case-details.html
```diff
@@ -16,39 +16,29 @@ <h1 class="jupiter u-mt-m" id="case_ref">{{attributes.Reference}}</h1>
         <section class="social-case-section u-mt-l">
             <h2 class="neptune">Property address</h2>
             <dl class="survey-info info-panel" id="case-attributes">
-                {% if attributes.Prem1 %}
-                <dt class="case-address__prem" id="case-prem1">
-                    {{attributes.Prem1}}
+                {% if attributes.ADDRESS_LINE1 %}
```
jcox-dev: Instead of having to maintain this for every change in a sample file, perhaps we can just have a for-loop?

```jinja2
<dl class="survey-info info-panel" id="case-attributes">
{% for name, attribute in attributes.iteritems() %}
  <dt class="case-address__{{ name|lower }}" id="case-{{ name|lower|replace('_', '-') }}">
    {{ attribute }}
  </dt>
{% endfor %}
</dl>
```

response_operations_ui/templates/social-view-case-details.html
```diff
@@ -3,7 +3,7 @@
 {% block main %}
 
 <div class="container page__container">
-    <h1 class="jupiter u-mt-m" id="case_ref">{{attributes.Reference}}</h1>
+    <h1 class="jupiter u-mt-m" id="case_ref">{{attributes.TLA}}{{attributes.REFERENCE}}</h1>
```
jcox-dev: Might be worth passing the full reference ("LMS0001", for example) as a template variable rather than relying on the attributes.

response_operations_ui/templates/social-view-case-details.html
emilioward: Good idea but then we'd be passing data to the html that doesn't need to be displayed to the user, any idea the best way to filter the data that doesn't need to be presented?

jcox-dev: If we have a list of attribute keys that we only want to use then that can be passed in as a variable and used in the loop instead:

```python
...
return render_template(attributes=sample_attributes['attributes'], 
                       list_of_attributes=["ADDRESS_LINE1", "ADDRESS_LINE2", ...])
```

```jinja2
<dl class="survey-info info-panel" id="case-attributes">
{% for name in list_of_attributes %}
  <dt class="case-address__{{ name|lower }}" id="case-{{ name|lower|replace('_', '-') }}">
    {{ attributes["name"] }}
  </dt>
{% endfor %}
</dl>
```

response_operations_ui/templates/social-view-case-details.html
```diff
@@ -16,41 +16,11 @@ <h1 class="jupiter u-mt-m" id="case_ref">{{attributes.Reference}}</h1>
...
-                    {{attributes.PostTown}}
-                </dt>
-                {% endif %}
-                {% if attributes.Postcode %}
-                <dt class="case-address__prem postcode" id="case-postcode">
-                    {{attributes.Postcode}}
-                </dt>
-                {% endif %}
+            {% for name in displayed_attributes %}
+              <dt class="case-address__{{ name|lower }}" id="case-{{ name|lower|replace('_', '-') }}">
```
damien-lloyd: This could be cleaned up a bit. Could you not add this logic to the Python code instead?

jcox-dev: This has already been cleaned up and (now) only has display logic in the template.

emilioward: Yeah this is a cleaned up version that @jcox-dev recommended, now we don't have to update the template with any new changes and just do them in the views 👍

PR #232, opened 2018-09-03
Dynamically populate secure messaging drop down

ryangrundy7:
# Motivation and Context
At present every time we onboard a new survey we manually have to update the dropdown box in r-ops to reflect the change so secure messaging can be used. The box should be dynamic and call from the survey service to populate it. 

# What has changed
- Secure-message dropdown now gets populated from the survey service
- FDI Grouping happens in Config instead of its own file.
- Added specific FDI create message and messages page tests

# How to test?
- Run r-ops tests and the acceptance tests
- Check the secure-message dropdown works and is populated with every survey
- Set a FDI survey collection exercise to ready to live, send a message and you should be directed to the FDI messages screen. Try with multiple FDI surveys.
- Create a survey using `/surveys/create` and it should be put into the surveys dropdown box
# Links
[Trello Card](https://trello.com/c/Xfc0UL3i/326-dynamically-populate-secure-messaging-drop-down-m)


tests/views/test_message.py
```diff
@@ -509,13 +568,16 @@ def test_conversation_key_error(self, mock_request):
...
-    def test_get_radio_buttons(self):
+    @requests_mock.mock()
+    def test_get_radio_buttons(self, mock_request):
+        mock_request.get(url_get_surveys_list, json=survey_list)
         response = self.client.get("/messages/select-survey")
 
         self.assertEqual(200, response.status_code)
         self.assertIn("Filter messages by survey".encode(), response.data)
         self.assertIn("ASHE".encode(), response.data)
-        self.assertIn("Bricks".encode(), response.data)
```
NickGrantONS: any reason for taking Bricks out?

ryangrundy7: I must have just accidentally took it out without realising and added BRES to it. Can add it back if need be but i don't think its makes a difference

PR #234, opened 2018-09-04
Fix reminder logic in template

tej99:
# Motivation and Context

When ADD_EVENT_DATES_ENABLED is set to false the logic in the template
breaks as it is trying to find reminder2 in the events dict when
it does not exist. I have added a defensive check before so it doesn't
fall over.

# What has changed

# How to test?

ACs should run when ADD_EVENT_DATES_ENABLED is set to false

# Links

### DIFF
--- a/response_operations_ui/templates/collection_exercise/ce-dates-section.html
+++ b/response_operations_ui/templates/collection_exercise/ce-dates-section.html
@@ -104,7 +104,7 @@
           <form action="{{ url_for('collection_exercise_bp.get_create_collection_event_form', short_name=survey.shortName, period=ce.exerciseRef, ce_id=ce.id, tag='reminder')}}" method="get">
             <button class="btn btn--secondary btn--small btn__table-row" id="create-event-date-reminder" type="submit">Add</button>
           </form>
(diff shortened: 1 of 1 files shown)


response_operations_ui/templates/collection_exercise/ce-dates-section.html
```diff
@@ -104,7 +104,7 @@ <h2 class="venus u-pt-m">Dates</h2>
           <form action="{{ url_for('collection_exercise_bp.get_create_collection_event_form', short_name=survey.shortName, period=ce.exerciseRef, ce_id=ce.id, tag='reminder')}}" method="get">
             <button class="btn btn--secondary btn--small btn__table-row" id="create-event-date-reminder" type="submit">Add</button>
           </form>
-        {% elif (not locked or events.reminder.is_in_future) and config.EDIT_EVENT_DATES_ENABLED %}
+        {% elif events.reminder and (not locked or events.reminder.is_in_future) and config.EDIT_EVENT_DATES_ENABLED %}
```
benjefferies: Might be worth pulling this logic out into the route and just passing a boolean such as is_reminder_editable?

PR #235, opened 2018-09-05
Localise times

tej99:
# Motivation and Context

Currently the UI hard codes 'GMT' at the end of each time which is
wrong. The time should be localised before showing to the user, which
is fixed in the PR. Also time zone is not added when updating and
creating events this has been fixed in this PR.

# What has changed

- Localise time before showing to user
- Add timezone before sending to collection exercise for creating/updating events

# How to test?

Ran ATs -> https://github.com/ONSdigital/rasrm-acceptance-tests/pull/173

# Links


tej99:
[UI for adding and updating remaining optional dates](https://trello.com/c/f1n44Kdo/166-ui-for-adding-and-updating-remaining-optional-dates)

PR #240, opened 2018-09-11
Add script to align events and action rules

tej99:
# Motivation and Context

Currently action rule trigger dates and event dates do not match
in various environments. This script updates all action rules and
events. This is needed before giving the ability to moderate CE
events, using response operations, to users.

# What has changed

python script added

# How to test?

run this against your local docker dev environment:
`pipenv run python scripts/align_events_and_rules.py 
    http://0.0.0.0:8145 
    admin secret`

Check all events (except events for CE that are ready for live or live) have been updated and action rules in the database has been updated.

Can run as a cloudfoundry task:
`cf run-task $RESPONSE_OPERATIONS_APP_NAME "python scripts/align_events_and_rules.py $COLLECTION_EXERCISE_URL $BASIC_AUTH_USER $BASIC_AUTH_PASSWORD" --name align-rules-events`

# Links

### DIFF
--- a/scripts/align_events_and_rules.py
+++ b/scripts/align_events_and_rules.py
@@ -1,0 +1,97 @@
+#!/usr/bin/python
+import argparse
+import datetime
+import requests
+from os import abort
+
+from dateutil import tz
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Align collection exercise events and rules')
+    parser.add_argument("url", help="Collection exercise service URL")
+    parser.add_argument("user", help="Basic auth user")
+    parser.add_argument("password", help="Basic auth password")
+    return parser.parse_args()
+
+
+def update_event(collex_id, event_tag, date, url, user, password):
+    path = "/collectionexercises/{id}/events/{tag}".format(id=collex_id, tag=event_tag)
+    response = requests.put(url+path,
+                            data=date,
+                            auth=(user, password),
+                            headers={'content-type': 'text/plain'})
+
+    status_code = response.status_code
+
+    if status_code != 204:
+        detail_text = response.text
+        print("{} <= {} ({})".format(status_code, date, detail_text))
+
+
+def get_collection_exercises(user, password, url):
+    print(url)
+    response = requests.get(url + "/collectionexercises", auth=(user, password))
+
+    status_code = response.status_code
+
+    if status_code == 200:
+        ces = response.json()
+        print("{} <= {} collection exercises retrieved".format(status_code, len(ces)))
+        return ces
+
+    print("{} <= {}".format(status_code, response.text))
+    abort()
+
+
+def is_mandatory_event(event):
+    mandatory_events = ['mps', 'go_live', 'reminder', 'reminder1', 'reminder2']
+    return event['tag'] in mandatory_events
+
+
+def align_events_and_rules(collection_exercises, user, password, url):
+    for collection_exercise in collection_exercises:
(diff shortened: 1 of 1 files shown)


tej99:
[UI for adding and updating remaining optional dates](https://trello.com/c/f1n44Kdo/166-ui-for-adding-and-updating-remaining-optional-dates)

scripts/align_events_and_rules.py
```diff
@@ -0,0 +1,98 @@
...
+import argparse
+import datetime
+from dateutil import tz
+import requests
+from os import abort
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Align collection exercise events and rules')
+    parser.add_argument("url", help="Collection exercise service URL")
```
NickGrantONS: we store this as scheme (e.g. http), host and port in the environment variables, so it might be worth passing these parameters individually, and then you wouldn't have to escape the slashes etc

scripts/align_events_and_rules.py
```diff
@@ -0,0 +1,98 @@
...
+                         url=url,
+                         user=user,
+                         password=password)
+
+
+def change_time_to_9_am(event_timestamp):
+    assert event_timestamp[-1] == 'Z'
+    date_format = '%Y-%m-%dT%H:%M:%S.%f'
+    date = datetime.datetime.strptime(event_timestamp[:-1], date_format)
+    london_timezone = tz.gettz('Europe/London')
```
NickGrantONS: do we want London time or GMT? Presumably we don't want BST.

scripts/align_events_and_rules.py
tej99: Just checked the devtest cf environments and the response-operations-ui config.py file, it looks like we have a `COLLECTION_EXERCISE_URL` environment variable so no need to pass them separately.

scripts/align_events_and_rules.py
tej99: London time, because I need to set all the events to be 9am local time. UTC is worked out by the collection exercise API and stored as UTC.

scripts/align_events_and_rules.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/python
+import argparse
```
MIslam92: import order?

scripts/align_events_and_rules.py
```diff
@@ -0,0 +1,98 @@
...
+import datetime
+from dateutil import tz
+import requests
+from os import abort
```
MIslam92: import order here seems off here?

tej99: Fixed

shahi645:
@tej99 you have some legit codacy issues

scripts/align_events_and_rules.py
shahi645: I was under the impression we store all time as UTC

tej99:
@shahi645 I've checked them I want the changes it's complaining about. This script is standalone so the error about the assert is irrelevant (AssertionError will be thrown which is the behaviour I want)

As for the disabling SSL certificate checks. Most of the environments do not have TLS/SSL (Just prod?) so probably best to disable the checking instead of getting all the way to prod and finding out the script has a problem with verifying the SSL certificates.

scripts/align_events_and_rules.py
tej99: Yep everything is stored as UTC.

tej99:
@shahi645 actually changed my mind I think i'm over thinking this... removed the assert and verify false 😅

PR #250, opened 2018-10-10
Post case event to transition casegroup status

andrewmil:
# Motivation and Context
Previously we were transitioning casegroup status with a call to the `/casegroups/transitions/...` endpoint. This doesn't align with other case lifecycle processes.
We now post a case event to trigger this transition 

# What has changed
Removed code to post to `/casegroups/transitions/...` endpoint and replaced it with a post to case event endpoint.

# How to test?
Check that changing casegroup status in the UI updates the required rows in the case database, 
i.e. casegroup status is updated and the related case is set to `INACTIONABLE`

# Links
[Trello](https://trello.com/c/xkWvMZFk/420-use-case-event-rather-than-case-group-transition-for-change-response-status)

## Associated PR's
[rm-case-service](https://github.com/ONSdigital/rm-case-service/pull/93)
[rm-acceptance-tests](https://github.com/ONSdigital/rasrm-acceptance-tests/pull/196)


response_operations_ui/controllers/case_controller.py
```diff
@@ -24,21 +24,38 @@ def get_case_by_id(case_id):
...
-                 case_group_event=case_group_event)
-    url = f'{app.config["CASE_URL"]}/casegroups/transitions/{collection_exercise_id}/{ru_ref}'
-    response = requests.put(url, auth=app.config['CASE_AUTH'], json={'event': case_group_event})
+def post_case_event(case_id, category, description):
+    logger.debug("Posting case event", case_id=case_id, category=category)
+    url = f'{app.config["CASE_URL"]}/cases/{case_id}/events'
+    case_event = {
+        "category": category,
+        "description": description,
+        "createdBy": "ROPS"
```
shahi645: could be explicit here and say `Response_Ops`. Your call though 🤷‍♂️

shahi645: Also, doesn't this endpoint need the `party_id`?

response_operations_ui/controllers/case_controller.py
```diff
@@ -24,21 +24,38 @@ def get_case_by_id(case_id):
...
+    }
+    response = requests.post(url, auth=app.config['CASE_AUTH'], json=case_event)
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        logger.exception('Error updating case group status', case_id=case_id, category=category)
```
shahi645: Probably would say, error posting case event, should it not be an `Api Error`?

shahi645:
@andrewmil Mostly happy with this, just about to test. Left some really minor comments

response_operations_ui/controllers/case_controller.py
shahi645: ok cool, i guess we have the `case_id` so theres no need for the `party_id`

PR #256, opened 2018-10-17
remove hard coded page and use request referrer

shahi645:
# Motivation and Context
Without JS the hardcoded page wouldn't work. I.e. 5xx's

# What has changed
Use request.referrer

# How to test?
Disable JS and click cancel on change response status page


owencjones:
LGTM.  Long term, might be best to have a `returnPath` variable or something, rather than use the referrer.

PR #258, opened 2018-10-24
Initial commit after changing party to not use sample_unit_type

AndrewTorrance:
# Motivation and Context
Changed party end points to not accept sample unit type

# What has changed
party end point definitions

# How to test?
Acceptance tests

# Links
https://trello.com/c/I5lc4gDq


AndrewTorrance:
The is PR has become stale and relies on BI cases being removed which may lead to further work. Hence closing this with a view to reappraising AFTER BI cases removed.

PR #261, opened 2018-11-07
Added ability to request list of surveys by type (business or social)

AndrewTorrance:
# Motivation and Context
With the advent of social surveys there is a need to not show social surveys in response_ops_ui

# What has changed
Survey service exposes an endpoint that allows the type of survey to be set. This has ben changed to access that 

# How to test?
Run up gui and look at survey list locally. Modify one survey in DB survey.survey change a survey to 
surveytype ='Social' and reload . Verify does not appear in list. Change back to Business and reload and it now shows

# Links
https://trello.com/c/qY4MVztp# Screenshots (if appropriate):

### DIFF
--- a/response_operations_ui/controllers/survey_controllers.py
+++ b/response_operations_ui/controllers/survey_controllers.py
@@ -87,7 +87,7 @@ def get_survey_ci_classifier(survey_id):
 
 def get_surveys_list():
     logger.debug('Retrieving surveys list')
-    url = f'{app.config["SURVEY_URL"]}/surveys'
+    url = f'{app.config["SURVEY_URL"]}/surveys/surveytype/Business'
     response = requests.get(url, auth=app.config['SURVEY_AUTH'])
 
     if response.status_code == 204:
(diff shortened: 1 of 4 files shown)


response_operations_ui/controllers/survey_controllers.py
```diff
@@ -87,7 +87,7 @@ def get_survey_ci_classifier(survey_id):
 
 def get_surveys_list():
     logger.debug('Retrieving surveys list')
-    url = f'{app.config["SURVEY_URL"]}/surveys'
+    url = f'{app.config["SURVEY_URL"]}/surveys/surveytype/Business'
```
mfcrocker: :art: This capitalisation is odd

necrophonic: I'd agree. Should probably be lowercased like the rest of the url

jcox-dev: It looks like the [survey service expects](https://github.com/ONSdigital/rm-survey-service/blob/c8963f9d3f49eacb565c8349e1b188d440849512/models/surveys.go#L518) a title case string. And the service would require some extra implementation to be case insensitive. 

As this isn't a public facing API we can probably keep this as is for now?

PR #273, opened 2018-12-11
Return user to original page when conversation closed

NickGrantONS:
# Motivation and Context
When a Response-Ops UI user closes a secure message conversation, they should be returned to the inbox page they were on before they closed the conversation.

# What has changed
Passed through the original page the user was on. If the page no longer exists - because the final conversation on the final page has been closed - then it will redirect to the last available page. Also works if the URL is edited to go to page 999, for example.

# How to test?
Run the acceptance tests on the same branch.

To manually test, create enough secure messages to fill more than one page, and close a conversation - you should be returned to the page you started on.

# Links
Trello: https://trello.com/c/lCfPjUPT

### DIFF
--- a/response_operations_ui/views/messages.py
+++ b/response_operations_ui/views/messages.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import math
 from datetime import datetime
 from distutils.util import strtobool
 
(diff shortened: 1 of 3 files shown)


response_operations_ui/views/messages.py
```diff
@@ -181,15 +182,22 @@ def view_selected_survey(selected_survey):
 
         is_closed = request.args.get('is_closed', default='false')
 
+        thread_count = message_controllers.get_conversation_count({'survey': survey_id,
+                                                                   'is_closed': is_closed})
+
+        recalculated_page = _calculate_page(page, limit, thread_count)
```
AndrewTorrance: minor comment :- maybe name var as recalculated_page_num . On first scan its not clear if _calculate_page returns a page object of some kind or number

PR #274, opened 2018-12-18
BUG: Unable to add collection exercises to Sand & Gravel

NickGrantONS:
# Motivation and Context
Sand & Gravel's short name in the database is "Sand&Gravel", which is the reference used to look up the survey from the survey service.

We display "Sand&Gravel" as "Sand & Gravel" (with spaces) but later we try to use this as the short name, failing when we can't find the survey "Survey%20&%20Gravel".

We currently can't open the create collection exercise screen from the link on the survey, and we can't create a new collection exercise even if we type in the correct URL.

This bug stops us from creating the 2019 Sand & Gravel collection exercises without manual DB updates or cURL'ing to the Collex REST endpoint.

# What has changed
Changed the format of the URL used in Response Ops to be more RESTful (slashes, not hyphens). Changed the templates to use the non-pretty-printed survey short name (i.e. strip out the spaces). Corrected a misleading comment.

# How to test?
Create a collection exercise for Sand & Gravel using the UI.

# Links
Trello: https://trello.com/c/LGmwUEHY/571-571-bug-unable-to-add-a-collection-exercise-for-surveys-where-the-shortname-contains-a-space-or-dash

### DIFF
--- a/response_operations_ui/templates/survey.html
+++ b/response_operations_ui/templates/survey.html
@@ -16,7 +16,7 @@
     <div id="created_ce_message" class="panel__body">
       <h1 class="venus">Collection exercise created</h1>
       <p class="mars">
-        <a href="/surveys/{{ survey.shortName }}/{{ newly_created_period }}" id="newly_created_ce_link">View/edit {{survey.shortName}} {{newly_created_period}}</a>
+        <a href="/surveys/{{ survey.shortName.replace(' ', '') }}/{{ newly_created_period }}" id="newly_created_ce_link">View/edit {{survey.shortName}} {{newly_created_period}}</a>
       </p>
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


response_operations_ui/templates/survey.html
```diff
@@ -16,7 +16,7 @@
     <div id="created_ce_message" class="panel__body">
       <h1 class="venus">Collection exercise created</h1>
       <p class="mars">
-        <a href="/surveys/{{ survey.shortName }}/{{ newly_created_period }}" id="newly_created_ce_link">View/edit {{survey.shortName}} {{newly_created_period}}</a>
+        <a href="/surveys/{{ survey.shortName.replace(' ', '') }}/{{ newly_created_period }}" id="newly_created_ce_link">View/edit {{survey.shortName}} {{newly_created_period}}</a>
```
jcox-dev: We should probably be using `url_for` instead of hard-coding the URLs:

```suggestion
        <a href="{{ url_for('collection_exercise_bp.view_collection_exercise', short_name=survey.shortName.replace(' ', ''), period=newly_created_period) }}" id="newly_created_ce_link">View/edit {{survey.shortName}} {{newly_created_period}}</a>
```

response_operations_ui/templates/survey.html
```diff
@@ -57,7 +57,7 @@ <h1 class="jupiter" name="page-survey-title">{{survey.longName}}</h1>
       <div class="ce-list">
         <h2 class="saturn u-pt-m">Collection exercises for {{survey.shortName}}</h2>
         <div class="u-mb-l">
-          <a href="/surveys/{{survey.surveyRef}}-{{survey.shortName}}/create-collection-exercise"
+          <a href="/surveys/{{survey.surveyRef}}/{{survey.shortName.replace(' ', '')}}/create-collection-exercise"
```
jcox-dev: As above:

```suggestion
          <a href="{{ url_for('collection_exercise_bp.get_create_collection_exercise_form', survey_ref=survey.surveyRef, short_name=survey.shortName.replace(' ', '')) }}"
```

PR #276, opened 2018-12-28
113 mark messages as unread

AndrewTorrance:
# Motivation and Context
The requirement was to allow the internal users to be able to mark a message as unread a proxy for further work required . 

# What has changed
The use cases need some explaining. We have a mark as unread link available on the last message in a conversation from a respondent. This only shows if the message was sent to GROUP or to the specific internal user who is looking at the message. IF that link shows then the internal user can select it to mark the message as unread and return to the list of conversations.

There is also a pair of back links on the page. In earlier versions pressing these would return the user to the conversation view and mark the message as read. Now this has slightly different functionality. If the message was sent to group or to the internal user who is logged in then pressing back will mark the message as read. If however the message went to specific internal user and that is NOT the logged in user then pressing back does not mark the message as read.

# How to test?
There are acceptance tests that cover the scenarios but its probably better to manually test these in this repo. 

From frontstage send a message to ONS. This will be sent to group , meaning that any internal user should be able to mark it as unread.  Log into response ops , look at the message and verify that mark as unread label is available . Click it and go the the conversation view . Note message is no longer marked in bold with a dot for unread. Go back into it and select mark unread. Now in the conversation view it is shown as unread. 

When the mark as unread link is selected then on returning to the conversation view you should see a flash message Message from <msg_from> to <msg_to> marked as unread

Log out as this user and log in as another user. ( get the name from public.users table for uaa I think) 
repeat as for the first user and verify this other user can mark as unread etc) . Now using this second user reply to the message . 

On frontstage reply to this message ( simulating a respondent replying to a message from the ONS which now is sent to a specific user ) .

In response ops , still as the second user , verify that you see the mark as unread , press mark as unread to go into the message . Then press BACK and verify that the message now shows read . Go back into the message again and select mark as unread - ready for the next user checks .

Now log in again as the first user . You should see the message in the conversation list . But you should not see the mark as unread link when you open it and pressing back does not mark the message as read.

You should, be able to repeat this switching between users and you should only see mark as unread if a message was sent to GROUP (i.e a new message from respondent) or if a message was sent specifically to the internal user ( i.e respondent replied to a message from that internal user) 

Note: There is a bug here that has been here a while, a ticket has been raised. If an internal user starts a conversation, and they log out and back in again they can no longer see that conversation. 

# Links
https://trello.com/c/Bvakzk6d

### DIFF
--- a/response_operations_ui/templates/conversation-view/cv-message-block.html
+++ b/response_operations_ui/templates/conversation-view/cv-message-block.html
@@ -14,6 +14,13 @@
             Sent:
             <span name="sm-sent-date" id="sm-sent-date-{{ loop.index }}" style="display:inline;">{{ message.get('sent_date') }}</span>
         </span>
+        {% if loop.last %}
+          {% if show_mark_unread == True %}
+            <a class="secure-message-mark-unread" id="sm-mark-as-unread"
+               href={{ url_for('messages_bp.mark_message_unread',
+               message_id=message.message_id, from=message.from, to=message.to) }}>Mark as unread</a>
+          {% endif %}
+        {% endif %}
     </div>
     <div class="secure-message-sent-message-body mars">
         <span style="white-space:pre-line" id="conversation-message-body-{{ loop.index }}" name="conversation-message-body">{{ message.get('body') }}</span>
--- a/response_operations_ui/views/messages.py
+++ b/response_operations_ui/views/messages.py
@@ -74,6 +74,7 @@ def view_conversation(thread_id):
 
     thread_conversation = message_controllers.get_conversation(thread_id)
     refined_thread = [_refine(message) for message in reversed(thread_conversation['messages'])]
+    latest_message = refined_thread[-1]
 
     try:
         closed_time = localise_datetime(datetime.strptime(thread_conversation['closed_at'], "%Y-%m-%dT%H:%M:%S.%f"))
@@ -89,8 +90,8 @@ def view_conversation(thread_id):
             {"title": "Unavailable"}
         ]
 
-    if refined_thread[-1]['unread']:
-        message_controllers.remove_unread_label(refined_thread[-1]['message_id'])
+    if latest_message['unread']:
+        message_controllers.remove_unread_label(latest_message['message_id'])
 
     page = request.args.get('page')
     form = SecureMessageForm(request.form)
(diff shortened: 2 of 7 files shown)


response_operations_ui/views/messages.py
```diff
@@ -433,3 +458,7 @@ def _calculate_page(requested_page, limit, thread_count):
         page = math.ceil(thread_count / limit)
 
     return page
+
+
+def _can_mark_as_unread(message):
+    return message['to_id'] == 'GROUP' or message['to_id'] == session['user_id']
```
jcox-dev: Newline missing.

response_operations_ui/templates/conversation-view/cv-message-block.html
```diff
@@ -14,6 +14,13 @@
             Sent:
             <span name="sm-sent-date" id="sm-sent-date-{{ loop.index }}" style="display:inline;">{{ message.get('sent_date') }}</span>
         </span>
+        {% if loop.last %}
+          {% if show_mark_unread == True %}
+            <a class="secure-message-mark-unread" id="sm-mark-as-unread"
+               href={{ url_for('messages_bp.mark_message_unread',
```
jcox-dev: Do quotes need to be added to either side of the curly brackets, e.g.: `href="{{ url_for...`

response_operations_ui/views/messages.py
```diff
@@ -74,6 +74,7 @@ def view_conversation(thread_id):
 
     thread_conversation = message_controllers.get_conversation(thread_id)
     refined_thread = [_refine(message) for message in reversed(thread_conversation['messages'])]
+    latest_message = refined_thread[-1]
```
jcox-dev: 👍

response_operations_ui/views/messages.py
AndrewTorrance: Done

response_operations_ui/views/messages.py
```diff
@@ -125,19 +126,39 @@ def view_conversation(thread_id):
...
+@messages_bp.route('/mark_unread/<message_id>', methods=['GET'])
+@login_required
+def mark_message_unread(message_id):
+
+    msg_from = request.args.get(get_parameter('from'), type=str, default="")
+    msg_to = request.args.get(get_parameter('to'), type=str, default="")
+
+    message_controllers.add_unread_label(message_id)
+
+    marked_unread_message = f"Message from {msg_from} to {msg_to} marked unread"
```
ryangrundy7: Have you checked that UX are okay with this message as it's different to whats on the design? It's better giving more detail but just making sure

PR #279, opened 2019-01-04
Css compiler fix

owencjones:
Testing improvements to `add-js-task-runner`


Dockerfile
```diff
@@ -1,10 +1,25 @@
...
 EXPOSE 8085
 RUN apt-get update -y && apt-get install -y python-pip && apt-get update -y && apt-get install -y curl
 RUN pip3 install pipenv && pipenv install --deploy --system
 
+
+# Node Install
+# Installs cUrl, uses cUrl to grab and run install script to add deb repos, installs node from this, installs build essentials to allow node-gyp
+# to compile C++ bindings, runs node -v and npm -v to ensure they actually run.
+RUN apt-get install -y curl \
+        && curl -sL https://deb.nodesource.com/setup_10.x | bash \
```
NickGrantONS: why are we using this nodesource.com package? why do we trust it and expect it to always be available?

response_operations_ui/assets/scss/includes/_overrides.scss
```diff
@@ -0,0 +1,335 @@
...
+
+/* Collection exercise - Reporting unit styles */
+
+.ceru-headline {
+    background: #f5f6ff;
+    padding: 1rem;
+    margin: 1rem 0;
+}
+
+.tempHack {
```
NickGrantONS: I guess this just got copied over for legacy reasons?

PR #281, opened 2019-01-11
603 my conversations

AndrewTorrance:
# Motivation and Context
The internal users want the ability to have 'my messages' tab. This is conversations where they were the last internal person to reply or if they sent a first message. 
This change adds that tab. 
Note we now have three 'tabs' my messages, open and close. Messages that appear in my messages also appear in open. So that any internal user can still access any conversation.

# What has changed
Added extra tab. my conversations is now an optional parameter that can be sent to secure message service 

# How to test?
Acceptance tests cover most of it. Bu the easiest way is to send a message from respondent. Should only appear in open messages. Reply as internal and it should also be visible in my messages. Change user and validate that it does not show in my messages for new internal user .But can be seen in open. Send several messages from respondent (more than 1 page worth) view in my messages and open. Validate that back goes back to the correct tab and page

# Links
 https://trello.com/c/djNU7lM5


response_operations_ui/templates/messages.html
```diff
@@ -62,6 +64,8 @@ <h1 class="saturn u-mt-l">Oops! Something went wrong</h1>
         {% if not messages %}
           {% if is_closed %}
           <p>No closed conversations</p>
+          {% elif my_conversations == true %}
```
insacuri: Why are there 2 separate ways of saying 'nothing here boss!'?

response_operations_ui/views/messages.py
```diff
@@ -75,6 +75,7 @@ def view_conversation(thread_id):
     thread_conversation = message_controllers.get_conversation(thread_id)
     refined_thread = [_refine(message) for message in reversed(thread_conversation['messages'])]
     latest_message = refined_thread[-1]
+    my_conversations = request.args.get('my_conversations', default='false')
```
insacuri:
```suggestion
    my_conversations = request.args.get('my_conversations', 'false')
```

I think this is the better way of doing it.  Less words is generally better and leaving out 'default=' is pretty standard.

If you agree, can you do this for all the other ones in here please?  Though after looking at the whole file, there's quite a bit of this so it might be worth leaving it.... your call ^^

AndrewTorrance: I kind of like the explicit default in this specific case. It tells the reader that if no param is found then it defaults to a specific value. It doesn't have a significant performance hit . When I looked I could see several calls made with named params out of order, i.e type given before default. So I modified calls to call default first , so if we do change all we have to do is remove 'default='

PR #282, opened 2019-01-15
Fixed Sand & Gravel 500 error pages

ryangrundy7:
# Motivation and Context
When a internal user clicks on messages for the Sand & Gravel survey, some of the links don't work because the survey gets sent through with spaces. The aim of the pr is to fix this and remove the whitespace so the user can click on Sand & Gravel messages. While looking at Sand & Gravel in response ops, I also found some other links that don't work because of the same bug so they've been fixed as well.
# What has changed
- Added a method to remove whitespace for a survey

# How to test?
- Execute a collection exercise for Sand & Gravel, make sure to click most of the options like attempting to add a reminder, change information for the collex and clicking on cancel/back when trying to add a sample/CI.
- Create a conversation for Sand & Gravel and make sure to click on open and closed messages for that survey, attempt to close a conversation and click back buttons on the conversation view to make sure it goes back correctly.
- Run with ATs
# Links
[Trello](https://trello.com/c/JBFvWCIM)


response_operations_ui/views/messages.py
```diff
@@ -161,6 +161,10 @@ def _view_select_survey(marked_unread_message=""):
                             flash_message=marked_unread_message))
 
 
+def remove_whitespace_from_survey_name(survey):
```
ryangrundy7: Wondering if this should be moved somewhere else like into the common folder?

insacuri: Agreed.  If this was only being used in the one view, it'd be fine.  But a view shouldn't be importing functions from another view.

response_operations_ui/common/remove_whitespace_from_survey.py
```diff
@@ -0,0 +1,2 @@
+def remove_whitespace_from_survey_name(survey):
```
NickGrantONS: is there no utils file you could put this in?

response_operations_ui/views/messages.py
```diff
@@ -109,7 +110,7 @@ def view_conversation(thread_id):
             thread_url = url_for("messages_bp.view_conversation", thread_id=thread_id) + "#latest-message"
             flash(Markup(f'Message sent. <a href={thread_url}>View Message</a>'))
             return redirect(url_for('messages_bp.view_selected_survey',
-                                    selected_survey=refined_thread[0]['survey']))
+                                    selected_survey=refined_thread[0]['survey'].replace(' ', '')))
```
NickGrantONS: why doesn't this use the replace whitespace function?

response_operations_ui/common/remove_whitespace_from_survey.py
AndrewTorrance: This does not remove whitespace , it removes spaces . To remove whitespace   my_var=''.join(my_var.split()).
I am not sure if this needs to be a separate function . If it is then the name should not include 'survey_name' as it will work with any string

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -30,6 +31,7 @@ def get_survey_by_id(survey_id):
 
 
 def get_survey_by_shortname(short_name):
+    short_name = remove_whitespace_from_survey_name(short_name)
```
AndrewTorrance: maybe short_name = ''.join(short_name.split()) ?

response_operations_ui/views/messages.py
```diff
@@ -240,7 +240,8 @@ def view_selected_survey(selected_survey):
                                page=page,
                                breadcrumbs=breadcrumbs,
                                messages=messages,
-                               selected_survey=formatted_survey,
+                               selected_survey=selected_survey,
+                               formatted_survey=formatted_survey,
```
AndrewTorrance: Could we have a more descriptive name ? I am looking at this and wondering why formatted ? Why selected ? Perhaps name_for_display and selected_survey_short_name ( or something better ?)

response_operations_ui/views/messages.py
ryangrundy7: Done

response_operations_ui/controllers/survey_controllers.py
ryangrundy7: I've changed it to here because it's the only place that does it 👍

response_operations_ui/views/messages.py
ryangrundy7: I've changed the formatted name to `displayed_short_name`

PR #283, opened 2019-01-15
Frontend quick wins

owencjones:
# Motivation and Context
Whilst we are in a slight lull, this is some minor frontend refactors that improve a few things here and there.

# What has changed
* Minor change to watch task
* CSS refactoring as appropriate
* Deletion of old CSS files

# How to test?
* run solution and dependencies
* run acceptance tests
* look at resops to ensure it looks as it should


response_operations_ui/assets/scss/includes/components/data-panel.scss
```diff
@@ -23,14 +23,14 @@
...
     }
-}
 
-.data-panel-header__title {
-    border: 0;
-    background: none;
-    color: #4263c2;
-    padding: 0;
-    display: none;
+    &__title {
```
NickGrantONS: SASS FTW

PR #289, opened 2019-01-31
added new functionality to respondents tab

ChristopherMShelton:
edit  contact details, resend verification email, enable/disable enrolment)

# Motivation and Context
Business requested ability to make these changes from a respondents page

# What has changed
Buttons added for edit contact details, resend verification email and enable/disable enrolment added and appropriate changes made to enable that.

# How to test?
run tests on view_edit_respondents_tab branch
alternatively can manual test by clicking through from the respondents tab.

# Links
https://trello.com/c/JOPM1IFq/394-us216-view-respondent-details-m
https://github.com/ONSdigital/rasrm-acceptance-tests/pull/258


response_operations_ui/views/respondents.py
```diff
@@ -50,4 +50,91 @@ def respondent_details(respondent_id):
...
+@respondent_bp.route('/<ru_ref>/change-enrolment-status', methods=['GET'])
+@login_required
+def confirm_change_enrolment_status(ru_ref):
+    return render_template('confirm-enrolment-change.html', business_id=request.args['business_id'], ru_ref=ru_ref,
+                           ru_name=request.args.get('ru_name'),
+                           trading_as=request.args['trading_as'], survey_id=request.args['survey_id'],
+                           survey_name=request.args['survey_name'], respondent_id=request.args['respondent_id'],
+                           first_name=request.args['respondent_first_name'],
+                           last_name=request.args['respondent_last_name'],
+                           change_flag=request.args['change_flag'],
```
owencjones: These functions you've added need to be tested - so I'm assuming that you're covering their testing with acceptance tests repo, because they aren't very pure, and thus not hugely testable directly.

If there isn't testing for this functionality at all, it needs adding.

response_operations_ui/views/respondents.py
```diff
@@ -50,4 +50,91 @@ def respondent_details(respondent_id):
...
+    reporting_units_controllers.resend_verification_email(party_id)
+    logger.info("Re-sent verification email.", party_id=party_id)
+    return redirect(url_for('respondent_bp.respondent_details', respondent_id=party_id,
+                            info='Verification email re-sent'))
+
+
+@respondent_bp.route('<respondent_id>/change-enrolment-status', methods=['POST'])
+@login_required
+def change_enrolment_status(respondent_id):
+
```
insacuri: 🎨 Remove extra line here

response_operations_ui/views/reporting_units.py
```diff
@@ -189,7 +189,7 @@ def view_resend_verification(ru_ref, party_id):
     email = respondent['pendingEmailAddress'] if 'pendingEmailAddress' in respondent \
         else respondent['emailAddress']
 
-    return render_template('re-send-verification-email.html', ru_ref=ru_ref, email=email)
+    return render_template('re-send-verification-email.html', ru_ref=ru_ref, email=email, tab='reporting_units')
```
insacuri: My gut says this should be `active_tab` or something similar.  Just having a `tab` variable doesn't properly describe what it's doing at a glance

insacuri: And anothe gut feeling is that having tab here is wrong.  I think somewhere in either frontstage or response-operations-ui, in the template theres an 'if <thing> in url then add active class to tab' or something like that

response_operations_ui/views/respondents.py
```diff
@@ -50,4 +50,91 @@ def respondent_details(respondent_id):
     ]
 
     respondent['status'] = respondent['status'].title()
-    return render_template('respondent.html', respondent=respondent, enrolments=enrolments, breadcrumbs=breadcrumbs)
+
+    info_message = None
```
insacuri: This whole block feels messy.  Have you looked into using Flash (http://flask.pocoo.org/docs/1.0/patterns/flashing/), it might be a cleaner way to do the same thing

response_operations_ui/views/respondents.py
```diff
@@ -50,4 +50,91 @@ def respondent_details(respondent_id):
...
+                           form=form, tab='respondents')
+
+
+@respondent_bp.route('/edit-contact-details/<respondent_id>', methods=['POST'])
+@login_required
+def edit_contact_details(respondent_id):
+    form = request.form
+    contact_details_changed = party_controller.update_contact_details(respondent_id, form)
+
+    ui_message = 'No updates were necessary'
```
insacuri: Same as above. Flash might be better here.
Additionally, I think this is the wrong way to do it.  This kind of code is just asking for weird bugs to occur when multiple things are provided and the message keeps changing.

response_operations_ui/views/respondents.py
insacuri: Also, as with the below comment, having lots of if statements that keep changing the message is asking for weird bugs to occur at some point.  There's probably a better way to do it (off the top of my head, using if/elif might be the simplest one, but there might be cleaner ways to do it)

response_operations_ui/views/respondents.py
```diff
@@ -50,4 +50,91 @@ def respondent_details(respondent_id):
...
+                                                        respondent_id=respondent_id,
+                                                        survey_id=request.args['survey_id'],
+                                                        change_flag=request.args['change_flag'])
+    return redirect(url_for('respondent_bp.respondent_details', respondent_id=respondent_id,  enrolment_changed='True'))
+
+
+@respondent_bp.route('/<ru_ref>/change-enrolment-status', methods=['GET'])
+@login_required
+def confirm_change_enrolment_status(ru_ref):
+    return render_template('confirm-enrolment-change.html', business_id=request.args['business_id'], ru_ref=ru_ref,
```
insacuri: This feels wrong to me, but I'm not sure what 'right' looks like.  I assume the url for this looks like `<host>/<ru_ref>/change-enrolment-status?ru_name=blah&trading_as=blah.......`  If that's right, then it seems like such a crazy long url and unnecessary.  Though maybe I'm wrong in thinking that,  I'd ask others to see what they think.

response_operations_ui/views/respondents.py
insacuri: Thinking about this more, is probably the way to do this.  You can flash multiple times, then in the template you can loop and render every flashed message.  
This way, if there are multiple things to report on, you'll see them all as opposed to just the last one

NickGrantONS:
you should definitely be using the Flash pattern if you're not using it, to display feedback messages to users as @insacuri indicates. There should be examples in the codebase you can copy.

response_operations_ui/views/reporting_units.py
owencjones: I'm not sure about basing it on the url, as it means two places to refactor if we change the URL for some reason, and can also be an issue with matching spuriously.

That said, I wouldn't necessarily pass it here either.

response_operations_ui/templates/respondent.html
```diff
@@ -40,8 +40,7 @@ <h1 class="jupiter" id="respondent-name">{{ respondent.firstName }} {{ responden
                         <dd id="respondent-account-status">
                             {% if respondent.status  == 'Suspended' %} Locked {% else %} {{ respondent.status }} {% endif %}
                           <br>
-                          <a id="resend-verification-email-btn" href="{{ url_for('respondent_bp.view_resend_verification', respondent_id=respondent.id) }}">Re-send verification email</a>
-                          </br>
+                          {% if respondent.status == 'Suspended' %}<a id="resend-verification-email-btn" href="{{ url_for('respondent_bp.view_resend_verification', respondent_id=respondent.id) }}">Re-send verification email</a> {% endif %}
```
gavinleeedwards: - Should show the 'resend email verification' link if the enrolment status is pending i.e.  if a user has just created an account.  
- When the account is locked/suspended, there should be an option to unlock the account rather than resend a verification email.  
- If there's a new (pending) email address then there should be the option to resend the verification email.

response_operations_ui/views/respondents.py
insacuri: We can ignore this for now.  I don't know enough to offer an alternative, and there are other things that need doing instead

response_operations_ui/views/respondents.py
```diff
@@ -50,4 +49,80 @@ def respondent_details(respondent_id):
...
+                                                        respondent_id=respondent_id,
+                                                        survey_id=request.args['survey_id'],
+                                                        change_flag=request.args['change_flag'])
+    return redirect(url_for('respondent_bp.respondent_details', respondent_id=respondent_id, enrolment_changed='True'))
+
+
+@respondent_bp.route('/<ru_ref>/change-respondent-status', methods=['POST'])
+@login_required
+def change_respondent_status(respondent_id):
+    reporting_units_controllers.change_respondent_status(respondent_id=respondent_id,
```
insacuri: As we discussed, if the change_flag isn't exactly 'ACTIVE' then nothing will be updated, no log line will inform anyone that nothing was done, and this function will say that everything worked because the 'account_status_changed' is hardcoded to true.

As a short term fix, changing the change_respondent_status function to throw an exception if something unexpected is given to it.  It sucks because it's another step in the 'throw a 500 if everything isn't perfect' direction, but at least we'll catch potential issues with this change.

response_operations_ui/templates/respondent.html
ChristopherMShelton: added this!

response_operations_ui/controllers/reporting_units_controllers.py
```diff
@@ -67,6 +67,10 @@ def change_respondent_status(respondent_id, change_flag):
             raise ApiError(response)
 
         logger.debug('Successfully changed respondent status', respondent_id=respondent_id, change_flag=change_flag)
+    else:
+        error = 'Incorrect change_flag given'
+        logger.error(error)
```
ryangrundy7: You shouldn't have to make a string of a message and pass it through. Can do what happens on line 66 such as 
```python
logger.error('Incorrect change_flag given', respondent_id=respondent_id, change_flag=change_flag)
```

insacuri: Agreed, but the next line is raising a ValueError with the same text.  Might as well not repeat ourselves

ChristopherMShelton: changed

response_operations_ui/views/respondents.py
ChristopherMShelton: added

response_operations_ui/views/respondents.py
ChristopherMShelton: added now

response_operations_ui/views/respondents.py
ChristopherMShelton: added now

ChristopherMShelton:
> There's no option to resend a verification email if the account has a pending enrolment (i.e. an account has been created but not yet verified).

added 'created' to the check of when to show the button

owencjones:
Changes that I asked for are now done.

@insacuri, @ryangrundy7 , and @gavinleeedwards happy with your requested changes?

gavinleeedwards:
> > There's no option to resend a verification email if the account has a pending enrolment (i.e. an account has been created but not yet verified).
> 
> added 'created' to the check of when to show the button

This isn't quite right.  Adding the 'resend' link for all created accounts will mean the 'resend verification' link is still shown for accounts that have been fully verified.  It needs to be there for accounts that are in 'Created' state but have a pending enrolment against them.

response_operations_ui/templates/reporting-unit.html
```diff
@@ -4,7 +4,7 @@
 {% block main %}
 
 
-{% include 'partials/info-box.html' %}
+{% include 'partials/flashed-messages.html' %}
```
ryangrundy7: Breaking acceptance tests by changing this. Need to add in a success id

ChristopherMShelton: I've pushed acceptance test changes to match

ryangrundy7: I had to delete my branch on AT to get it working, was really weird

gavinleeedwards:
All works fine for me now 👍

response_operations_ui/views/reporting_units.py
```diff
@@ -1,7 +1,7 @@
 import logging
 from datetime import datetime, timezone
 
-from flask import Blueprint, render_template, request, redirect, url_for
+from flask import Blueprint, render_template, request, redirect, url_for, flash
```
ryangrundy7: Small styling issue, imports really should be in alphabetical order

response_operations_ui/views/respondents.py
```diff
@@ -1,11 +1,11 @@
 import logging
 
-from flask import Blueprint, redirect, render_template, request, url_for
+from flask import Blueprint, redirect, render_template, request, url_for, flash
```
ryangrundy7: same here

ChristopherMShelton: done

response_operations_ui/views/reporting_units.py
ChristopherMShelton: done

PR #293, opened 2019-03-04
Search respondents ui

owencjones:
# Motivation and Context
UI changes to allow searching of respondents by first name, last name, and email, rather than just by email.

# What has changed
* Views changed to include first name and last name fields,
* Search form separation for respondents and reporting units
* Added mock party controller to allow testing locally without party connection
* Added mock data for mock controller
* Added new results view page, with pagination
* Styling
* Created utils helper for respondent section of site
* Unit tests
* Added new UI tests to avoid adding non-functional tests to acceptance tests repo
* Added some tests for previously untested areas of code
* Updated to pattern library 2.0.0

# How to test?
* Run locally with `PARTY_FAKE_DATA` variable set to 1
* Confirm you can search respondents (you can't click through to a dummy respondent)
* Acceptance tests not yet addressed

# Links
https://trello.com/c/Mi1Zpw1l/707-us215-front-end-search-for-a-respondents-account-name-part-of-email-address-8

**When approved, don't merge - I need to merge acc tests at the same time**


config.py
```diff
@@ -109,6 +110,8 @@ class DevelopmentConfig(Config):
     PARTY_USERNAME = os.getenv('PARTY_USERNAME', 'admin')
     PARTY_PASSWORD = os.getenv('PARTY_PASSWORD', 'secret')
     PARTY_AUTH = (PARTY_USERNAME, PARTY_PASSWORD)
+    PARTY_FAKE_DATA = bool(os.getenv('FAKE_PARTY_DATA', False))
```
NickGrantONS: I'm not a fan of testing parameters making their way into production. Is there some way we can confine this only to the testing code?

response_operations_ui/controllers/mock_party_controller.py
```diff
@@ -0,0 +1,32 @@
+import json
+
+from flask import current_app as app
+
```
owencjones: So, for clarity, this is the file that replaces `party_controller.py` to allow UI-only tests to run without spinning up party or other dependencies.

The long term plan is that `party_controller` will become a class, and mock data will be handled in that, in a similar way to how party does it.  I'm adding a ticket to the backlog for that.

response_operations_ui/controllers/party_controller.py
```diff
@@ -125,6 +130,48 @@ def search_respondent_by_email(email):
     return response.json()
 
 
+def search_respondents(first_name, last_name, email_address, page=0):
+    if bool(os.environ.get('FAKE_PARTY_DATA', False)):
```
owencjones: Currently, this uses an environment variable to allow switching during session, but this would be replaced in later with test modes that don't affect the wider system.

tests/controllers/test_party_controller.py
```diff
@@ -0,0 +1,94 @@
+import unittest
+import mock
+from collections import namedtuple
+
+from response_operations_ui import create_app
+from response_operations_ui.controllers.party_controller import search_respondents, _get_search_respondents_url
+from response_operations_ui.exceptions.exceptions import SearchRespondentsException
+
```
owencjones: It worries me that `party_controller` didn't already have a test file.  This one only tests my code, but I'm adding a card to address this, as I think it's quite important.

tests/test_exceptions.py
```diff
@@ -0,0 +1,43 @@
+from collections import namedtuple
+
+import unittest
+from response_operations_ui.exceptions.exceptions import ApiError, InternalError, UpdateContactDetailsException,\
+    SearchRespondentsException
+
+fake_response = namedtuple('Response', 'url status_code text exception')
+
```
owencjones: Only one of the exceptions in the `exceptions.py` file is mine, however I added tests for all of them (except one which doesn't return anything at all) because it was an opportunity to do this.

tests/views/test_respondents.py
```diff
@@ -201,3 +108,74 @@ def test_change_respondent_status(self, mock_request):
                                     f"?tab=respondents&change_flag=ACTIVE",
                                     follow_redirects=True)
         self.assertEqual(response.status_code, 200)
+
```
owencjones: This is the UI style testing that I've added, it negates the need to mock large amounts of other functions, and allows you to run UI style tests, without including them in the acceptance test repo (acceptance tests and UI tests are different).

Included here as both testing for this, and also a useful talking point as to how we want to proceed - if we decide on a different direction, I'll swap it for something else.

config.py
owencjones: It's only in the `DevelopmentConfig` and `TestingConfig` classes - as are many other similar things, like the lines above.  I kind of agree with you, but the paradigm is already set that way I'd say.

I think there would be value in adding some safeguards to this though, to prevent setting of the flags that enable this in production and pre-prod?  Maybe a check for env vars that would only be set on a live instance?

tests/controllers/test_mock_party_controller.py
```diff
@@ -0,0 +1,161 @@
...
+    {'firstName': 'Hadleigh', 'lastName': 'Duffyn', 'emailAddress': 'hduffynl@buzzfeed.com'},
+    {'firstName': 'Kelby', 'lastName': 'Dimblebee', 'emailAddress': 'kdimblebeem@ask.com'},
+    {'firstName': 'Torrey', 'lastName': 'Palin', 'emailAddress': 'tpalinn@technorati.com'},
+    {'firstName': 'Malinde', 'lastName': 'Featley', 'emailAddress': 'mfeatleyo@oakley.com'},
+    {'firstName': 'Chico', 'lastName': 'Thrasher', 'emailAddress': 'cthrasherp@cbc.ca'},
+    {'firstName': 'Thorvald', 'lastName': 'Puleston', 'emailAddress': 'tpulestonq@latimes.com'},
+    {'firstName': 'Hendrick', 'lastName': 'Pollins', 'emailAddress': 'hpollinsr@sfgate.com'},
+    {'firstName': 'Lyn', 'lastName': 'MacGaughie', 'emailAddress': 'lmacgaughies@webnode.com'},
+    {'firstName': 'Patrick', 'lastName': 'Mothersdale', 'emailAddress': 'pmothersdalet@istockphoto.com'}
+]
```
ChristopherMShelton: is it worth sticking this in another file to reuse?

owencjones: At this stage, I don't think so, as it's only the data that this endpoint would return.  Wouldn't be difficult to if you're really keen on that happening though.

response_operations_ui/controllers/mock_party_controller.py
```diff
@@ -0,0 +1,32 @@
...
+def mock_search_respondents(first_name, last_name, email_address, page=1):
+    with open('response_operations_ui/data/mock_search_respondents_data.json', 'r') as json_data:
+        data = json.load(json_data)
+        limit = app.config['PARTY_RESPONDENTS_PER_PAGE']
+
+        chunk_range = range((len(data) + limit - 1) // limit)
+        data_by_surname = sorted(data, key=lambda dict: dict["lastName"].lower())
+        sorted_data = filter_json_by_passed_parameters(data_by_surname, first_name, last_name, email_address)
+        data_chunks = [sorted_data[i * limit:(i + 1) * limit] for i in chunk_range]
```
ChristopherMShelton: maybe add a short doc string just to say what this does, obvious enough if you know imo

tests/controllers/test_mock_party_controller.py
ChristopherMShelton: Nah that's fair enough, wasn't sure if we'd have any use for a block of account details for anything else

response_operations_ui/controllers/party_controller.py
NickGrantONS: again... not loving the test code in the production code

response_operations_ui/templates/respondent-search/search-respondents-results.html
```diff
@@ -0,0 +1,19 @@
...
+    <div class="grid">
+        <div class="grid__col col-12@l">
+            {% include 'respondent-search/_inline-form.html' %}
+    
+            <h2 class="u-fs-l">{{ respondent_count }} respondents found.</h2>
+            <h3 class="u-mb-m">Displaying <strong>{{ first_index }} - {{ last_index }}</strong> of <strong>{{ respondent_count }}</strong> </h3>
+    
+            {% if respondents %}{% include 'respondent-search/_table.html' %}{% endif %}
+
+            {% if respondent_count > 25 %}{% include 'respondent-search/_pagination.html' %}{% endif %}
```
NickGrantONS: probably shouldn't be hard-coded to `25`. This is a configurable parameter, isn't it?

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +18,93 @@
...
+    form.email_address.data = email_address
+
+    party_response = party_controller.search_respondents(first_name, last_name, email_address, page)
+
+    respondents = party_response.get('data', [])
+    total_respondents_available = party_response.get('total', 0)
+
+    filtered_respondents = filter_respondents(respondents)
+
+    RESULTS_PER_PAGE = app.config["PARTY_RESPONDENTS_PER_PAGE"]
```
NickGrantONS: this is the constant that you should use in the place where you've got 25 hard-coded

tests/views/test_respondents.py
```diff
@@ -23,6 +24,9 @@
 iac_1 = 'jkbvyklkwj88'
 iac_2 = 'ljbgg3kgstr4'
 
+get_respondent_root = f'http://localhost:8085/party-api/v1/respondents/'
```
NickGrantONS: did you mean to leave `localhost:8085` hard-coded in here?

response_operations_ui/controllers/party_controller.py
insacuri: ^ This.  We've got this in secure_message and I hate it there also.  I really reallllly don't want that to be an accepted paradigm to put test code in classes that'll be used in production

config.py
```diff
@@ -109,6 +110,8 @@ class DevelopmentConfig(Config):
...
+    PARTY_RESPONDENTS_PER_PAGE = os.getenv('PARTY_RESPONDENTS_PER_PAGE', 25)
```
NickGrantONS: This looks like a dupe to me. See line 56

response_operations_ui/controllers/mock_party_controller.py
owencjones: Done. Mark resolved if you're happy

config.py
NickGrantONS: we should refactor so that real implementations can be swapped out with mock implementations, for testing, and separate the production config from the testing config

config.py
owencjones: It's a dupe to make future changes easier.  Most of the `DevelopmentConfig` members are dupes of `TestingConfig` that just make differences easier to note and make.

response_operations_ui/controllers/party_controller.py
```diff
@@ -125,6 +130,48 @@ def search_respondent_by_email(email):
...
+    kw_to_url_params = {
+        'first_name': 'firstName',
+        'last_name': 'lastName',
+        'email_address': 'email',
+        'page': 'page'
+    }
+
+    for resops_name, party_name in kw_to_url_params.items():
+        arg = kwargs.get(resops_name)
+        if arg and arg != '':
```
ryangrundy7: I don't think you need the `and arg != ''` part because `if arg` will be truthy if any value is passed in and if empty it should skip.

response_operations_ui/controllers/party_controller.py
```diff
@@ -125,6 +130,48 @@ def search_respondent_by_email(email):
...
+
+
+def _get_search_respondents_url(**kwargs):
+    url = f'{app.config["PARTY_URL"]}/party-api/v1/respondents?'
+
+    # Dict provides translation of variable names in response ops to url param names in party
+    kw_to_url_params = {
+        'first_name': 'firstName',
+        'last_name': 'lastName',
+        'email_address': 'email',
```
ryangrundy7: This needs to change from `email` to `emailAddress` otherwise you'll get a 400 from the party service

response_operations_ui/controllers/party_controller.py
```diff
@@ -125,6 +130,48 @@ def search_respondent_by_email(email):
...
+                                             first_name=first_name,
+                                             last_name=last_name,
+                                             email_address=email_address,
+                                             page=page)
+        data = response.json()
+
+    return data
+
+
+def _get_search_respondents_url(**kwargs):
```
insacuri: 2 things.  This should probably be called `_get_search_respondents_url` second.  requests.get has a params argument that should probably be used here instead of a function to insert the params into the url by hand

response_operations_ui/controllers/party_controller.py
```diff
@@ -125,6 +130,48 @@ def search_respondent_by_email(email):
...
+        logger.debug('Respondent Search: Using fake data for controller')
+        data = mock_search_respondents(first_name, last_name, email_address, page)
+    else:
+        url = _get_search_respondents_url(first_name=first_name,
+                                          last_name=last_name,
+                                          email_address=email_address,
+                                          page=page)
+        response = requests.get(url, auth=app.config['PARTY_AUTH'])
+        if response.status_code != 200:
+            raise SearchRespondentsException(response,
```
insacuri: Is a new exception type necessary?  These details will be in the url (via parameters), so an APIError is probably more correct to throw here.

response_operations_ui/forms.py
```diff
@@ -45,6 +45,30 @@ class SecureMessageForm(FlaskForm):
...
+    page = HiddenField('page')
+    submit = SubmitField('Search')
+
+    def validate(form):
+        first_name = form.first_name.data or ''
+        last_name = form.last_name.data or ''
+        email_address = form.email_address.data or ''
+        source = form.source.data or 'home'
+
+        if not (len(first_name) > 0 or len(last_name) > 0 or len(email_address) > 0):
```
insacuri: Can this be rewritten `if not first_name or last_name or email_address:` I think empty values are false

response_operations_ui/forms.py
```diff
@@ -45,6 +45,30 @@ class SecureMessageForm(FlaskForm):
...
+            return False
+
+        if source != 'home' and source != 'results':
+            return False
+
+        if first_name != '' or last_name != '' or email_address != '':
```
insacuri: Is this needed?  The top one covers this doesn't it, and an absence of a false value is valid right?  or is there something I've forgotten?

response_operations_ui/templates/respondent-search/_table.html
```diff
@@ -0,0 +1,24 @@
...
+        {% for respondent in respondents %}
+        <tr>
+            <td>
+                <a href="{{ respondent.href }}">{{ respondent.name }}</a>
+            </td>
+            <td>
+                {{ respondent.email }}
+            </td>
+            <td>
+                <span class="status status--small {{ respondent.status_class }}">{{ respondent.status | title }}</span>
```
insacuri: This is one for someone else,  is it correct to have display/css knowledge in the code?  My gut says no, but I don't know what would be better.  Maybe a partial?  That seems like a lot to just cleanly sort out the classes on a tag.

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +18,93 @@
...
+    query_string = urlencode({
+        'email_address': form.email_address.data or '',
+        'first_name': form.first_name.data or '',
+        'last_name': form.last_name.data or '',
+        'source': source,
+        'page': page
+    })
+
+    if not form_valid:
+        redirect_url = url_for('respondent_bp.respondent_home')
```
insacuri: No need for the final return.  Just do it here `return redirect(url_for('respondent_bp.respondent_home'))`

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +18,93 @@
...
+    else:
+        redirect_url = urljoin(url_for('respondent_bp.respondent_search'), '?' + query_string)
```
insacuri: url_for allows for parameters to be added.  https://stackoverflow.com/questions/7478366/create-dynamic-urls-in-flask-with-url-for 

It's better to do it all here instead of constructing parameter strings yourself (this is true in almost every instance I think, any tool that deals with urls will usually have its own inbuilt way to pass parameters).
Additionally, if the form isn't valid then it does a bunch of stuff it doesn't need to do.  The code should be something like 
```python
if not form_valid:
    #do all the invalid stuff
else:
    return redirect(url_for('respondent_bp.respondent_home'))
```

config.py
owencjones: Done. On it's way in

response_operations_ui/templates/respondent-search/search-respondents-results.html
owencjones: Good catch.

response_operations_ui/views/respondents.py
owencjones: I feel like I could've worked this out, but ta!

tests/views/test_respondents.py
owencjones: Yeah, it's not a problem in this context.

response_operations_ui/controllers/party_controller.py
owencjones: Done in next push

response_operations_ui/controllers/party_controller.py
owencjones: Done in next push

response_operations_ui/controllers/party_controller.py
owencjones: The new exception type allows me to include data I otherwise couldn't - namely the source and email address tried to search.  This seems like it _might_ make debugging easier.

Also follows the convention of other new exceptions, like `UpdateContactDetailsException`

response_operations_ui/controllers/party_controller.py
owencjones: > This should probably be called `_get_search_respondents_url`
I'm confused.   That _is_ what it's called.

Regarding the keywords, do you feel particularly strongly about it?  I'd like to move this on quickly if poss, and this works.

response_operations_ui/forms.py
owencjones: Done in the next commit

response_operations_ui/forms.py
owencjones: I think you're right yeah

response_operations_ui/templates/respondent-search/_table.html
owencjones: Myself, I would say this is right.  The class determination logic isn't in the code, so I have separated concerns.

response_operations_ui/views/respondents.py
owencjones: done in next commit

response_operations_ui/views/respondents.py
owencjones: Refactored in next commit

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +18,93 @@
...
+    respondents = party_response.get('data', [])
+    total_respondents_available = party_response.get('total', 0)
+
+    filtered_respondents = filter_respondents(respondents)
+
+    RESULTS_PER_PAGE = app.config["PARTY_RESPONDENTS_PER_PAGE"]
+
+    offset = (int(page) - 1) * RESULTS_PER_PAGE
+
+    first_index = 1 + offset
```
insacuri: You're only using both of these variables once, might as well inline them in the render_template below

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +18,93 @@
...
+@respondent_bp.route('/search')
+@respondent_bp.route('/search/', methods=['GET'])
 @login_required
 def respondent_search():
-    form = SearchForm()
-    breadcrumbs = [{"title": "Respondents"}]
-    response = None
+    breadcrumbs = [{"title": "Respondents"}, {"title": "Search"}]
+
+    args = get_controller_args_from_request(request)
```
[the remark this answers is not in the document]
owencjones: I've pulled it out in the next push

response_operations_ui/controllers/party_controller.py
```diff
@@ -125,6 +127,44 @@ def search_respondent_by_email(email):
...
+        'first_name': 'firstName',
+        'last_name': 'lastName',
+        'email_address': 'emailAddress',
+        'page': 'page'
+    }
+
+    for resops_name, party_name in kw_to_url_params.items():
+        arg = kwargs.get(resops_name)
+        if arg:
+            url += f'{party_name}={arg}&'
```
NickGrantONS: This looks vulnerable to an injection attack to me

response_operations_ui/views/respondents.py
owencjones: Removed in next push

response_operations_ui/controllers/party_controller.py
owencjones: That depends on whether the service the url is calling is sanitising it's inputs - I'm told it is.

Either way, I can sanitise this, but the service will remain open to attach itself.

config.py
```diff
@@ -109,6 +110,7 @@ class DevelopmentConfig(Config):
     PARTY_USERNAME = os.getenv('PARTY_USERNAME', 'admin')
     PARTY_PASSWORD = os.getenv('PARTY_PASSWORD', 'secret')
     PARTY_AUTH = (PARTY_USERNAME, PARTY_PASSWORD)
+    PARTY_RESPONDENTS_PER_PAGE = os.getenv('PARTY_RESPONDENTS_PER_PAGE', 25)
```
insacuri: If it's defined above then you don't need to define it a second time right?

response_operations_ui/common/respondent_utils.py
```diff
@@ -0,0 +1,34 @@
...
+        try:
+            filtered_respondents.append({
+                'href': '/respondent-details/' + str(respondent['id']),
+                'name': respondent['firstName'] + ' ' + respondent['lastName'],
+                'email': respondent['emailAddress'],
+                'status': status_enum_to_string(respondent['status']),
+                'status_class': status_enum_to_class(respondent['status'])
+            })
+        except KeyError:
+            return []
```
insacuri: Looking at this again, this should probably change.  If any single one of the respondents has something wrong with it, then you get nothing back.   Is there a way we can gracefully degrade this?  Maybe with empty fields or  'Not found' values?  At the minimum a log line should happen if we're going to return nothing.

response_operations_ui/controllers/party_controller.py
insacuri: I copied it then forgot to change it.... I was going to suggest `_build_search_respondents_url` because it is building it

insacuri: And for the 2nd thing, I get you want to move quickly, but with this we have another function that we have to maintain, and fix problems with if we find any; meanwhile, we have a tool that's been tested a ton and definitely works.
Less code means there's less stuff that can break

response_operations_ui/forms.py
```diff
@@ -45,6 +45,29 @@ class SecureMessageForm(FlaskForm):
...
+    def validate(form):
+        first_name = form.first_name.data or ''
+        last_name = form.last_name.data or ''
+        email_address = form.email_address.data or ''
+        source = form.source.data or 'home'
+
+        if not (first_name or last_name or email_address):
+            return False
+
+        if source != 'home' and source != 'results':
```
insacuri: Can be rewritten to `if source not in ['home', 'results']:` tidier and easier to add to in future if we need to

config.py
owencjones: You don't, but everything else seemed to be.  Removed it now.

response_operations_ui/templates/respondent-search/_inline-form.html
```diff
@@ -0,0 +1,23 @@
+<form action="{{ url_for('respondent_bp.search_redirect') }}" method="post" class="u-mt-l">
```
insacuri: Why do some of the files in here start with underscore?  I can't find another example of that anywhere else

response_operations_ui/templates/respondent-search/_table.html
insacuri: That was my point, I was unsure whether the code should know about styles?  My gut says style information should live within the template, but the way you've done it looks so reasonable, which is why I asked the question.

response_operations_ui/common/respondent_utils.py
owencjones: That was the intention, but I don't really know the chances of getting data back with poor integrity.  I think on reflection I'll switch it to just not add the entry and log that it was skipped, so we still get the rest of the list.

response_operations_ui/forms.py
owencjones: On the way in.

response_operations_ui/templates/respondent-search/_inline-form.html
owencjones: Templates that are parts of a page, rather than a page itself begin with an underscore in this.  Didn't realise that that convention hasn't been followed elsewhere.

Can convert them back if need be?

response_operations_ui/common/respondent_utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def status_enum_to_class(status):
+    dict = {
+        'ACTIVE': 'status--success',
+        'CREATED': 'status--info',
+        'SUSPENDED': 'status--warning',
+        'LOCKED': 'status--dead'
+    }
+    return dict.get(status, None)
```
insacuri: 2 minor things, the variable name dict should probably be something else; whilst it's a keyword, there are other uses for that word that have semantic meaning so I try to steer away from it where possible.
Second thing, no need to define None as the default as that's what it defaults too already.

insacuri: This comment applies to the other one too

response_operations_ui/common/respondent_utils.py
insacuri: That's fair.  In the past there was a ton of 'we didn't get what we wanted perfectly?  500 error'.  Which makes developing pretty miserable in places.  I think graceful degradation is preferable where ever possible to not blast users out of their journey

response_operations_ui/common/respondent_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        try:
+            filtered_respondents.append({
+                'href': '/respondent-details/' + str(respondent['id']),
+                'name': respondent['firstName'] + ' ' + respondent['lastName'],
+                'email': respondent['emailAddress'],
+                'status': status_enum_to_string(respondent['status']),
+                'status_class': status_enum_to_class(respondent['status'])
+            })
+        except KeyError:
+            logger.debug(f'Could not add respondent to retrieved list, as data structure was not that expected.',
```
insacuri: You're going to hate me, but we shouldn't be logging any personal information.  For this, at most, we should be logging the respondent id.  There shouldn't be any names/email addresses in the logs, which could potentially happen for this.

owencjones: NP, fair point

response_operations_ui/controllers/party_controller.py
owencjones: So, you want to remove the function entirely?

response_operations_ui/templates/respondent-search/_inline-form.html
insacuri: I think changing them is best. If we find they're confusing or whatever, then we can fix that how we want down the line.  It looks like the collection_exercise section of the templates has done the same in putting all the bits in the same place as the main template.  I think it's better to be consistent than right for this.

response_operations_ui/common/respondent_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def status_enum_to_string(status):
+    enum_lookup = {
+        'ACTIVE': 'Active',
+        'CREATED': 'Created',
+        'SUSPENDED': 'Suspended',
+        'LOCKED': 'Locked'
+    }
+    return enum_lookup.get(status, None)
```
insacuri: These defaults need to be removed still

response_operations_ui/controllers/party_controller.py
insacuri: Yes. Just create a list of parameters (I think request.get takes a list of params as an arguement... or maybe it's a dict...) and pass that into the request.get call.  It'll bloat that function a tiny bit but it won't be much.

owencjones: Done. YAGNI

response_operations_ui/templates/respondent-search/_table.html
owencjones: I see what you mean there - my take on it is that I prefer the logic in the view rather than the template itself.

I suppose there could be a middle ground of the class translator in a Jinja2 macro, but unless you feel strongly about it, I'd rather not.

response_operations_ui/controllers/party_controller.py
owencjones: Function removed now anyway.

response_operations_ui/templates/respondent-search/_inline-form.html
owencjones: Yeah all done, easy enough with autorefactor.

response_operations_ui/common/respondent_utils.py
owencjones: Changed to `enum_lookup`.

response_operations_ui/common/respondent_utils.py
owencjones: 🤦‍♂️

owencjones: Done now.

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +17,84 @@
...
+    return redirect(url_for('respondent_bp.respondent_search',
+                            email_address=form.email_address.data or '',
+                            first_name=form.first_name.data or '',
+                            last_name=form.last_name.data or '',
+                            source=source,
+                            page=page))
+
+
+@respondent_bp.route('/search')
+@respondent_bp.route('/search/', methods=['GET'])
```
insacuri: These 2 lines should become just:
`@respondent_bp.route('/search', methods=['GET'])`

because you get urls that look like `<host>/respondents/search/?email_address=&first_name=Person&last_name=&page=1` otherwise

response_operations_ui/views/respondents.py
```diff
@@ -14,22 +17,78 @@
...
+    form.email_address.data = email_address
+
+    party_response = party_controller.search_respondents(first_name, last_name, email_address, page)
+
+    respondents = party_response.get('data', [])
+    total_respondents_available = party_response.get('total', 0)
+
+    filtered_respondents = filter_respondents(respondents)
+
+    RESULTS_PER_PAGE = app.config["PARTY_RESPONDENTS_PER_PAGE"]
```
insacuri: More minor stuff, this shouldn't be uppercased as no other variables we use are uppercased (even if the config item is)

PR #294, opened 2019-03-05
Added class to form to move back button down

AndrewTorrance:
# Motivation and Context
Cancel button too close the close buttons

# What has changed
Added margin to the form to form so as to push button down . Not added to button as button repeats top and bottom

# How to test?
Run response ops and create message, see increased gap above cancel button


NickGrantONS:
Before & after:
screen shot 2019-03-07 at 08 53 31
screen shot 2019-03-07 at 08 57 02

PR #296, opened 2019-03-12
Use correct algorithm for uaa jwt

insacuri:
Fixes the warning fix from earlier.   For now we can live with the deprecation warning during the unit tests as changing it to be correct causes the unit tests to fail in a 'non-trivial to fix' way.

## How to test
- Fire up this service and attempt to sign in.  You should no longer see an error ending in `raise InvalidAlgorithmError('The specified alg value is not allowed')\njwt.exceptions.InvalidAlgorithmError: The specified alg value is not allowed"`


necrophonic:
Original fix was good but obviously has knock on effects that should be addressed in a more complete fix. Regressing the changes works for now.

PR #297, opened 2019-03-14
Make css changes to use new font sizes

insacuri:
# Motivation and Context
With the updated pattern library, the use of planet names to determine font sizes has been replaced with a set of utility classes that do the same thing.

# What has changed
All the planet names have been changed with their corresponding class name as described in the sdc-global-design-patterns commit that has been pasted below.

# How to test?
Ensure nothing has changed with any of the font sizes in response-operations-ui.  All the changes are a direct translation so the names changed, but not the styles they represented.

# Links
https://trello.com/c/z7YdQ0SA/800-make-css-changes-to-use-new-font-sizes-2
https://github.com/ONSdigital/sdc-global-design-patterns/commit/b002c49d11775c89a384d07dfa2deff81c73e7ff#diff-e78493d7c43ea74a734220671670b8fa


response_operations_ui/templates/conversation-view/conversation-view.html
```diff
@@ -16,7 +16,7 @@
     {% include 'conversation-view/cv-handle-errors.html' %}
     {% include 'conversation-view/cv-back-link.html' %}
 
-    <h1 class="saturn u-mt-s" name="page-messages-title">{{ subject }}</h1>
+    <h1 class="u-fs-l u-mt-s" name="page-messages-title">{{ subject }}</h1>
```
ryangrundy7: The difference in this from master is that the subject box isn't the same length as the message box

response_operations_ui/templates/close-conversation.html
```diff
@@ -4,7 +4,7 @@
   <div role="main" id="main" class="page__main">
     <div class="spoke__header">
       <div class="spoke__header-title">
-        <h1 class="jupiter">Close conversation</h1>
+        <h1 class="u-fs-xl">Close conversation</h1>
```
ryangrundy7: The title now is a lot smaller using this

ryangrundy7: The spoke.html is still using the old cdn hash so it needs to be updated to 2.0.0 to fix it.

insacuri: ah, that'll probably fix a bunch of these issues

response_operations_ui/templates/conversation-view/conversation-view.html
insacuri: Fixed this by adding `text--block` to the input field below

response_operations_ui/templates/response-status.html
```diff
@@ -78,7 +78,7 @@ <h1 class="u-fs-xl">Change response status</h1>
                   <div class="field__item js-focusable-box">
                     <input required class="input input--radio js-focusable" name="event"
                        value="{{ event }}" id="state-{{loop.index}}" type="radio">
-                  <label class="label label--inline venus " for="state-{{loop.index}}">{{ status }}</label>
+                  <label class="label label--inline u-fs-r--b " for="state-{{loop.index}}">{{ status }}</label>
```
ryangrundy7: This is making the checkboxes bold 🎨

insacuri: ^ checkbox text

response_operations_ui/templates/close-conversation.html
owencjones: Yeah, there are two locations (because why wouldn't there be) to change the hash

response_operations_ui/templates/response-status.html
owencjones: Change to `u-fs-r` for no bolding.

response_operations_ui/templates/conversation-view/conversation-view.html
owencjones: 👍

response_operations_ui/templates/response-status.html
insacuri: Fixed

PR #303, opened 2019-03-29
Fix js linting

owencjones:
# Motivation and Context
Javascript had numerous linting issues, and some syntax errors causing lack of functionality.  This is now changed.

# What has changed
* JS Linting errors refactored
* jQuery Removed
* Alert library introduced to use in place of browser native alert functions
* Reworded go live warning to be more professional
* Added tests for Alert library
* Adding JS testing to pipeline CI

# How to test?
Run response ops, and manually check how UI works
Run acceptance tests

# Links
https://trello.com/c/nxqye5q4


response_operations_ui/static/js/read-csv.js
```diff
@@ -1,88 +1,91 @@
...
-    var ciCount;
+            reader.onerror = errorHandler;
 
-    while (allTextLines.length) {
-        var line = allTextLines.shift().split(":")
-        if (line && line.length && line[0]) {
-            lines.push(line);
+            reader.readAsText(files[0]);
+        } else {
+            // @TODO: Implement better way to handle this
```
insacuri: I think this comment should either be expanded to say how it could/should be fixed, or just removed.  I've never liked having TODOs in code, especially ones that don't give a hint on how to do the thing that needs to be done.
Maybe make a trello card too so that it doesn't get forgotten about?

response_operations_ui/static/js/data-panel.js
```diff
@@ -1,30 +1,32 @@
...
+                }
+            });
+        }
+    }
 
+    function openPanel() {
+        const panelName = window.location.hash;
+        if (panelName) {
+            document.querySelectorAll('.data-panel-header').classList.addClass('active');
+            document.querySelectorAll('.data-panel-body').style.display = 'block';
```
saedwards: Wouldn't it be best practice to 'program to interfaces, not implementations?'. Browser environments, whilst they conform better to modern day standards are still seen as volatile environments.

It doesn't have to be jQuery, it could even be a custom implementation for obtaining referenced to DOM instances/collections for something more lightweight (remember you talking about sizzle previously).

response_operations_ui/static/js/validate-ci.js
```diff
@@ -1,40 +1,47 @@
...
+            });
+
+            document.getElementById('btn-load-ci').classList.remove('unready');
+        } else {
+            errorPanel.classList.add('panel');
+            errorPanel.classList.add('panel--simple');
+            errorPanel.classList.add('panel--error');
+
+            errorPanelBody.classList.add('panel__body');
+            paragraphsToAddClassTo = Array.prototype.slice.call(errorPanelBody.querySelectorAll('p'));
```
insacuri: Any value in putting a comment like 
`Equivalent to $("#ciFileErrorPanelBody p").addClass("hidden"); in jquery` ?

gulpfile.js
```diff
@@ -19,6 +19,7 @@ const CONSTANTS = {
     CSS_DIR:         _.get(config, 'CSS_DIR', '').replace('$ROOT', PROJECT_ROOT),
     JS_SRC_DIR:      _.get(config, 'JS_SRC_DIR', '').replace('$ROOT', PROJECT_ROOT),
     JS_DEST_DIR:     _.get(config, 'JS_DEST_DIR', '').replace('$ROOT', PROJECT_ROOT),
+    TESTS_DIR:     _.get(config, 'TESTS_DIR', '').replace('$ROOT', PROJECT_ROOT),
     TASKS_DIR:       join(PROJECT_ROOT, 'gulp', 'tasks'),
     IS_DEBUG:        !Boolean(process.env.node_env === 'production'),
     PROJECT_ROOT
```
saedwards: IDE is informing me that GulpError in the context object is not used and can removed, guess it's being directly imported in other places

response_operations_ui/static/js/read-csv.js
owencjones: Heh, ironically, I've actually fixed what this TODO referred to, and omitted to remove the todo.

Fix on the way.

gulpfile.js
owencjones: Yep, the context object is used to pass a context to subsequent tasks.

response_operations_ui/static/js/data-panel.js
owencjones: Yeah, ideally I would move to some wrapper for `querySelector` with fallback, but to be honest, this stuff was a quick win I tacked on.

Happy if I resolve this by creating a future task to address this - bearing in mind that all this JS is non-vital?

response_operations_ui/static/js/validate-ci.js
owencjones: Probably arguable to make reference to jQuery specifically - but I can see value to commenting this part, given that it uses a slightly ugly method to convert the `NodeList` to a proper `Array`.

Change on the way.

insacuri:
LGTM, but I won't tick approve while we have someone more qualified then myself reviewing this as well ^^

owencjones:
@saedwards Yeah, the move to bundling JS outside of the flask `Bundle` command is one of the next steps along the way, along with transpiling newer versions of ECMA for compatibility etc.

response_operations_ui/static/js/disable-send.js
```diff
@@ -1,5 +1,11 @@
...
+    const createMessageForm = document.getElementById('create-message-form');
+    const buttonSendMessage = document.getElementById('btn-send-message');
+
+    if (createMessageForm && buttonSendMessage) {
+        createMessageForm.addEventListener('submit', function(event) {
+            buttonSendMessage.attributes.disabled = 'disabled';
+        });
+    }
+}(window));
+
```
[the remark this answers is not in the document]
owencjones: Sadly, because the script isn't loaded until late in the markup - plan will be to have an actual DOMContentLoaded handler when we move to a bundled execution.

response_operations_ui/static/js/data-panel.js
owencjones: Agreed, there are various tickets to address this.  A lot of JS functionality on here can actually be removed, but it's a varied iterative process.

Cheers.

PR #306, opened 2019-04-04
Review warnings in resops tests

insacuri:
# Motivation and Context
The unit tests were throwing a lot of warnings when run.  This PR fixes them.

# What has changed
The unit tests were encoding/decoding the jwt differently from how it does it when the service is running (using symmetric keys instead of asymmetric ones used by the service).  The keys are configured in the config.py file and now we can restrict the algorithm used by the jwt module.

Additionally, there were deprecation warnings in the uaa tests.  The way we were testing the exception being raised wasn't working and I had to change the test itself.  Because the result of the function isn't a raised exception but instead is `None` and a logger.exception line, I had to change the thing that's being tested.  We maybe could(should?) change the `request_uaa_public_key` function to `raise` after logging instead of returning `None` on error... but that's a PR for a different time.  Feel free to raise a card if you agree with me whilst reviewing it.

# How to test?
Run the unit tests - Should pass with 0 warnings
Run the app - should work just as before as no functionality should've changed.

# Links
https://trello.com/c/P9hzNOLu/802-spike-review-warnings-in-resops-tests-1-day


insacuri:
Don't merge yet.  I just had a thought about the algorithm used.  I'm certain it's RS256, but I want to be sure that's what it's encoding it with on preprod/prod as it could be RS256 or RS384 or RS512.  I'm fairly sure it's the first one, but I want to be sure.

config.py
```diff
@@ -133,6 +133,18 @@ class TestingConfig(DevelopmentConfig):
     WTF_CSRF_ENABLED = False
     SESSION_TYPE = "filesystem"
     SESSION_PERMANENT = False
-    UAA_PUBLIC_KEY = 'Test'
+    UAA_PUBLIC_KEY = '''-----BEGIN PUBLIC KEY-----
```
AndrewTorrance: Requesting approval of IA

owencjones: I think it's likely fine, but depends where this test key came from.  So long as it's not reused anywhere, or just copied from someone's `~/.ssh/` dir, should be fine.

response_operations_ui/common/token_decoder.py
```diff
@@ -14,6 +14,7 @@ def decode_access_token(access_token):
     decoded_jwt = jwt.decode(
         access_token,
         key=uaa_public_key,
+        algorithms=['RS256'],
```
owencjones: Is this actually a change of algo, or just affirming the use of a different one?

If it's a change, what were we using before?  I assume we went with RS because it's asymmetric?

tests/test_uaa.py
```diff
@@ -25,14 +25,18 @@ def test_get_uaa_public_key_with_no_config_set(self, mock_request):
 
     @requests_mock.mock()
     def test_get_uaa_public_key_server_error_response(self, mock_request):
+        """When the getting of the public key fails a HTTPError exception is raised.  This test, however,
+        assertIsNone because the exception is consumed, an exception is logged and then returns None."""
```
owencjones: Closing `"""` of a block comment should be on a new line. (PEP 257).

Turns out we don't have linting on tests, so I guess it's up to you whether or not you act on this.

owencjones:
@AndrewTorrance Do we need to name drop someone in IA on this PR, WRT your comment?

config.py
insacuri: The key was generated using http://travistidwell.com/jsencrypt/demo/ and isn't used anywhere else.  It's only used for the tests and isn't used anywhere else.
I needed it so I could emulate how the key would be encoded when the code is running for real.

response_operations_ui/common/token_decoder.py
insacuri: This isn't a change of algo.  This change came from a deprecation warning in pyjwt that in a future version, it won't be possible to not define an algorithm when encoding/decoding (you only need to specify when you encode currently I think..)

This change is just explicitly saying what we have always been using.  I tested this in 2 ways, first looking at the payload we get back from uaa when we request the public key, as it tells us the algorithm used.  Second, I added a print statement to my local copy of pyjwt (roughly around https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jws.py#L213 ) to see what it was using.  Originally it was using RS256 when running the service and HS256 in the unit tests, but now it uses RS256 for both.

tests/test_uaa.py
insacuri: I'll change it.  Good spot ^^

response_operations_ui/common/token_decoder.py
owencjones: Seems like a good plan - certainly better than using a different algo in each.  RS is better for this too.

config.py
owencjones: I'd be tempted to either comment to mention that it's a demo key, or to create a zeroed key to make it obvious.

response_operations_ui/common/token_decoder.py
rhiGriff: This should be HS256 as that is what was always used to be, we also use HS256 in frontstage

PR #310, opened 2019-04-16
Update packages to fix jinja security issue

insacuri:
# Motivation and Context
Jinja 2.10 has a security issue.  This is fixed in 2.10.1.  This PR updates jinja and a number of other packages by virtue of `pipenv update`

# What has changed
Hopefully nothing other then the version of packages we're using

# How to test?
Unit and acceptance tests should still pass.  Nothing else should've changed.
It's important that this is tested in the reviewers local environment as there's a chance that one reviewer could catch something that another did not.

# Links

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -40,10 +40,10 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:47f9c83ef4c0c621eaef743f133f09fa8a74a9b75f037e8624f83bd1b6626cb7",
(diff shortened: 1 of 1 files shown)


insacuri:
I tested a bunch of pages by hand and everything seems to be working as expected

insacuri:
It just got mentioned that this PR causes a warning to appear in the unit tests.  There has been a card made to address this and shouldn't hold up this PR as the warning is in flask-session and there's not a lot we can do about it.

PR #317, opened 2019-05-17
Unpin packages in response operations ui

insacuri:
# Motivation and Context
A lot of packages were pinned in the repo for no decent reason.  The pipfile.lock acts as a snapshot of what's currently running in master, so we can unpin the versions safely, knowing that we can revert to a known set of packages if required.

# What has changed
The packages in have all been unpinned except for redis because of a feature that's been removed since version 3.X (we're on 2.10.X)

# How to test?
  - Unit tests should still pass
  - Using the site shouldn't break it
  - Acceptance tests *should* still pass - I couldn't get them to pass, but it was failing on a different thing each time.

# Links
https://trello.com/c/uWG3BbYg/893-unpin-packages-in-response-operations-ui-3


insacuri:
@owencjones I think the second one is better.  It's fairly infrequent where package upgrades break anything, and it's good to know about those breaks little and often instead of in one big go.

owencjones:
@insacuri Good stuff, only really wanted to check this had been considered.

insacuri:
@ryangrundy7 The error isn't one we can fix, as it's to do with how Flask-Session works.  Leaving the warning in keeps it in our minds that we need to do something about it at some point

PR #318, opened 2019-06-03
Js test read csv

owencjones:
# Motivation and Context
Introduce unit testing for CSV reading in frontend

# What has changed
* Refactored `read-csv.js`
* Added tests for `read-csv.js`

# How to test?
* Review code
* Confirm `npm test` tests pass
* Ensure that reading a sample CSV still produces a summary of the CSV selected.

# Links
https://trello.com/c/kkdGNMJM


owencjones:
@jonnyshaw89 Yeah, you probably have a point in places - there are bits where these are testing implementation over function, but one thing I'm trying to do is showcase the testing framework itself,  because people aren't familiar with it, so some of this is a kind of benign showcase.

Cheers for the ✅

PR #335, opened 2019-08-05
Add google analytics

insacuri:
# Motivation and Context
ras-frontstage has google analytics but response-operations doesn't.  We're adding it so that we can start seeing the journeys that internal users take though our systems so we can have more data to figure out what's important.

# What has changed
Added google analytics code snippet

# How to test?
Can't be tested until it hits preprod.  We've added the container value (the GOOGLE_ANALYTICS value) into the preprod build.  Once we've deployed it and verified the analytics are working, we'll need to do the same to prod

# Links
https://trello.com/c/A9wl7wX5/794-add-analytics-to-rops-iv5

### DIFF
--- a/config.py
+++ b/config.py
@@ -9,6 +9,7 @@ class Config(object):
     DEBUG = os.getenv('DEBUG', False)
     TESTING = False
     PORT = os.getenv('PORT', 8085)
+    GOOGLE_ANALYTICS = os.getenv('GOOGLE_ANALYTICS', None)
     LOGGING_LEVEL = os.getenv('LOGGING_LEVEL', 'INFO')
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -9,6 +9,7 @@ class Config(object):
     DEBUG = os.getenv('DEBUG', False)
     TESTING = False
     PORT = os.getenv('PORT', 8085)
+    GOOGLE_ANALYTICS = os.getenv('GOOGLE_ANALYTICS', None)
```
necrophonic: Should this be `GOOGLE_ANALYTICS_ID` for clarity?

insacuri: Yes, but ras-frontstage uses this.  I figured consistency > being correct in this case. Or should I make this right to hopefully make future ones right?

necrophonic: Semantically correct is always nice :)

PR #340, opened 2019-09-05
restrict cases by survey and fix minor error on first page of respond…

AndrewTorrance:
# Motivation and Context
response operations was retrieving all cases for all surveys for an ru , then getting all the data in each case , and then passing all the data to the view where the number displayed was restricted. This change adds a parameter to the case lookup to only retrieve the last n cases per survey so that subsequent look ups are reduced.

Also , during testing I noticed a bug in the respondent search first page. If the number of respondents was less than one page then the showing 'x of y' was incorrect. This was corrected also.

# What has changed
Added a new environment variable MAX_CASES_RETRIEVED_PER_SURVEY . This is defaulted to 12 which is the desired value so there is no current need to implement that in the various environments. Added max_cases_per_survey param to the case lookup.

# How to test?
See case PR for testing details
 

# Links
https://trello.com/c/IRgw9xaj
 
https://github.com/ONSdigital/rm-case-service/pull/114


config.py
```diff
@@ -31,6 +31,7 @@ class Config(object):
     CASE_USERNAME = os.getenv('CASE_USERNAME')
     CASE_PASSWORD = os.getenv('CASE_PASSWORD')
     CASE_AUTH = (CASE_USERNAME, CASE_PASSWORD)
+    MAX_CASES_RETRIEVED_PER_SURVEY = os.getenv('MAX_CASES_RETRIEVED_PER_SURVEY', 12)
```
mfcrocker: I wouldn't set a default on this and instead would add it to our deployment config, but do whatever Chris did for the max survey retrieval thing he did

response_operations_ui/controllers/case_controller.py
```diff
@@ -94,10 +94,11 @@ def get_case_groups_by_business_party_id(business_party_id):
     return response.json()
 
 
-def get_cases_by_business_party_id(business_party_id):
+def get_cases_by_business_party_id(business_party_id, max_case_per_survey):
     logger.info('Retrieving cases', business_party_id=business_party_id)
     url = f'{app.config["CASE_URL"]}/cases/partyid/{business_party_id}'
-    response = requests.get(url, auth=app.config['CASE_AUTH'], params={"iac": "True"})
+    response = requests.get(url, auth=app.config['CASE_AUTH'],
+                            params={"iac": "True", "max_cases_per_survey": max_case_per_survey})
```
mfcrocker: It's odd that we get `app.config['CASE_AUTH']` here but pass in `app.config['MAX_CASES_RETRIEVED_PER_SURVEY']` as a parameter. Would like to see this consistent in one direction or the other

response_operations_ui/views/respondents.py
```diff
@@ -85,7 +85,8 @@ def respondent_search():
                            respondents=filtered_respondents,
                            respondent_count=total_respondents_available,
                            first_index=1 + offset,
-                           last_index=results_per_page + offset,
+                           last_index=(results_per_page + offset) if total_respondents_available >= results_per_page
```
insacuri: This is a bit long, can this be moved to a separate variable above for clarity?

response_operations_ui/controllers/case_controller.py
AndrewTorrance: Disagree on this one . This is consistent across the app. Auth  , headers and params are set distinctly

config.py
AndrewTorrance: You say that but he  literally changed 4 to 12 directly in the view, so I am not doing that . Why not set a default ? Its one less thing to worry about and we can change it at anytime by setting an environment variable. If we add it as an env then we have one more thing to maintain and we do not know if it will ever change ?

response_operations_ui/views/respondents.py
AndrewTorrance: changed

response_operations_ui/controllers/case_controller.py
mfcrocker: I would like to see this changed, tagging @necrophonic for resolution

config.py
mfcrocker: I have a slight worry about this causing unintended behaviours down the line but am happy to let it go for now

PR #345, opened 2019-09-24
Reset password

mfcrocker:
# Motivation and Context
Allows internal users to self-serve password reset changes

# What has changed
Added a number of new pages and controllers
Added the functionality from `ras-party` to generate tokens for password reset emails
Added a number of new environment variables

# How to test?
Difficult to test in dev due to the tie-in to GovNotify


config.py
```diff
@@ -57,6 +68,13 @@ class Config(object):
     PARTY_AUTH = (PARTY_USERNAME, PARTY_PASSWORD)
     PARTY_RESPONDENTS_PER_PAGE = os.getenv('PARTY_RESPONDENTS_PER_PAGE', 25)
 
+    RAS_NOTIFY_SERVICE_URL = os.getenv('RAS_NOTIFY_SERVICE_URL', 'http://notify-gateway-service/emails/')
```
rhiGriff: I would be more inclined to remove the RAS from these variables as we have been moving away from using RAS or RM as a prefix to the variables.

config.py
```diff
@@ -73,6 +91,9 @@ class Config(object):
     UAA_CLIENT_ID = os.getenv('UAA_CLIENT_ID')
     UAA_CLIENT_SECRET = os.getenv('UAA_CLIENT_SECRET')
 
+    EMAIL_TOKEN_SALT = os.getenv('EMAIL_TOKEN_SALT', 'aardvark')
+    EMAIL_TOKEN_EXPIRY = int(os.getenv('EMAIL_TOKEN_EXPIRY', '306000'))
```
rhiGriff: Do we know if the value here is in seconds or miliseconds etc? And can we please put a comment above as to which one it is please? I also don't think that this equates to the 24h shown in the email template.

config.py
mfcrocker: Is our preference towards consistency or moving away from RAS/RM? This is consistent with other services

config.py
mfcrocker: It's in seconds and yeah, completely wrong - missed that A/C :) changing

config.py
insacuri: Usually I'd say consistency trumps all, but changing these from ras_<thing> to <thing> is a pain so we should make that job easier for us when the time comes by having 1 less thing to do.

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,61 @@ def sign_in(username, password):
...
+
+
+def login_admin():
+    headers = {'Content-Type': 'application/x-www-form-urlencoded',
+               'Accept': 'application/json'}
+    payload = {'grant_type': 'client_credentials',
+               'response_type': 'token',
+               'token_format': 'opaque'}
+    try:
+        url = '{}/oauth/token'.format(app.config['UAA_SERVICE_URL'])
```
AndrewTorrance: f strings are preferable to sing format

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,61 @@ def sign_in(username, password):
...
+        logger.exception(f'Failed to log into UAA', status_code=response.status_code)
+        abort(response.status_code)
+
+
+def get_user_by_email(email):
+    headers = {'Content-Type': 'application/json',
+               'Accept': 'application/json',
+               'Authorization': 'Bearer {}'.format(login_admin())}
+
+    url = '{0}/Users?filter=email+eq+%22{1}%22'.format(app.config['UAA_SERVICE_URL'], email)
```
AndrewTorrance: f string ?

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,61 @@ def sign_in(username, password):
...
+    response = requests.get(url, headers=headers)
+    if response.status_code != 200 or response.json()['totalResults'] == 0:
+        pass
```
AndrewTorrance: This pass does not do anything ? Not sure what the intention was , did you mean return ?

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,61 @@ def sign_in(username, password):
...
+
+    user_response = get_user_by_email(email)
+    if user_response is None:
+        return False
+    user_id = user_response['resources'][0]['id']
+
+    payload = {'password': password}
+    url = '{0}/Users/{1}/password'.format(app.config['UAA_SERVICE_URL'], user_id)
+    reset_response = requests.get(url, headers=headers, data=payload)
+
```
AndrewTorrance: might be worth logging response details in case of non 200 ?

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+    return render_template('forgot-password.html', form=form)
+
+
+@passwords_bp.route('/forgot-password', methods=['POST'])
+def post_forgot_password():
+    form = ForgotPasswordForm(request.form)
+    form.email_address.data = form.email_address.data.strip()
+    email = form.data.get('email_address')
+
+    if form.validate():
```
AndrewTorrance: The validate doesnt check if they are users . Will it send an email to non users ?

response_operations_ui/controllers/notify_controller.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Send message to gov.uk notify wrapper
+        :param email: email address of recipient
+        :param template_id: the template id on gov.uk notify to use
+        :param personalisation: placeholder values in the template
+        :param reference: reference to be generated if not using Notify's id
+        :rtype: 201 if success
+        """
+
+        if not self.config['SEND_EMAIL_TO_GOV_NOTIFY']:
```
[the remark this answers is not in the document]
mfcrocker: I think there's too much room here for accidentally sending personally identifiable details to logs by doing this

response_operations_ui/controllers/uaa_controller.py
mfcrocker: I did!

response_operations_ui/views/passwords.py
mfcrocker: It's done in `send_password_change_email` - as commented on that method, to avoid fishing attacks we still return the email sent template in the case that the user doesn't exist, we just don't actually send them an email.

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+@passwords_bp.route('/resend-password-email-expired-token/<token>', methods=['GET'])
+def resend_password_email_expired_token(token):
+    email = token_decoder.decode_email_token(token)
+    return send_password_change_email(email)
+
+
+def send_password_change_email(email):
+    first_name = uaa_controller.get_first_name_by_email(email)
+    if (first_name != ""):
+        internal_url = app.config['RAS_INTERNAL_WEBSITE_URL']
```
AndrewTorrance: I think if we used 
verification_url = url_for('passwords_bp.reset-password', token={token_decoder.generate_email_token(email)}) then we would not have to use the config value at all.

mfcrocker: Wouldn't that return `?token=blahblahblah`, which isn't what we want

It's a good idea though and if we can find a way to do this I'm happy to get rid of the variable altogether

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,64 @@ def sign_in(username, password):
...
+
+def get_first_name_by_email(email):
+    response = get_user_by_email(email)
+    if response is not None:
+        return response['resources'][0]['name']['givenName']
+    return ""
+
+
+def user_exists_by_email(email):
```
insacuri: What uses this?

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,64 @@ def sign_in(username, password):
...
+               'Accept': 'application/json',
+               'Authorization': 'Bearer {}'.format(login_admin())}
+
+    user_response = get_user_by_email(email)
+    if user_response is None:
+        return False
+    user_id = user_response['resources'][0]['id']
+
+    payload = {'password': password}
+    url = '{0}/Users/{1}/password'.format(app.config['UAA_SERVICE_URL'], user_id)
```
insacuri: Rewrite this to `url = f'{app.config['UAA_SERVICE_URL']}/Users/{user_id}/password'`

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,64 @@ def sign_in(username, password):
...
+    reset_response = requests.get(url, headers=headers, data=payload)
+
+    if reset_response.status_code != 200:
+        logger.error('Error received from UAA on change password', status_code=reset_response.status_code)
```
insacuri: Is there a message that gets sent back in the case of a failure?  It's probably worth logging that out as saying there is an error but not saying what it is isn't very helpful for debugging

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,64 @@ def sign_in(username, password):
...
+    if response.status_code != 200 or response.json()['totalResults'] == 0:
+        return
+
+    return response.json()
+
+
+def get_first_name_by_email(email):
+    response = get_user_by_email(email)
+    if response is not None:
+        return response['resources'][0]['name']['givenName']
```
insacuri: Will all of these things be guaranteed to be there?  If givenName is empty, will the key still be there for it?
I only ask because debugging this becomes a pain as there are potentially 3 KeyErrors and 1 IndexError that could occur if the data isn't in the exact shape that you want

response_operations_ui/exceptions/exceptions.py
```diff
@@ -41,3 +41,18 @@ def __init__(self, response, **kwargs):
...
+
+
+class UserDoesNotExist(Exception):
+    def __init__(self, message):
+        super().__init__()
+        self.message = message
+
+
+class NotifyError(Exception):
+
```
insacuri: 🎨 remove extra line

response_operations_ui/controllers/notify_controller.py
```diff
@@ -0,0 +1,60 @@
...
+            "emailAddress": email,
+        }
+        if personalisation:
+            notification.update({"personalisation": personalisation})
+        if reference:
+            notification.update({"reference": reference})
+
+        url = f'{self.notify_url}{str(template_id)}'
+
+        response = requests.post(url, json=notification)
```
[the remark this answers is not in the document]
mfcrocker: Yep, turns out Party hides it in a helper class. Hooray code readability!

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -50,3 +50,64 @@ def sign_in(username, password):
...
+
+def get_user_by_email(email):
+    headers = {'Content-Type': 'application/json',
+               'Accept': 'application/json',
+               'Authorization': 'Bearer {}'.format(login_admin())}
+
+    url = f"{app.config['UAA_SERVICE_URL']}/Users?filter=email+eq+%22{email}%22"
+    response = requests.get(url, headers=headers)
+    if response.status_code != 200 or response.json()['totalResults'] == 0:
+        return
```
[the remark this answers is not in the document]
mfcrocker: Logging is done in the calling method

response_operations_ui/controllers/uaa_controller.py
mfcrocker: Nothing, just thought it was a useful helper class. Happy to remove if you want

response_operations_ui/controllers/uaa_controller.py
mfcrocker: Thought I'd caught all these, thanks!

response_operations_ui/controllers/uaa_controller.py
mfcrocker: Nope, the response is basically just the name of the status code in the tests I've done. https://docs.cloudfoundry.org/api/uaa/version/74.1.0/index.html#change-user-password

response_operations_ui/controllers/uaa_controller.py
mfcrocker: They should be unless the UAA API changes

response_operations_ui/templates/reset-password.html
```diff
@@ -0,0 +1,90 @@
...
+    {% set errorNoMatchTitle = "Your passwords do not match" %}
+    {% set errorNoMatchDescription = "Please check the passwords and try again" %}
+
+    {% set errorStrongerPasswordTitle = "Your password doesn't meet the requirements" %}
+    {% set errorStrongerPasswordDescription = "Please choose a different password" %}
+
+<div class="panel panel--error">
+    <div class="panel__header">
+        <h1 class="panel__title u-fs-r--b">
+            {{ errorNotEnteredTitle if errorType.password[0] == errorNotEnteredTitle }}
```
insacuri: Is this how frontstage does this?  It's not great, but it works so I'm not gonna suggest changing it >_<

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+from response_operations_ui.common import token_decoder
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+BAD_AUTH_ERROR = 'Unauthorized user credentials'
+
+passwords_bp = Blueprint('passwords_bp', __name__, static_folder='static', template_folder='templates')
+
+
+@passwords_bp.route('/forgot-password', methods=['GET'])
```
insacuri: My hero for using single functions for GET and POST. In a few places we handle both in the same function >_<

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+    return render_template('forgot-password-check-email.html', email=email)
+
+
+@passwords_bp.route('/reset-password/<token>', methods=['GET'])
+def get_reset_password(token, form_errors=None):
+    form = ResetPasswordForm(request.form)
+
+    try:
+        duration = app.config['EMAIL_TOKEN_EXPIRY']
+        token_decoder.decode_email_token(token, duration)
```
insacuri: Should this be `token = token_decoder.decode_email_token(token, duration)`?  or do we not care about what the token decodes into?
If it's the latter then this should be `_ = token_decoder.decode_email_token(token, duration)` to show that we're purposefully discarding the output

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+                                                           personalisation=personalisation)
+        except NotifyError:
+            logger.error('Error sending password change request email to Notify Gateway')
+            return render_template('forgot-password-error.html')
+
+        logger.info('Successfully sent password change request email')
+    else:
+        # We stil want to render the template for an email without an account to avoid
+        # people fishing for valid emails
+        logger.info('Requested password reset for email not in UAA')
```
insacuri: Might be worth using that obfuscated email code I mentioned in a previous comment to give this some context

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+        internal_url = app.config['INTERNAL_WEBSITE_URL']
+        verification_url = f'{internal_url}/passwords/reset-password/{token_decoder.generate_email_token(email)}'
+
+        personalisation = {
+            'RESET_PASSWORD_URL': verification_url,
+            'FIRST_NAME': first_name
+        }
+
+        try:
+            NotifyController(app.config).request_to_notify(email=email,
```
insacuri: You don't have to pass the app.config to this anymore

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+def resend_password_email_expired_token(token):
+    email = token_decoder.decode_email_token(token)
+    return send_password_change_email(email)
+
+
+def send_password_change_email(email):
+    first_name = uaa_controller.get_first_name_by_email(email)
+    if first_name != "":
+        internal_url = app.config['INTERNAL_WEBSITE_URL']
+        verification_url = f'{internal_url}/passwords/reset-password/{token_decoder.generate_email_token(email)}'
```
insacuri: Log this out so we can use this functionality in dev!!!111oneone

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+            'RESET_PASSWORD_URL': verification_url,
+            'FIRST_NAME': first_name
+        }
+
+        try:
+            NotifyController(app.config).request_to_notify(email=email,
+                                                           template_name='confirm_password_change',
+                                                           personalisation=personalisation)
+        except NotifyError:
+            logger.error('Error sending password change request email to Notify Gateway')
```
insacuri: Log the message we get form the error here.  Nothing else will log that for you and we'll need it for debugging

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,141 @@
...
+            return render_template('forgot-password-error.html')
+
+        logger.info('Successfully sent password change request email')
+    else:
+        # We stil want to render the template for an email without an account to avoid
```
insacuri: Spelt 'still' wrong

insacuri:
You can get the notifyController's request_to_notify function in a verity of ways:
1) invalid template_id
2) successful sent (test_case_controller shows a good use of the responses library to mock the request)
3) None 201 response, you can test the exception happens and maybe the messages that you get back?

response_operations_ui/views/passwords.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if form.validate():
+        return send_password_change_email(email)
+
+    return render_template('forgot-password.html', form=form, email=email)
+
+
+@passwords_bp.route('/forgot-password/check-email', methods=['GET'])
+def forgot_password_check_email():
+    encoded_email = request.args.get('email', None)
```
insacuri: Small thing.  Remove the `None` on this line as `.get(thing, default)` defaults to None if it can't find it

insacuri:
You can also:
- test the `/reset-password/<token>` GET and POST functions with a dud token.
- the `/forgot-password/check-email` without an email parameter

config.py
```diff
@@ -152,4 +172,5 @@ class TestingConfig(DevelopmentConfig):
 Z5VVFymXN2n+A6UeWAnuO8/E1inhk99dBzKEGdw=
 -----END RSA PRIVATE KEY-----'''
     SECRET_KEY = 'sekrit!'
-
+    SECURITY_USER_NAME = 'username'
```
insacuri: This is minor, but everywhere uses admin/secret as the default security_user_name and security_user_password.

PR #348, opened 2019-10-02
Maintainance banner res ops ui

jins-ons:
# Motivation and Context
Currently when we do a deploy we don't put maintenance banner when we work on response ops.Mostly this is handled by pre-communication with the business but this limits times we can do deployments and potentially causes issues if someone is in the middle of a session.
We need a better strategy for managing response ops downtime.

# What has changed
Few lines of code added :
To access redis new decorator variable session .
Modified Partials/header.html as well to add availability msg variable
Added new test file which include Unit test cases to test the fix .

# How to test?
Manual and unit test passed

# Links
https://trello.com/c/WGSjK35Y


jins-ons:
another branch created to commit the html changes related to this issue
https://github.com/ONSdigital/response-operations-ui/tree/banner-correction
PR
https://github.com/ONSdigital/response-operations-ui/pull/349

PR #352, opened 2019-10-09
Frontend refactor

boxadesign:
# Motivation and Context
An initial refactor of the front end templates to integrate the design system components in the correct way.

# What has changed
- Bash script has been added that loads the specified version of the design system release. It pulls in the zipped code, unpacks it and stores it in the templates directory. This script will run on `make build`, `make load-templates` and as part of the docker build.
- The macros are then available to be used directly in the templates. As the macros are `.njk` there are a couple of filters in there to plug gaps between jinja and nunjucks.
- The main change has been to the `base` templates.
- New "config" templates have been created to handle the footer, navigation, breadcrumbs and sign out. Existing templates have been removed.
- The `surveys.html` and `sign_in.html` template have been re-written to use all required macros as a test to understand what's involved in fully using the design system macros.
- The rest of the templates have had some class name changes and macros added where needed i.e. buttons.

# How to test?
Make sure that all pages render as expected and all elements perform the correct function.


Makefile
```diff
@@ -27,3 +27,6 @@ watch_and_start:
 
 docker: test
 	docker build -t sdcplatform/response-operations-ui:latest .
+
+load-templates:
```
insacuri: Is this essential?  If so, then it should be added to make build by adding it to the line above
`build: load-templates`

scripts/load_templates.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+cd "${DIR}"/.. || exit
+
+DESIGN_SYSTEM_VERSION="13.8.0"
+
+TEMP_DIR=$(mktemp -d)
```
insacuri: If this needs to run before the service is usable, will this work in CF?  I'm not sure if you've got access to a filesystem in it... 
I don't actually know myself, just something that might be an issue?

response_operations_ui/__init__.py
```diff
@@ -14,6 +14,7 @@
 from response_operations_ui.logger_config import logger_initial_config
 from response_operations_ui.user import User
 from response_operations_ui.views import setup_blueprints
+from response_operations_ui.common.jinja_filters import blueprint as filter_blueprint
```
insacuri: Minor point, but couldn't you just name the blueprint in `common.jinja_filters` to `filter_blueprint` so you don't have to use `as` in the import statement?

Makefile
boxadesign: I've added this to make build.

scripts/load_templates.sh
boxadesign: I updated the docker file and it appears to build ok locally now.

response_operations_ui/__init__.py
boxadesign: Done!

boxadesign:
@insacuri I've made a few changes that hopefully sort the issues. I've run `docker build` locally and it seems happy.

Makefile
```diff
@@ -3,6 +3,7 @@ build:
 	rm -rf node_modules
 	npm install
 	npx gulp build
+	pipenv run ./scripts/load_templates.sh
```
mfcrocker: This should probably be `load-templates` so that if we need to change the load-templates process, we only have to update it in one place

response_operations_ui/templates/logout.html
```diff
@@ -1,15 +1,13 @@
 {% extends "layouts/base.html" %}
 
-{% block page_title %} Sign-In | Survey Data Collection {% endblock %}
+{% set page_title = "Sign-In " %}
```
mfcrocker: :art: there's an extra space here

Makefile
boxadesign: Yeah good point.

insacuri:
Just noticed a bug.
If you go to a reporting unit then select a company.  
If you then select a survey they're enrolled in that has someone who's enrolled for it.
If you click 'create message' it redirects you to the create-message page with a bunch of errors on it (like you've tried to press submit without filling in any fields).  This doesn't happen in master

insacuri:
It's also missing all of the survey/ru_ref/etc info too

insacuri:
After playing around with it, everything looks reasonable

response_operations_ui/templates/conversation-view/conversation-view.html
```diff
@@ -9,7 +10,7 @@
 {% endif %}
 {% set ru_id = messages[0]['ru_id'] %}
 {% set survey = messages[0]['survey_id'] %}
-{% block page_title %} {{ subject }} | Survey Data Collection {% endblock %}
+{% set page_title = subject %}
```
insacuri: Doesn't this change the whole page title?

response_operations_ui/templates/conversation-view/cv-reply-form.html
```diff
@@ -19,8 +19,20 @@
...
-            class="btn btn--secondary btn--border u-mt-s u-ml-xs" id="btn-close-conversation" >Close conversation</a>
+        {{
+            onsButton({
+                "text": "Send",
+                "classes": 'u-mt-s',
+                "id": "btn-send-message"
+            })
+        }}   
+        {{
+            onsButton({
```
insacuri: Does this remain as a link rather then a button?

response_operations_ui/templates/create-message.html
```diff
@@ -1,6 +1,7 @@
 {% extends "layouts/base.html" %}
+{% from "components/button/_macro.njk" import onsButton %}
 
-{% block page_title %} Create Message | Survey Data Collection {% endblock %}
+{% set page_title = "Create Message" %}
```
insacuri: Same thing with this page title.  This changes it doesn't it?

response_operations_ui/templates/conversation-view/conversation-view.html
boxadesign: Yes but `| Survey Data Collection` is appended in the base template as it's on every page.

response_operations_ui/templates/conversation-view/cv-reply-form.html
boxadesign: Yes. We pass a url into the button component.

response_operations_ui/templates/create-message.html
boxadesign: As before, it's the same just done in a different way

PR #358, opened 2019-10-15
log out ru_ref

lauraedwards1:
# Motivation and Context

We didn't log out the ru_ref enough in the logging

# What has changed

Where ru_ref available, log it to make debugging easier

# How to test?

# Links

### DIFF
--- a/response_operations_ui/controllers/party_controller.py
+++ b/response_operations_ui/controllers/party_controller.py
@@ -68,6 +68,7 @@ def get_respondent_by_party_id(respondent_party_id):
 
 
 def survey_ids_for_respondent(respondent, ru_ref):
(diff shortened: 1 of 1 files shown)


response_operations_ui/controllers/party_controller.py
```diff
@@ -68,13 +68,15 @@ def get_respondent_by_party_id(respondent_party_id):
 
 
 def survey_ids_for_respondent(respondent, ru_ref):
+    logger.info('Survey ids for respondent', ru_ref=ru_ref)
```
mfcrocker: I'd probably not bother with this log line - the code that calls the method that calls this method logs out `logger.info("Gathering data to view reporting unit", ru_ref=ru_ref)`, which feels good enough to me

PR #360, opened 2019-10-21
Track tab through secure message pages

AndrewTorrance:
# Motivation and Context
Secure message UI was not tracking the tab of origin of an action, so when a user interacted they would be returned to the open tab and often the wrong page. This change was to always return them to the tab of origin on the correct page. This required tab information to be passed through multiple pages that interacted with conversations and messages , not only those involved with conversation tab views.

# What has changed
Changed the tab definition from a series of booleans ( is_open, is_closed, my_conversations and initial) to a single conversation tab variable and allowed it to have values of open, closed, initial and my_conversation. Added unit tests to validate returning to the correct tab and page

# How to test?
Run unit tests . Also can be validated manually by firing up response ops and frontstage and interacting with conversations and messages . One scenario which used the link in the flash message to interact was too complex as a unit test and was accomplished via an acceptance test

# Links
 
https://trello.com/c/mego4FuA
https://github.com/ONSdigital/rasrm-acceptance-tests/compare/secure-message-return-to-tab?expand=1


insacuri:
Tested it manually, seems good.  Just need to do the code review now

tests/views/test_message.py
```diff
@@ -815,3 +818,108 @@ def test_mark_unread_returns_expected_flash_message(self, mock_request, mock_get
...
+        page = 4
+        thread_id = 'fb0e79bd-e132-4f4f-a7fd-5e8c6b41b9af'
+
+        with self.client.session_transaction() as session:
+            session['messages_survey_selection'] = 'Ashe'
+        mock_get_jwt.return_value = "blah"
+        mock_request.get(url_get_thread, json=thread_json)
+        mock_request.get(url_get_surveys_list, json=survey_list)
+        mock_request.patch(url_get_thread, json=thread_json)
+        mock_request.patch(url_get_thread, json=thread_json)
```
insacuri: Is this duplication needed?

response_operations_ui/controllers/message_controllers.py
```diff
@@ -39,6 +39,10 @@ def get_conversation(thread_id):
 def get_conversation_count(params):
     logger.info("Retrieving count of threads")
 
+    params['is_closed'] = 'true' if params['is_closed'] else 'false'
```
insacuri: Does the if work if it's missing?  If there isn't a 'is_closed' key in the dictionary then it'll throw a KeyError.

tests/views/test_message.py
AndrewTorrance: Thanks , good catch

response_operations_ui/controllers/message_controllers.py
AndrewTorrance: Will change to get with default

PR #363, opened 2019-10-23
Removed design system components from gitignore and build process

boxadesign:
# Motivation and Context
Cloudfoundry can't unzip design system components so we need to keep them in the repo.

# What has changed
- Removed the components from `gitignore`
- Removed the `make load-templates` command from Travis build, docker build and `make build`.

# How to test?

Run the service and check there are no template rendering errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,10 +106,6 @@ minimised/
 *.floo
 *.flooignore
 
-# Ignore Design-System templates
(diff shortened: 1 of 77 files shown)


boxadesign:
https://github.com/ONSdigital/response-operations-ui/pull/360

PR #371, opened 2019-11-13
No reply to deleted user

insacuri:
# Motivation and Context
Once users are deleted, we need to make it so that it's not possible to add to the conversation or modify its open/closed status

# What has changed
When a respondent cannot be found, the ability to modify the conversation is replaced with a message informing the user this cannot be done. 
Additional logic has been added to handle the possibility of someone hitting the endpoint with the correct information to bypass this restriction with appropriate error messages appearing.

# How to test?
*To test the removal of the reply functionality*
- Run acceptance tests to seed data
- Add conversation between respondent and internal user. Add 1 or 2 messages each side
- Delete the user `curl -v -d '{"email": "<your-email-here>"}' -H "Content-Type: application/json" -X DELETE http://admin:secret@localhost:8081/party-api/v1/respondents/email`
- Go to the conversation and see the reply box removed.

*To test the additional logic*
- Open templates/conversation-view/conversation-view.html
- Remove the if/else for checking for the deleted user variable
- Go to the conversation, the reply box should be there again
- Attempt to send a message.  You should see an error saying it can't be sent because the user has been deleted

# Links
https://trello.com/c/3TkwE4E8


response_operations_ui/views/messages.py
```diff
@@ -77,20 +77,14 @@ def view_conversation(thread_id):
...
-    try:
-        breadcrumbs = _get_conversation_breadcrumbs(thread_conversation['messages'])
-    except IndexError:
-        breadcrumbs = [
-            {"text": "Messages", "url": "/messages"},
-            {"text": "Unavailable"}
-        ]
+    for message in refined_thread:
+        # This needs to change once we figure out what the name of a missing/deleted user is going to be.
+        if 'None - ' in message['username']:
```
mfcrocker: Will this prevent us from replying to message threads if a different internal ONS user involved in the message thread has their account deleted? That's obviously not functionality we have at the moment but something to consider going forward

response_operations_ui/views/messages.py
```diff
@@ -104,27 +97,33 @@ def view_conversation(thread_id):
...
+                _get_message_json(form, thread_id=refined_thread[0]['thread_id'])
             )
             thread_url = url_for("messages_bp.view_conversation", thread_id=thread_id, page=page,
                                  conversation_tab=conversation_tab) + "#latest-message"
             flash(Markup(f'Message sent. <a href={thread_url}>View Message</a>'))
             return redirect(url_for('messages_bp.view_selected_survey', selected_survey=refined_thread[0]['survey'],
                                     page=page, conversation_tab=conversation_tab))
 
-        except (ApiError, InternalError):
+        except (ApiError, InternalError) as e:
```
AndrewTorrance: This section may read better if it had separate except blocks and a finally instead of  looking at the error type ?

response_operations_ui/views/messages.py
insacuri: Currently only respondents get this message happening.  If an ONS user is deleted/not there, then the call fails in a different, more obvious way

response_operations_ui/views/messages.py
insacuri: Removed class check, but couldn't get it looking right

PR #372, opened 2019-11-14
Search pagination

AndrewTorrance:
# Motivation and Context
Pagination required to deal with large number of results when vague search terms used 

# What has changed
Pagination added to ui , and party search also changed.
Also unrelated was collection exercises not showing dates which was on master and also present on this branch so fixed that at the same time 

# How to test?
Run response ops with new party ( see party pr ) 
Also run ce dates acceptance tests 

# Links
https://trello.com/c/fqC2NEdK


response_operations_ui/controllers/reporting_units_controllers.py
```diff
@@ -11,18 +12,21 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-def search_reporting_units(query):
+def search_reporting_units(query, limit, page):
+    start = datetime.datetime.utcnow()
```
insacuri: Need to remove debugging code

response_operations_ui/templates/reporting-unit-search/reporting-units-search.html
```diff
@@ -0,0 +1,27 @@
...
+{% block main %}
+    <h1>Reporting units</h1>
+    <div class="ru-search">
+
+        <form action="" method="post">
+            {{ form.csrf_token }}
+
+            <div class="field u-mb-s">
+                <label class="label">Type a reference or company name</label>
+                {{ form.query(size=20, class_="ru-search-box input input--text")}}
```
insacuri: is this for both names and ru_refs?  if so, is 20 big enough

response_operations_ui/templates/reporting-unit-search/reporting-units.html
```diff
@@ -0,0 +1,73 @@
...
+    <h1>Reporting units</h1>
+    
+        <div class="ru-search">
+
+        <form action="{{ url_for('reporting_unit_bp.search_redirect') }}" method="post">
+            {{ form.csrf_token }}
+
+            <div class="field u-mb-s">
+                <label class="label">Type a reference or company name</label>
+                {{ form.query(size=20, class_="ru-search-box input input--text")}}
```
insacuri: same question about size as the previous one

response_operations_ui/views/reporting_units.py
```diff
@@ -169,19 +171,62 @@ def edit_contact_details(ru_ref, respondent_id):
...
+        return redirect(url_for('reporting_unit_bp.search_reporting_units', query=query))
 
-    return render_template('reporting-units.html', business_list=business_list, form=form, breadcrumbs=breadcrumbs)
+
+@reporting_unit_bp.route('/search', methods=['GET'])
+@login_required
+def search_reporting_units():
+    search_key_words = request.values.get('query', '')
+    page = request.values.get('page', '1')
+    limit = app.config["PARTY_BUSINESS_RESULTS_PER_PAGE"]
```
insacuri: A future iteration of this should have this be a number sent to us by the client.  But totally fine for the first stab at it

response_operations_ui/templates/reporting-unit-search/reporting-units-search.html
AndrewTorrance: I agree its possibly too low , but it has been 20 for at least several months , its not new

response_operations_ui/controllers/reporting_units_controllers.py
AndrewTorrance: Up to you guys , I left the timing in in case we wanted to splunk alert on elapsed time ?

response_operations_ui/views/reporting_units.py
AndrewTorrance: Good catch , I think your right . I put the setting at the point of entry so it would be easy to change , and then did the wrong thing anyway , Doh!  I think respondent search does something similar , but sets the limit further away from the route handler func in the controller.  Moving respondent to the view so that its trivial to pass them in if we want to add them later

PR #374, opened 2019-11-19
Rename @ru_id to @business_details

insacuri:
# Motivation and Context
In secure-message, we're changing ru_id to business_id to more accurately reflect what it is.  This required us changing the @ru_id part of the payload to a more sensible name.  Instead of @business_id, it's been changed to @business_details as it doesn't only include the id so it's more accurate.

# What has changed
- All instances of @ru_id have been changed to @business_details

# How to test?
- Everything works as it did before.  Unit and acceptance tests still pass

# Links
https://trello.com/c/nct0vF5E

### DIFF
--- a/response_operations_ui/views/messages.py
+++ b/response_operations_ui/views/messages.py
@@ -468,14 +468,14 @@ def _get_to_name(message):
 
 def _get_ru_ref_from_message(message):
     try:
-        return message['@ru_id']['sampleUnitRef']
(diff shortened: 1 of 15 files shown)


response_operations_ui/views/messages.py
```diff
@@ -468,14 +468,14 @@ def _get_to_name(message):
 
 def _get_ru_ref_from_message(message):
```
AndrewTorrance: function name should change as well

PR #378, opened 2019-11-25
Create user

mfcrocker:
# Motivation and Context
This allows trusted people with the administrator password to send account creation emails to new members of staff, taking this away from being a DevOps issue

# How to test?
Spin up the accompanying branch of ras-rm-docker-dev (`new-envs-for-create-account`) and start response-operations. You should be able to go through the create account process by using the default admin password `secret` and by taking the email confirmation link out of the logs


response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+@account_bp.route('/request-new-account', methods=['POST'])
+def post_request_new_account():
+    form = RequestAccountForm(request.form)
+    form.email_address.data = form.email_address.data.strip()
+    email = form.data.get('email_address')
+
+    if form.validate():
+        local_part, domain_part = email.rsplit('@', 1)
```
insacuri: This bit should be put into the CreateAccountForm

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+        logger.info('This is the domain part', domain_part=domain_part)
+        if domain_part != 'ons.gov.uk' and domain_part != 'ext.ons.gov.uk' and domain_part != 'ons.fake':
+            logger.info('Account requested for non-ONS email address')
+            errors = {'email_address': ['Not a valid ONS email address']}
+            return get_request_new_account(form_errors=errors, email=email)
+    
+        password = form.data.get('password')
+        admin_password = app.config['CREATE_ACCOUNT_ADMIN_PASSWORD']
```
insacuri: Do we not validate all the config items on app startup?

config.py
```diff
@@ -90,6 +94,10 @@ class Config(object):
     EMAIL_TOKEN_SALT = os.getenv('EMAIL_TOKEN_SALT', 'aardvark')
     # 24 hours in seconds
     EMAIL_TOKEN_EXPIRY = int(os.getenv('EMAIL_TOKEN_EXPIRY', '86400'))
+    # 72 hours in seconds
+    CREATE_ACCOUNT_EMAIL_TOKEN_EXPIRY = int(os.getenv('CREATE_ACCOUNT_EMAIL_TOKEN_EXPIRY', '259200'))
```
insacuri: Just thinking about it, is there a reason why we wanted 72hrs for the expiry of this token?  24 seems like long enough and means we don't have to add another config item

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -136,3 +125,30 @@ def change_user_password(email, password):
...
                            new_password=password)
+
+
+def create_user_account(email, password, user_name, first_name, last_name):
+    access_token = login_admin()
+
+    headers = {'Content-Type': 'application/json',
+               'Accept': 'application/json',
+               'Authorization': f'Bearer {access_token}'}
+
```
insacuri: Trailing whitespace

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -136,3 +125,30 @@ def change_user_password(email, password):
...
+            "value": email,
+            "primary": True
+        }],
+        "active": True,
+        "verified": True,
+        "password": password
+    }
+
+    url = f"{app.config['UAA_SERVICE_URL']}/Users"
+    return requests.post(url, data=dumps(payload), headers=headers)
```
insacuri: Can we have the same `raise_for_status` error handling that we normally have around controller code here?

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+        if password != admin_password:
+            logger.info('Invalid password provided for create account')
+            errors = {'password': ['Incorrect admin secret']}
+            return get_request_new_account(form_errors=errors, email=email)
```
insacuri: I've seen a bunch of this and I don't think it's the best way to do it.  I don't think (though if you can show me counter examples then I'd happily have my mind changed) that having this code call a function which is actually the code for the endpoint is good practice.  
The function signature looks weird and I think we'll have other problems in the future if a change was made to an endpoint that's being called this way, now we have to trust that the functionality isn't being broken anywhere else.

With the number of times you're calling this, it'd be better to have a `_handle_errors` function (or something less terribly named ^^) where you do basically the same thing that you do in get_request_new_account but have it not be a function that correlates to an endpoint.  That function can return a template and you'd have the same functionality but less risk for future development

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+
+        return send_create_account_email(email)
+
+    errors = {'email_address': ['Not a valid ONS email address']}
+    return get_request_new_account(form_errors=errors, email=email)
+
+
+@account_bp.route('/account-exists', methods=['GET'])
```
insacuri: Why does this exist?  Instead of having a whole new endpoint for an error, you can just `return render_template('request-new-account-exists.html', email=email)` instead of the redirect code

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+    try:
+        email = URLSafeSerializer(app.config['SECRET_KEY']).loads(encoded_email)
+    except BadSignature:
+        logger.error('Unable to decode email from URL', encoded_email=encoded_email)
+        abort(404)
+
+    return render_template('request-new-account-exists.html', email=email)
+
+
+@account_bp.route('/request-new-account-check-email', methods=['GET'])
```
insacuri: Same comment as the new_account_exists endpoint.  This doens't need to exist.  Instead of redirecting to a new endpoint, you can just render this template

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -78,18 +78,7 @@ def get_user_by_email(email, access_token=None):
...
-
-    return response.json()
-
-
-def get_first_name_by_email(email):
-    response = get_user_by_email(email)
-    if response is not None:
-        return response['resources'][0]['name']['givenName']
-    return ""
+    return requests.get(url, headers=headers)
```
insacuri: We should have the `raise_for_status` error handling code around this.  I think the pattern is that the error from the api call should be at least caught in the controller and then the appropriate thing is done with that error (whatever that is)

response_operations_ui/forms.py
```diff
@@ -242,3 +242,38 @@ def validate_password(form, field):
         if password.isalnum() or not any(char.isupper() for char in password) or not any(char.isdigit() for char in
                                                                                          password):
             raise ValidationError('Your password doesn\'t meet the requirements')
+
+
+class RequestAccountForm(FlaskForm):
```
insacuri: This should have a docstring saying what it's used for.  Took me a look around the codebase to figure that out

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+    try:
+        email = URLSafeSerializer(app.config['SECRET_KEY']).loads(encoded_email)
+    except BadSignature:
+        logger.error('Unable to decode email from URL', encoded_email=encoded_email)
+        abort(404)
+
+    return render_template('request-new-account-check-email.html', email=email)
+
+
+def send_create_account_email(email):
```
insacuri: Docstring giving a description of what it does would be good

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,204 @@
...
+
+    if response is not None:
+        if response.status_code == 201:
+            logger.info('Successfully created user account', token=token)
+            send_confirm_created_email(email, first_name)
+            flash('Account successfully created', category='account_created')
+            return redirect(url_for('sign_in_bp.sign_in'))
+        if response.status_code == 409:
+            # Username already exists
+            form_errors = {'user_name': ["Username already in use; please choose another"]}
```
insacuri: It's probably not an issue because it's an internal system, but we're okay giving the user information about what's valid and what isn't?  I ask because it's not something we'd do for an externally facing site as we'd want to give something probing the site as little info as possible.

response_operations_ui/views/passwords.py
```diff
@@ -127,8 +127,14 @@ def resend_password_email_expired_token(token):
...
 
-    first_name = uaa_controller.get_first_name_by_email(email)
-    if first_name != "":
+    response = uaa_controller.get_user_by_email(email)
+    if response.status_code != 200:
+        logger.error('Error retrieving user from UAA', status_code=response.status_code,
+                     email=url_safe_serializer.dumps(email))
+        return render_template('forgot-password-error.html')
+
+    if response.json()['totalResults'] > 0:
```
insacuri: What happens if totalResults isn't in the dict?  Currently it'd throw a `TypeError`, log a `A generic exception happened` line, then show the error-500 page.

response_operations_ui/views/passwords.py
```diff
@@ -127,8 +127,14 @@ def resend_password_email_expired_token(token):
...
+        first_name = response['resources'][0]['name']['givenName']
```
insacuri: Same with this, on an IndexError or KeyError it'd give a generic exception line and render the error-500 page.  We should handle this exception more gracefully

response_operations_ui/views/passwords.py
mfcrocker: It's not possible for this key to be missing on a `200` response from the UAA API

response_operations_ui/views/passwords.py
mfcrocker: As above

response_operations_ui/views/accounts.py
mfcrocker: I disagree - this is absolutely information we'd need to give on an externally-facing site if someone chose a duplicate username - the process can't continue with a duplicate and we need to inform the user as to why. Avoiding site probing is always an aim but not always achievable.

config.py
mfcrocker: The business case is that because the email is being requested by someone other than the person who'll actually click through the link, there'll be more of a lag for communicating the need to go set up their account etc.

We can chat about how long to set this with Mark etc if you'd like but there are already going to be unavoidable config changes as part of this card and one more isn't a huge issue

response_operations_ui/views/accounts.py
mfcrocker: I believe if one is missing that doesn't have a default set (like in the case of `CREATE_ACCOUNT_ADMIN_PASSWORD`) then the app will error out on startup

response_operations_ui/views/accounts.py
insacuri: On reflection this is kinda obvious.  We can't just say 'no account... no we're not going to tell you why'

response_operations_ui/views/passwords.py
insacuri: Oki dokes.  The stacktrace would let us know what happened if that element magically disappeared anyway

config.py
insacuri: No need to do this.

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -79,19 +80,17 @@ def get_user_by_email(email, access_token=None):
 
     url = f"{app.config['UAA_SERVICE_URL']}/Users?filter=email+eq+%22{email}%22"
     response = requests.get(url, headers=headers)
-    if response.status_code != 200 or response.json()['totalResults'] == 0:
+    try:
+        response.raise_for_status()
+    except HTTPError:
+        url_safe_serializer = URLSafeSerializer(app.config['SECRET_KEY'])
+        logger.error('Error retrieving user from UAA', status_code=response.status_code,
+                     email=url_safe_serializer.dumps(email))
```
insacuri: Not allowed to log emails

mfcrocker: This is scrambled

response_operations_ui/forms.py
```diff
@@ -242,3 +242,43 @@ def validate_password(form, field):
...
+                                                   message='Your email must be less than 254 characters')])
+    password = PasswordField('Enter the admin password',
+                             validators=[InputRequired('Enter the admin password')])
+
+    @staticmethod
+    def validate_email_address(_, field):
+        email = field.data
+        _validate_email_address(email)
+        local_part, domain_part = email.rsplit('@', 1)
+        if domain_part != 'ons.gov.uk' and domain_part != 'ext.ons.gov.uk' and domain_part != 'ons.fake':
```
insacuri: Missed this on the first time around.  A cleaner way to do it would be:
```suggestion
        if domain_part not in ['ons.gov.uk', 'ext.ons.gov.uk', 'ons.fake']:
```

Second, the ons.fake part of this is a testing domain that could change at any time.  I think it's wrong to hard code test domain in this but I'm not sure what's correct.  Maybe put it in the testing config?  Have a think, it's wrong but I'm not gonna block the review on this cause I don't know what right looks like

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,163 @@
...
+
+        try:
+            NotifyController().request_to_notify(email=email,
+                                                 template_name='request_create_account',
+                                                 personalisation=personalisation)
+        except NotifyError as e:
+            logger.error('Error sending create account request email to Notify Gateway', msg=e.description)
+            return render_template('request-new-account-error.html')
+
+        logger.info('Successfully sent create account request email', email=url_safe_serializer.dumps(email))
```
insacuri: Both of these log lines need an `encoded_email` instead of `email` key

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,164 @@
...
+    }
+    return render_template('create-new-account.html', form=form, data=template_data)
+
+
+@account_bp.route('/create-account/<token>', methods=['POST'])
+def post_create_account(token):
+    form = CreateAccountForm(request.form)
+
+    if not form.validate():
+        return get_create_account(token, form_errors=form.errors)
```
insacuri: This is the same thing as the other thing we talked about.  This shouldn't be using another routes function to do its job.  On the error, get all the bits you need and render the error template (you don't need the token verification functionality that's in the get_create_account function)

insacuri:
I've got a few thoughts about the presentation of this (just testing it now)
1) The create account button should be closer to the bottom of the screen then the top
2) The admin password field could maybe do with some smaller grey help text around it saying what it is and where to get it?
3) Is the idea that the supervisor (or the password holder) is going to type it in themselves?  Otherwise everyone will eventually know the password and it'll provide barely any security at all

insacuri:
I also got an error when creating the account. After filling in the username/name/password fields,   there was 403 but other then that there was 0 logging around it.  There needs to be better logging, and also I don't know how to fix this issue >_<

mfcrocker:
1. I agree and raised this but it was decided by UX
2. I'm not against this idea but again, wasn't in the designs
3. The idea is this is a stopgap where a trusted adult knows the password and will request an account on behalf of new starters. They won't share the password out and eventually this'll be a permissions-level action

mfcrocker:
This will now need the changes @rhiGriff made in her PR to work properly

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,181 @@
...
```
insacuri: This still isn't fixed.

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+@account_bp.route('/resend_account_email_expired_token/<token>', methods=['GET'])
+def resend_account_email_expired_token(token):
+    email = token_decoder.decode_email_token(token)
+    return send_create_account_email(email)
+
+
+@account_bp.route('/create-account/<token>', methods=['GET'])
+def get_create_account(token, form_errors=None):
```
insacuri: Once the last thing stops using this function, this function won't need `form_errors` anymore right?  By GETing this page, theres no errors to render as far as I'm aware?  You can also make the template data in this function tidier too I think

response_operations_ui/views/accounts.py
```diff
@@ -0,0 +1,181 @@
...
+from response_operations_ui.common import token_decoder
+from response_operations_ui.exceptions.exceptions import NotifyError
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+account_bp = Blueprint('account_bp', __name__, static_folder='static', template_folder='templates')
+
+
+@account_bp.route('/request-new-account', methods=['GET'])
+def get_request_new_account(form_errors=None, email=None):
```
insacuri: Is there a need to create the form and add the template data on a get?

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -136,3 +135,43 @@ def change_user_password(email, password):
...
+    try:
+        response.raise_for_status()
+        return
+    except HTTPError:
+        if response.status_code == 409:
+            # Username already exists
+            errors = {'user_name': ["Username already in use; please choose another"]}
+        else:
+            errors = {'status_code': response.status_code, 'message': response.reason}
+            logger.error('Received an error when creating an account in UAA',
```
insacuri: You should add the reason to the logger.error here.  The way the code currently works is the reason only ever makes it to the user (and never the person looking at the logs) if the error has (user_name) in it.  Adding this will aid in debugging

insacuri:
The failing status on here makes me realise that there's probably tests needing to be written for this

boxadesign:
The templates will need to be refactored to use components in the new way. I can do this to speed things up?

mfcrocker:
Happy to add more testing if it'll add value, though it's worth noting that my changes to `password.py` are boy-scouting only

response_operations_ui/forms.py
```diff
@@ -249,3 +249,43 @@ def validate_password(form, field):
...
+                                                   message='Your email must be less than 254 characters')])
+    password = PasswordField('Enter the admin password',
+                             validators=[InputRequired('Enter the admin password')])
+
+    @staticmethod
+    def validate_email_address(_, field):
+        email = field.data
+        _validate_email_address(email)
+        local_part, domain_part = email.rsplit('@', 1)
+        if domain_part not in ['ons.gov.uk', 'ext.ons.gov.uk', 'ons.fake']:
```
AndrewTorrance: Cant decide if this should be a config item or not . Would a non ons email be useful during testing ? If so then may be better as a config item ?

response_operations_ui/forms.py
```diff
@@ -249,3 +249,43 @@ def validate_password(form, field):
...
+            logger.info('Account requested for non-ONS email address')
+            raise ValidationError('Not a valid ONS email address')
+
+
+class CreateAccountForm(FlaskForm):
+    first_name = StringField('First name', validators=[DataRequired(message='First name is required')])
```
AndrewTorrance: Validators should have a max size as well as data required ?

response_operations_ui/forms.py
rhiGriff: I don't think this needs to be config for this we only want certain domains so this is fine.

response_operations_ui/forms.py
rhiGriff: They don't need a max size but if used for a field previously they should be consistent

response_operations_ui/forms.py
```diff
@@ -249,3 +249,43 @@ def validate_password(form, field):
...
+    last_name = StringField('Last name', validators=[DataRequired(message='Last name is required')])
+    user_name = StringField('User name', validators=[DataRequired(message='Username is required')])
+    password = PasswordField('New password',
+                             validators=[DataRequired(message='Password is required'),
+                                         EqualTo('password_confirm', message='Your passwords do not match'),
+                                         Length(min=8, max=160,
+                                                message='Your password doesn\'t meet the requirements')])
+
+    password_confirm = PasswordField('Re-type new password')
```
rhiGriff: This should probably be a required field too

mfcrocker: It'll fail validation on passwords not matching so making this field required is unnecessary

PR #381, opened 2019-12-02
Filter by ru

AndrewTorrance:
# Motivation and Context
There is a need to be able to filter by ru_ref in secure message screens

# What has changed
Passed around both ru_ref and business_id, because secure message uses business id and the ui mandated ru_ref. If only ru ref was passed around it would mean an extra call to party on each tab change. A new compoenent for ru entry was added in the design system . And the extra params added to Secure message calls. As this was being written Secure message was also changing to replace ru_ref with business_id which also changed some of the returned data , this was integrated also .
Several acceptance tests deleted , and replaced with new unit tests in this repo. 

# How to test?
run tests 
Run acceptance tests . Then modify the db to have all the  same survey_id  by modifying the db locally . Then enter an ru ref in the screen and verify that the filter works as expected.Also verify all secure message functionality manually as this has touched every secure message endpoint .

# Links
https://trello.com/c/Zcr7STCe


response_operations_ui/controllers/message_controllers.py
```diff
@@ -61,12 +59,23 @@ def get_conversation_count(params):
...
 
-def get_thread_list(params):
-    logger.info("Retrieving threads list")
+def _get_secure_message_threads_params(survey_id, business_id, conversation_tab):
+    """creates a params dictionary"""
+    params = {'survey': survey_id,
+              'is_closed': 'true' if conversation_tab == 'closed' else 'false',
+              'my_conversations': 'true' if conversation_tab == 'my messages' else 'false',
+              'new_respondent_conversations': 'true' if conversation_tab == 'initial' else 'false'}
+    if business_id:
```
insacuri: Would it be reasonable to just add the business_id even if it's empty?  I'm probably wrong, but this seems oddly verbose and that would tidy it up a bit

response_operations_ui/controllers/message_controllers.py
```diff
@@ -61,12 +59,23 @@ def get_conversation_count(params):
...
+        params['business_id'] = business_id
+    return params
+
 
-    params['is_closed'] = 'true' if params.get('is_closed') else 'false'
-    params['my_conversations'] = 'true' if params.get('my_conversations') else 'false'
-    params['new_respondent_conversations'] = 'true' if params.get('new_respondent_conversations') else 'false'
+def get_thread_list(survey_id, business_id, conversation_tab, page, limit):
+
+    logger.info("Retrieving threads list")
```
insacuri: More metadata here would be useful.  Any/all of survey_id, business_id and conversation_tab would all aid in debugging

response_operations_ui/controllers/party_controller.py
```diff
@@ -31,6 +31,15 @@ def get_party_by_ru_ref(ru_ref):
     return response.json()
 
 
+def try_get_party_id_by_ru_ref(ru_ref):
+    """Attempts to get party id from ru_ref, does not error if not found"""
```
insacuri: So what does it return if found?  This would benefit from sphinx style docstrings:
```
"""Calls party service to get the party_id of a business based on its ru_ref
:param ru_ref: An 11 digit string referring to a business
:returns: A party_id if the ru_ref is found.  An empty string otherwise.
:raises KeyError: Raised if the response from party is successful but doesn't contain an 'id' key.
"""
```

response_operations_ui/controllers/party_controller.py
```diff
@@ -31,6 +31,15 @@ def get_party_by_ru_ref(ru_ref):
     return response.json()
 
 
+def try_get_party_id_by_ru_ref(ru_ref):
```
insacuri: This function name isn't great.  But as well as that, why are you creating a whole new function for something that already exists?  There's a perfectly functional `get_party_by_ru_ref` function that does all your error handling for you.  You also use this function only once! You can use the other function and then extract the id or the empty string that way.
I get that you need to deal with the error case as you don't want it to go bang if it's not found.  I'll describe what to do in that bit

response_operations_ui/views/messages.py
```diff
@@ -206,29 +242,39 @@ def view_selected_survey(selected_survey):
...
+        ru_ref_filter = request.args.get('ru_ref_filter', default='')
+        # business_id = request.args.get('business_id', default='')
+        business_id_filter = request.args.get('business_id_filter', default='')
+        
+        form = SecureMessageRuFilterForm()
+
+        if form.validate_on_submit():
+            new_ru_ref = form.ru_ref_filter.data
+            if new_ru_ref and new_ru_ref != ru_ref_filter:
+                business_id_filter = party_controller.try_get_party_id_by_ru_ref(new_ru_ref)
```
insacuri: Here you should use the original `get_party_by_ru_ref` function that is tested and does error handling for you.
For this I know you don't want it go to bang if it's not found, you can catch the error and handle it appropriately.  Somthing like
```
try:
    business_id_filter = party_controller.get_party_id_by_ru_ref(new_ru_ref)
except ApiError as e:
    if . e.status_code == 404:
        pass #I think pass lets you just carry on and not care?
    else:
         raise
```

That might need a bit of tweaking to be more terse, but you can just handle the error, check if it's a 404, and not care if it is.

response_operations_ui/templates/components/search/_macro.njk
```diff
@@ -2,17 +2,19 @@
     {% from "components/input/_macro.njk" import onsInput %}
     {% from "components/button/_macro.njk" import onsButton %}
 
-    {% set labelPlaceholder = ' label--placeholder ' if params.accessiblePlaceholder %}
-    {% set inputPlaceholder = ' input--placeholder ' if params.accessiblePlaceholder %}
-    {% set inputClasses = params.input.classes if params.input.classes %}
+    {% set labelPlaceholder = ' label--placeholder ' if params.accessiblePlaceholder else "" %}
```
insacuri: doesn't it automatically get set to an empty string if it's not true?

response_operations_ui/templates/close-conversation.html
```diff
@@ -26,7 +26,7 @@ <h1 class="u-fs-xl">Close conversation</h1>
           Reference:
         </dt>
         <dd class="">
-          {{ru_ref}}
+          {{display_ru_ref}}
```
insacuri: I think it'd be fine to leave this as is and just add the ru_ref_filter to distinguish between the two?  Just an opinion

response_operations_ui/controllers/message_controllers.py
AndrewTorrance: Disagree , these are the params passed to secure message and it currently expects params to be populated if they are designated. This was repeated code called from two places , so the current one makes the code Dryer than it was

response_operations_ui/templates/messages.html
```diff
@@ -1,45 +1,84 @@
 {% extends "layouts/base.html" %}
-{% set page_title = "Messages" %}
+{% from "components/search/_macro.njk" import onsSearch %}
 
+{% set page_title = "Messages" %}
+{% set hide_breadcrumbs = true %}
```
insacuri: Why are we hiding the breadcrumbs for this?  I've not fired up the code or looked at the design so it might be obvious but it seems wrong to do so as they're meant to be a navigation aid

response_operations_ui/controllers/message_controllers.py
AndrewTorrance: Added

response_operations_ui/views/messages.py
```diff
@@ -206,29 +242,39 @@ def view_selected_survey(selected_survey):
         limit = request.args.get('limit', default=10, type=int)
         flash_message = request.args.get('flash_message', default="", type=str)
         conversation_tab = request.args.get('conversation_tab', default='open')
-
-        thread_count = message_controllers.get_conversation_count(
-            {'survey': survey_id,
-             'is_closed': conversation_tab == 'closed',
-             'my_conversations': conversation_tab == 'my messages',
-             'new_respondent_conversations': conversation_tab == 'initial'})
+        # ru_ref = request.args.get('ru_ref', default='')
```
insacuri: Need to remove commented out code

response_operations_ui/controllers/party_controller.py
AndrewTorrance: I dont mind documenting when needed , but in this case the function name is pretty descriptive. 
Have modified the doc string to say returned value if not found

response_operations_ui/controllers/party_controller.py
AndrewTorrance: Because get_party_by_ru_ref  raises api errors and I did not want that in this case. The user can enter any 11 digit string and they are not gaurenteed to be an ru ref.

response_operations_ui/views/messages.py
AndrewTorrance: Nope, disagree , see reasoning in comment above

response_operations_ui/templates/components/search/_macro.njk
AndrewTorrance: One for Matt Simon , I dont know how those files are generated

response_operations_ui/templates/close-conversation.html
AndrewTorrance: Your right, this was changed to this as at one point I was not using ru_ref_filter but ru_ref, But since I changed to use _fliter variables , this can also change back for clarity

response_operations_ui/templates/messages.html
AndrewTorrance: Turned off by Matt Simon as part of using design system , breadrumbs implemented via design system

response_operations_ui/views/messages.py
AndrewTorrance: Thanks , dev stuff left in by mistake

response_operations_ui/controllers/message_controllers.py
AndrewTorrance: Added

response_operations_ui/controllers/message_controllers.py
```diff
@@ -36,12 +36,10 @@ def get_conversation(thread_id):
         raise ApiError(response)
 
 
-def get_conversation_count(params):
+def get_conversation_count(survey_id, conversation_tab, business_id):
     logger.info("Retrieving count of threads")
```
insacuri: I know you didn't add this line, but can you add some metadata to the log line like you did in the other function)

response_operations_ui/controllers/message_controllers.py
insacuri: yeah that's fair

response_operations_ui/views/messages.py
insacuri: Even after a while away I still think having 2 functions that do basically the same thing is wrong.  I updated the example to be cleaner.  If you still disagree then a 3rd party might be better to decide.  If they think my way isn't right, then your new function needs to handle 400/5XX errors as it currently doesn't do that.
```
try:
    business_id_filter = party_controller.get_party_id_by_ru_ref(new_ru_ref)
except ApiError as e:
    if e.status_code != 404:
        raise
    business_id_filter = ""
```

response_operations_ui/views/messages.py
```diff
@@ -206,29 +242,37 @@ def view_selected_survey(selected_survey):
...
+
+        if form.validate_on_submit():
+            new_ru_ref = form.ru_ref_filter.data
+            if new_ru_ref and new_ru_ref != ru_ref_filter:
+                business_id_filter = party_controller.try_get_party_id_by_ru_ref(new_ru_ref)
+                if business_id_filter:
+                    ru_ref_filter = new_ru_ref
+                else:
+                    ru_ref_filter = ''
+                    flash_message = f"Filter not applied: {new_ru_ref} is an unknown RU ref"
```
insacuri: Bit nitpicky but this wording could probably be improved.  Maybe something like `{new_ru_ref} not found in system` is a bit more succinct?

insacuri:
Just 2 things to mention, not problems but things to consider:
- If you put in a partially completed ru_ref into the url and press enter then you get the result of a not found without any of the errors
- The url for the secure message inbox should be cleaner when you're not doing anything.  When you just view a single survey there are 3 params and the url is starting to look like a bit of a dumping ground.  Maybe it'd be worth only including those params if they're filled and leaving them empty otherwise.  I know it's only a url but they should be clean where possible I feel

response_operations_ui/views/messages.py
jterryons: I have to agree with Adam here, we shouldnt be creating an identical method without error handling. Devs will end up using it, if accidentally, and that will cause us more problems later on.

AndrewTorrance: Its been moved to be a ui functin that calls the existing endpoint and handles any exceptions it throws

response_operations_ui/views/messages.py
```diff
@@ -206,29 +242,37 @@ def view_selected_survey(selected_survey):
...
+        ru_ref_filter = request.args.get('ru_ref_filter', default='')
+        business_id_filter = request.args.get('business_id_filter', default='')
+
+        form = SecureMessageRuFilterForm()
+
+        if form.validate_on_submit():
+            new_ru_ref = form.ru_ref_filter.data
+            if new_ru_ref and new_ru_ref != ru_ref_filter:
+                business_id_filter, ru_resolution_error = _try_get_party_id_from_filter_ru(new_ru_ref)
+                if business_id_filter:
```
warrenbailey: why do we need the party id here?

response_operations_ui/views/messages.py
```diff
@@ -206,29 +242,37 @@ def view_selected_survey(selected_survey):
...
-             'my_conversations': conversation_tab == 'my messages',
-             'new_respondent_conversations': conversation_tab == 'initial'})
+        ru_ref_filter = request.args.get('ru_ref_filter', default='')
+        business_id_filter = request.args.get('business_id_filter', default='')
+
+        form = SecureMessageRuFilterForm()
+
+        if form.validate_on_submit():
+            new_ru_ref = form.ru_ref_filter.data
+            if new_ru_ref and new_ru_ref != ru_ref_filter:
```
warrenbailey: is this 'and' necessary, wondering if you could just reuse the ru_ref_filter and check its not empty?

response_operations_ui/views/messages.py
```diff
@@ -276,14 +325,23 @@ def close_conversation(thread_id):
     conversation_tab = request.args.get('conversation_tab')
     page = request.args.get('page')
 
+    ru_ref_filter = request.args.get('ru_ref_filter')
+
```
warrenbailey: Are there filters needed in the close conversation view?

warrenbailey: seems they are needed to maintain the filter between all secure message windows

response_operations_ui/views/messages.py
warrenbailey: it's retained here to reduce the number of calls on party

PR #389, opened 2020-01-02
Render template on POST

warrenbailey:
# Motivation and Context

As it currently stands a search for a respondent uses a POST request that redirects to a GET request, in order to pass the parameters from the form data the request appends the first name, last name and email to the get request as url parameters.

Instead let's render the template on POST request as it has all the data it needs at that point anyway.

# What has changed

Modified the respondents.py file and removed the redirection.

# How to test?

Start the respondent operations UI and attempt to search for a user by email. This functionality should continue to work but the inputted data is no longer in the url. 

# Links

https://trello.com/c/rnMDzAee/1423-stop-email-address-being-passed-as-a-url-parameter-in-rops-3

### DIFF
--- a/response_operations_ui/views/respondents.py
+++ b/response_operations_ui/views/respondents.py
@@ -35,29 +35,14 @@ def search_redirect():
         flash('At least one input should be filled')
         return redirect(url_for('respondent_bp.respondent_home'))
 
-    return redirect(url_for('respondent_bp.respondent_search',
-                            email_address=form.email_address.data or '',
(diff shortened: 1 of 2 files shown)


warrenbailey:
Apparently today is the only day that this test will fail: https://github.com/ONSdigital/response-operations-ui/blob/883c963ab03f713259d14b60819074fd30bc70b6/tests/test_dates.py#L24

It ends up being `Yesterday at 00:00` rather than `01 Jan 2020 00:00`

PR #391, opened 2020-01-13
Added bold to table row if unread

boxadesign:
# Motivation and Context
Messages table row for unread messages needed to bold the text

# How to test?
Check the message table and see that unread message table rows are bold


warrenbailey:
Just running the acceptance tests

PR #395, opened 2020-01-27
No enrolements

jterryons:
# Motivation and Context
Resolves https://trello.com/c/e5mHbNx8/1484-ritm0307110-re-instate-deleted-account-15-01-20

# What has changed
if they have no enrolments then they should not see any, but instead they are getting a 500 response. This is because the 204 response code lives within a catch block and given that a 204 is a valid response then response.raise_for_status() never throws. The code attempts to json parse nothing which throws the 500 that they are seeing.

# How to test?
try to view enrolments for respondent with no enrolments. you used to get a 500, now you get a screen with no enrolments

# Links
https://trello.com/c/e5mHbNx8/1484-ritm0307110-re-instate-deleted-account-15-01-20

### DIFF
--- a/response_operations_ui/controllers/case_controller.py
+++ b/response_operations_ui/controllers/case_controller.py
@@ -83,10 +83,10 @@ def get_case_groups_by_business_party_id(business_party_id):
 
     try:
         response.raise_for_status()
-    except requests.exceptions.HTTPError:
         if response.status_code == 204:
             logger.info('No case groups found for business', party_id=business_party_id)
             return []
+    except requests.exceptions.HTTPError:
         logger.exception('Failed to retrieve case groups', business_party_id=business_party_id)
         raise ApiError(response)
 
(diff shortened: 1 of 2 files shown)


response_operations_ui/controllers/case_controller.py
```diff
@@ -103,8 +103,11 @@ def get_cases_by_business_party_id(business_party_id):
 
     try:
         response.raise_for_status()
+        if response.status_code == 204:
+            logger.info('No cases found for business', business_party_id=business_party_id)
+            return []
```
insacuri: Is this (or any of the changes) going to work?  On success it returns response.json (a dictionary), on a 404 or 204, it returns a list.

response_operations_ui/controllers/case_controller.py
```diff
@@ -83,10 +83,10 @@ def get_case_groups_by_business_party_id(business_party_id):
 
     try:
         response.raise_for_status()
-    except requests.exceptions.HTTPError:
         if response.status_code == 204:
             logger.info('No case groups found for business', party_id=business_party_id)
             return []
```
insacuri: Is this (or any of the changes) going to work?  On success it returns response.json (a dictionary), on a 204, it returns a list.

PR #444, opened 2020-04-01
Updated pipfile

bkhan2:
# What has changed
updated pyscss version from 1.3.5 to 1.3.7

# How to test?
run `make build-kubernetes`


warrenbailey:
I don't think we should be upgrading all dependencies in this. Can we just upgrade the broken one?

Pipfile
```diff
@@ -28,6 +28,7 @@ python-dateutil = "*"
 requestsdefaulter = "*"
 flask-zipkin = "*"
 bs4 = "*"
+werkzeug = "*"
```
warrenbailey, line 31: should we try:

werkzeug = "~=0.16"

PR #458, opened 2020-04-16
updated logging messages. updated notify location in config.py

bkhan2:
# What has changed
updated logging, when notify notifications don't go out
# How to test?
- spin up ras-rm-docker-dev
- stop docker version of response-ops
- spin up local version of response-ops
- go to localhost:8085 to see local instance of response-ops

in the code:
- comment out the if block (line 27)
- `if not app.config`

In the app:
- "create" a new account
- it should say "Something went wrong"
- check logs in terminal (where you started the local version) to see if the error logging is coming through (lines 48-50)

If you could also see if the logs I've put are sufficient, if more or less information would be required. I have logged what I would find useful when looking at logs but if that may be a concern going forward, please let me know how to improve.


response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,15 +37,17 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
-        if response.status_code == 201:
+        status_number = response.status_code
+
+        if status_number == 201:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json().get("id"))
         else:
             ref = reference if reference else 'reference_unknown'
-            raise NotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify",
-                              reference=ref)
+            logger.error(f'STATUS CODE: {status_number}')
```
insacuri, line 48: There's no need to do this kind of logging.  We use structlog for structured logging.  This means that as well as a log message you can attach any number of bits of metadata to it.  For examples, look at pretty much any of the logging in the repo.
This should be something like 
```python
logger.error("There was a problem sending a notification to the notify-gateway", status_code=status_code, error_text=response.text, url=response.url)
```

response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,15 +37,17 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+            logger.error(f'URL: {url}')
+            raise NotifyError(f'FAILED TO SEND EMAIL TO: {notification["emailAddress"]}', reference=ref)
```
insacuri, line 50: The message in this isn't great and it lacks the context you'd need for a useful exception to be thrown (which is why we're fixing this in the first place).  Also never just put an error message in caps, it's not helpful.
The logging that was there before was fine, but `"There was a problem sending a notification to the notify-gateway"` is probably a little better because it's not needlessly wordy.  If you look at the NotifyError in exceptions.py, you'll notice it takes an error, a description and any other number of kwargs (key value pairs essentially, e.g., raise NotifyError("Text", response.text, key1=value1, key2=value2, etc)).

This should probably be something like 
```
raise NotifyError("There was a problem sending a notification to the notify-gateway", response.text, reference=ref)
```
I've not tested any of my suggestions btw so there's a chance that it might not be as easy as what I've mentioned, but it'll be a useful starting point

response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,14 +37,16 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
         response = requests.post(url, auth=auth, json=notification)
-        if response.status_code == 201:
+        status_code = response.status_code
+
+        if 201 <= status_code <= 399:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json().get("id"))
         else:
             ref = reference if reference else 'reference_unknown'
-            raise NotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify",
+            raise NotifyError(f'There was a problem sending a notification to Notify-Gateway. URL = {url},'
```
mfcrocker, line 48: As per @insacuri 's dismissed review, you don't need to pass these variables in as fstrings, `NotifyError` takes kwargs
```suggestion
            raise NotifyError("There was a problem sending a notification to Notify-Gateway.", url=url, status_code=status_code, message=response.text, reference=ref)
```

response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,14 +37,16 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
             notification.update({"reference": reference})
 
         url = f'{self.notify_url}{template_id}'
-
         auth = app.config['SECURITY_USER_NAME'], app.config['SECURITY_USER_PASSWORD']
         response = requests.post(url, auth=auth, json=notification)
-        if response.status_code == 201:
+        status_code = response.status_code
+
+        if 201 <= status_code <= 399:
```
mfcrocker, line 44: The only valid responses from notify-gateway are 201 and 400 [source](https://github.com/ONSdigital/rm-notify-gateway/blob/master/API.md)

```suggestion
        if status_code == 201:
```

response_operations_ui/controllers/notify_controller.py
```diff
@@ -38,17 +38,16 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
-        else:
-            ref = reference if reference else 'reference_unknown'
-            raise NotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify",
-                              reference=ref)
 
+        try:
+            response.raise_for_status()
+        except HTTPError:
+            ref = reference if reference else 'reference_unknown'
+            raise NotifyError('There was a problem sending a notification to Notify-Gateway', reference=ref)
```
mfcrocker, line 49: It would be good to have the response text in the log line (you should just be able to merge this suggestion)

```suggestion
            raise NotifyError('There was a problem sending a notification to Notify-Gateway', message=response.text, reference=ref)
```

response_operations_ui/controllers/notify_controller.py
```diff
@@ -38,17 +38,16 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+
```
mfcrocker, line 50: The Travis build is failing because this line has whitespace

```suggestion

```

response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,16 +39,19 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+            response.raise_for_status()
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json().get("id"))
-        else:
+        except HTTPError:
             ref = reference if reference else 'reference_unknown'
-            raise NotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify",
-                              reference=ref)
-
+            logger.error('There was a problem sending a notification to Notify-Gateway', 
+                         ref=ref, status_code=status_code, message=response.text)
```
insacuri, line 52: It's minor, but we only use that status_code variable once, and only if there's an error.  Might as well just do `status_code=response.status_code` and do away with assigning the variable above

response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,16 +39,19 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+            raise NotifyError()
```
insacuri, line 53: You can't and shouldn't leave this empty.  If you see what other stuff in  the code handles these errors,  you'll notice that they all have `msg=e.description` in the log lines.  You should at the very minimum change this line to 
`raise NotifyError('There was a problem sending a notification to Notify-Gateway') and tbh it should probably be 
`raise NotifyError('There was a problem sending a notification to Notify-Gateway', error=response.text) so that if other code wants to query what went wrong, then it can.

Honestly I think this needs looking at properly down the line, but what you're doing here is an improvement.

response_operations_ui/controllers/notify_controller.py
```diff
@@ -39,16 +39,16 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+            response.raise_for_status()
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json().get("id"))
-        else:
+        except HTTPError:
             ref = reference if reference else 'reference_unknown'
-            raise NotifyError("There was a problem sending a notification to Notify-Gateway to GOV.UK Notify",
-                              reference=ref)
-
+            logger.error(error=response.text, status_code=response.status_code, ref=ref)
+            raise NotifyError('There was a problem sending a notification to Notify-Gateway')
```
insacuri, line 50: I should've just said to change it to this `raise NotifyError('There was a problem sending a notification to Notify-Gateway', error=response.text)` instead of explaining my thought process >_<

If we have the useful text that we can put into the exception, then we should add it as it might be needed in the future and would be frustrating to find out that the NotifyError offers the error text but nothing fills it in!

PR #469, opened 2020-04-21
Create script for mass emailing of enrolment codes

insacuri:
# Motivation and Context
Enrolment codes needed to be sent to respondents via email.  Gov notify allows you to provide a csv file with emails + all the variables to personalise the email.  This script sorts out that csv file for you.

I've added a do not merge label on it to start as there's a few minor details to iron out, but it can be reviewed for code quality now.

# What has changed

# How to test?

# Links


scripts/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+A csv file is generated from these that will be uploaded to gov.notify(https://www.notifications.service.gov.uk/)
+to send the ras-email-enrolment-code email.  This is currently being done as a special case during the covid-19 situation
+as businesses aren't able to receive their enrolment code for surveys as they're not present at their registered office
+address.
+
+To run the script, do the following:
+```bash
+cd scripts
+pipenv run python generate_csv_for_emails.py <business_filename> <print_filename> construction XXX YYY
```
mfcrocker, line 25: I would be more specific here about what XXX and YYY are rather than expecting people to look a couple of paragraphs up. I'd also consider taking out the "construction" parameter until there's a need to do this for different surveys (only because we're currently on 6 args which is pretty brutal to use :D )

insacuri: Good spot.  That was fully my intention and I got the answer but forgot to update this ^^

PR #498, opened 2020-06-09
Delete Functionality for reminder emails

akssinhaamit:
# Motivation and Context

# What has changed

# How to test?

# Links

### DIFF
--- a/response_operations_ui/views/update_event_date.py
+++ b/response_operations_ui/views/update_event_date.py
@@ -29,6 +29,7 @@ def update_event_date(short_name, period, tag):
                      short_name=short_name, period=period)
         abort(404)
     events = collection_exercise_controllers.get_collection_exercise_events_by_id(exercise['id'])
+    show = get_reminder_del_visibility(events, None, tag)
     event_name = get_event_name(tag)
     formatted_events = convert_events_to_new_format(events)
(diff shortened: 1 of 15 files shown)


response_operations_ui/views/update_event_date.py
```diff
@@ -50,7 +51,20 @@ def update_event_date(short_name, period, tag):
                            period=period,
                            survey=survey,
                            event_name=event_name,
-                           date_restriction_text=date_restriction_text)
+                           date_restriction_text=date_restriction_text,
+                           show=show)
+
+
+def get_reminder_del_visibility(events, show, tag):
```
jterryons, line 58: Im not really clear on what this is trying to do, could you put a docstring in?

akssinhaamit: sorry for that... I will provide the docstring.
"""This function checks if the tag being viewed is a reminder. 
    If No it returns show=None.
    If yes it creates a list of existing reminders in sequence,
    if tag being viewed is the last item in the sorted existing reminders it returns show=true else false.
    Args:
        param1: existing events,
        param2: None,
        param3: tag being viewed.
    Retrns:
        None: if the viewed tag is not a reminder
        True: if the viewed tag is the last in existing sorted reminder list.
        False: if the viewed tag is not the last in existing sorted reminder list.
    """

PR #503, opened 2020-06-22
Add logging and fix sample data bug

insacuri:
# Motivation and Context
Kept getting an error 500 page when using overview page in preprod.  Sample data comes as an int.  Needed to convert it to a string before it got to the template.

Also added more logging as it was a bit too light.

# What has changed

# How to test?

# Links

### DIFF
--- a/response_operations_ui/common/filters.py
+++ b/response_operations_ui/common/filters.py
@@ -65,5 +65,7 @@ def get_current_collection_exercise(collection_exercises):  # noqa: C901
         return {}
 
     logger.info("About to return most current collection exercise",
-                collection_exercise_id=closest_collection_exercise['id'])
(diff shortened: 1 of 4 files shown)


response_operations_ui/common/filters.py
```diff
@@ -65,5 +65,7 @@ def get_current_collection_exercise(collection_exercises):  # noqa: C901
         return {}
 
     logger.info("About to return most current collection exercise",
-                collection_exercise_id=closest_collection_exercise['id'])
+                collection_exercise_id=closest_collection_exercise['id'],
+                exercise_ref=closest_collection_exercise.get('exerciseRef'),
+                user_descrption=closest_collection_exercise.get('userDescription'))
```
jdfox, line 70: tiny typo "user_descrption"

PR #510, opened 2020-07-16
Strip whitespace from email address in EditContactDetailsForm

jdfox:
# Motivation and Context

There are two separate functions implemented for editing contact details at:
- `/edit-contact-details/<respondent_id>`
- `/<ru_ref>/edit-contact-details/<respondent_id>`
The former validates, but the other doesn't. Fixing validation in forms.py. Also changing email validation so that leading and trailing spaces are silently stripped, rather than throwing an error to the user: spaces in the email still raise an error in the edit dialogue.

# What has changed
Added a custom filter to the email field of `EditContactDetailsForm` in forms.py to str.strip() entries.  Added tests for this on respondent search and RU search.
Also bumped minor version of node package lodash to fix a medium-severity vulnerability.

# How to test?
- `make test` locally
- deploy gateway with new image to a sandbox cluster
- run acceptance tests on sandbox cluster 
- set port forwarding from localhost with ras-rm-cli
- edit contact details from respondent and RU search; ensure that spaces are silently stripped from both; ensure spaces in emails still raise an error.

# Links
- https://collaborate2.ons.gov.uk/jira/browse/RAS-40
- https://trello.com/c/kfHL2etU


response_operations_ui/forms.py
```diff
@@ -75,16 +75,22 @@ class RuSearchForm(FlaskForm):
 
 
 class EditContactDetailsForm(FlaskForm):
+    def strip_whitespace(s):
+        if isinstance(s, str):
+            s = s.strip()
```
insacuri, line 80: Looks good, though I have a question.  Was this only an issue of space before and after?  Was there an issue of space in the middle?  
What happens if I provide `test email@example.com` does it fail?  I'm fairly sure it should

jdfox: That's right, if a user enters a leading or trailing space the code should silently strip the spaces and continue. But if there is a space inside the email address it should fail validation.

PR #512, opened 2020-07-21
Refactor spoke templates

boxadesign:
# Motivation and Context
- Upgrade the R'Ops to latest design system version.
- Refactor spoke templates to be compliant with the design system patterns.

# What has changed
Removed the "spoke" template aka fake modal. These pages now use the standard base template.

# How to test?
- Most of the files changed in this PR are design system components that can be ignored i.e anything `/components`
- Ensure all pages render and function correctly.


response_operations_ui/templates/update-event-date.html
```diff
@@ -1,128 +1,208 @@
...
-              {{ form.month(id="date-en-gb-month", class_="input input--select input--w-auto") }}
-            </div>
-            <div class="field field--input">
-              <label class="label u-fs-r" for="date-en-gb-year" id="label-date-range-from-year" data-qa="label-year">Year</label>
-              {{ form.year(id="date-en-gb-year", class="input input--text input--StringField input--w-4", type="number") }}
-            </div>
-          </div>
-        </fieldset>
+        {{
+            onsDateInput({
```
insacuri, line 92: Currently doesn't work right. On an edit screen it should load up the old values and in this one it doesn't do that.
Moreover, when defaults are put in place, the time values don't work as they're select values don't act right when given a string as they're SelectFields instead of StringFields.

insacuri: Actually SelectField vs Stringfield thing was a mistake I made in the template, but the point still stands that it doesn't replay current values if available here (and I imagine everywhere else this change was made)

PR #513, opened 2020-07-23
Upload bres collection instruments

insacuri:
# Motivation and Context
POC of how we might allow BRES collection instruments to be uploaded via response ops

In testing, you can upload duplicates.  We'd probably have to fix that to make this less error prone

# What has changed

# How to test?

# Links

### DIFF
--- a/response_operations_ui/controllers/collection_instrument_controllers.py
+++ b/response_operations_ui/controllers/collection_instrument_controllers.py
@@ -47,11 +47,11 @@ def upload_collection_instrument(collection_exercise_id, file, form_type=None):
     return True
 
 
-def upload_bres_collection_instrument(collection_exercise_id, file, ru_ref):
+def upload_ru_specific_collection_instrument(collection_exercise_id, file, ru_ref):
     """
-    Uploads a BRES collection instrument. This is different to a standard SEFT as we supply an ru_ref to link the
-    instrument to both a reporting unit AND a collection exercise where a standard SEFT is linked ONLY to a
-    collection exercise.
+    Uploads an reporting unit specific collection instrument. This is different to a standard SEFT as we supply an
+    ru_ref to link the instrument to both a reporting unit AND a collection exercise where a standard SEFT is linked
+    ONLY to a collection exercise.  An example of a survey that uses ru specific collection instruments is BRES.
 
     :param collection_exercise_id: The collection exercise this instrument will be linked to
     :type collection_exercise_id: str
--- a/response_operations_ui/controllers/collection_instrument_controllers.py
+++ b/response_operations_ui/controllers/collection_instrument_controllers.py
@@ -12,6 +12,15 @@ logger = wrap_logger(logging.getLogger(__name__))
 
 
 def upload_collection_instrument(collection_exercise_id, file, form_type=None):
+    """
+    Uploads a standard SEFT collection instrument.  This will be linked to the collection exercise via the id provided.
+
+    :param collection_exercise_id: The collection exercise this instrument will be linked to
+    :type collection_exercise_id: str
+    :param file: The collection instrument (spreadsheet) to be uploaded
+    :param form_type: A formtype to act as metadata for the instrument
+    :return: True on success, False on failure.
+    """
     logger.info('Uploading collection instrument', collection_exercise_id=collection_exercise_id, form_type=form_type)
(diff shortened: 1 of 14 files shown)


response_operations_ui/views/collection_exercise.py
```diff
@@ -253,7 +260,11 @@ def _upload_collection_instrument(short_name, period):
```
warrenbailey, line 260: I wonder if it would be worth having a function before this validate method is called that determines the type and the name of collection instrument.

Then you could split the _validate_collection_instrument function in two, one for bres and one for others and call the appropriate function when you know the type of CI. I feel that might be cleaner, you could also then use the results further now and avoid to do the split multiple times.

response_operations_ui/views/collection_exercise.py
```diff
@@ -307,19 +327,29 @@ def _validate_collection_instrument():
...
-                }
+            if file.filename.split(".")[0].isdigit():
+                logger.info("BRES collection instrument detected", filename=file.filename)
+                if len(file.filename.split(".")[0]) != 11:
+                    logger.info('Invalid BRES file format uploaded', filename=file.filename)
+                    error = {
+                        "section": "ciFile",
+                        "header": "Error: Invalid file name format for BRES collection instrument",
+                        "message": "Please provide a file with a valid 11 digit ru ref in the file name"
+                    }
```
warrenbailey, line 342: All these nested ifs feel a little complicated to me, I think we should split this function. See above

response_operations_ui/views/collection_exercise.py
```diff
@@ -100,12 +100,19 @@ def view_collection_exercise(short_name, period):
     info_panel = request.args.get('info_panel')
     sorted_nudge_list = get_existing_sorted_nudge_events(ce_details['events'])
 
+    # This is an ugly fix for errors being written to the permanently to the session. This guarantees that an error will
+    # be displayed once and then removed. If errors are ever tidied up (using flash for instance) then this code can go.
+    error_json = None
```
warrenbailey, line 105: nice fix but worth extracting to it's own function?

response_operations_ui/views/collection_exercise.py
```diff
@@ -307,19 +327,29 @@ def _validate_collection_instrument():
...
-            # file name format is surveyId_period_formType
-            form_type = _get_form_type(file.filename) if file.filename.count('_') == 2 else ''
-            if not form_type.isdigit() or len(form_type) != 4:
-                logger.info('Invalid file format uploaded', filename=file.filename)
-                error = {
-                    "section": "ciFile",
-                    "header": "Error: invalid file name format for collection instrument",
-                    "message": "Please provide file with correct form type in file name"
-                }
+            if file.filename.split(".")[0].isdigit():
```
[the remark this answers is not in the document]
insacuri: Yeah I like that.  I toyed around with it for ages but couldn't figure out what to call it.  `ci_name` is perfect

response_operations_ui/controllers/collection_instrument_controllers.py
```diff
@@ -38,8 +47,48 @@ def upload_collection_instrument(collection_exercise_id, file, form_type=None):
     return True
 
 
+def upload_bres_collection_instrument(collection_exercise_id, file, ru_ref):
+    """
+    Uploads a BRES collection instrument. This is different to a standard SEFT as we supply an ru_ref to link the
+    instrument to both a reporting unit AND a collection exercise where a standard SEFT is linked ONLY to a
```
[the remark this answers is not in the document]
insacuri: Agreed. Renamed to `upload_ru_specific_collection_instrument`

response_operations_ui/views/collection_exercise.py
insacuri: Done

response_operations_ui/views/collection_exercise.py
insacuri: Would you mind if I held off on those changes until the post-BRES work?  A lot of this code is going to be moved around and reorganised and it might be a good idea to get it in there instead of now

response_operations_ui/views/collection_exercise.py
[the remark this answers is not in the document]
insacuri: Done

response_operations_ui/views/collection_exercise.py
warrenbailey: np :)

PR #514, opened 2020-07-29
strip whitespace at update to party controller

jdfox:
# Motivation and Context

Fix for https://collaborate2.ons.gov.uk/jira/browse/RAS-40, where in testing whitespace was stripped from the form during contact update (https://github.com/ONSdigital/response-operations-ui/pull/510) but not from the updated email pushed to Party service.

# What has changed
Added strip() of email to update_contact_details in party controller.

# How to test?
Same as last pull request https://github.com/ONSdigital/response-operations-ui/pull/510, but this time actually check for spaces in the database with `SELECT * FROM partysvc.respondent;`.

# Links

### DIFF
--- a/_infra/helm/response-operations-ui/Chart.yaml
+++ b/_infra/helm/response-operations-ui/Chart.yaml
@@ -18,4 +18,4 @@ version: 1.1.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.1.8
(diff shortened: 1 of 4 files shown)


warrenbailey:
Might be worth a unit test?

jdfox:
> Might be worth a unit test?

Added a check for whitespace in the unit test added in https://github.com/ONSdigital/response-operations-ui/pull/510.

PR #516, opened 2020-08-03
able to create new survey

bkhan2:
# Motivation and Context
Add functionality for users to create surveys.

# What has changed
Added a line of HTML linking to a function

### DIFF
--- a/response_operations_ui/templates/surveys.html
+++ b/response_operations_ui/templates/surveys.html
@@ -18,6 +18,7 @@
   {% endif %}
 
   <h1 name="page-surveys-title">Surveys</h1>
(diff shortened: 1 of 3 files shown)


response_operations_ui/templates/surveys.html
```diff
@@ -18,6 +18,7 @@
   {% endif %}
 
   <h1 name="page-surveys-title">Surveys</h1>
+  <p><a href="{{ url_for('surveys_bp.create_survey') }}"> Create new survey </a></p>
```
insacuri, line 21: I know it's minor but I think those spaces are superfluous

PR #531, opened 2020-08-20
Seft collection instrument

akssinhaamit:
# Motivation and Context

# What has changed

# How to test?

# Links


response_operations_ui/templates/collection_exercise/ce-instrument-section.html
```diff
@@ -1,8 +1,71 @@
 <section class="ce-section {{'panel panel--simple panel--error' if missing_ci else ''}}">
+
+  {# 
+    This section needs to be updated once RAS-117 is completed.
+    Create new survey should provide an attribute, whether the survey is SEFT or EQ  
+  #}
+  {% set short_name = survey.shortName %}
```
insacuri, line 7: Feel free to disagree with me, but I think this should be done in the view code rather in the template.  I think the aim is to keep logic in the templates to a minimum where possible and it'd be easier to do something like 
```
{% if seft %}
....
```

response_operations_ui/templates/survey.html
```diff
@@ -83,16 +83,10 @@ <h1 name="page-survey-title" class="u-mb-l">{{survey.longName}}</h1>
...
-            {{
-                onsButton({
-                    "text": "Create collection exercise",
-                    "id": "create-collection-exercise",
-                    "name": "create-collection-exercise-link"
-                })
-            }}    
-    </form>
+      <p class="field u-mt-l">
+        <a href="{{ url_for('collection_exercise_bp.get_create_collection_exercise_form', survey_ref=survey.surveyRef, short_name=survey.shortName.replace(' ', '')) }}"
```
insacuri, line 87: Might need to check if links that look like buttons are allowed

insacuri (changes requested):

tests/views/test_collection_exercise.py
```diff
@@ -558,7 +567,7 @@ def test_choose_collection_instrument_when_first(self, mock_request, mock_detail
         response = self.client.get(f'/surveys/{short_name}/{period}', follow_redirects=True)
 
         self.assertEqual(response.status_code, 200)
-        self.assertIn("Add a collection instrument. Must be XLSX".encode(), response.data)
+        self.assertNotIn("Add a collection instrument. Must be XLSX".encode(), response.data)
```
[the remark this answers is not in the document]
akssinhaamit: Sure... makes sense
will change it

response_operations_ui/templates/collection_exercise/ce-instrument-section.html
akssinhaamit: I will agree with you, but this code has been done just till the time we get Bilals PR in place. We can still take this logic to the view. I will try to put this logic to view code.

insacuri (approved):

PR #541, opened 2020-09-07
Fix bold text in unread messages 

boxadesign:
# Motivation and Context
Fixes issue where unread messages were not displayed in bold.

### DIFF
--- a/response_operations_ui/templates/messages.html
+++ b/response_operations_ui/templates/messages.html
@@ -153,6 +153,35 @@
           {% else %}
             {% set ruRef = 'Unavailable' %}
           {% endif %}
+          {% if message.business_name and message.unread %}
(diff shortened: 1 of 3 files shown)


response_operations_ui/templates/messages.html
```diff
@@ -166,30 +195,30 @@ <h1 class="u-fs-l u-mt-l">Oops! Something went wrong</h1>
                   "id": 'ru-ref-' + loop.index|string
                 },
                 {
-                  "value": message.business_name if message.business_name else 'Unavailable',
+                  "value": businessName,
```
insacuri, line 198: There is a chance the business name might not exist, is there a reason why you removed this?

insacuri (approved):
Tested, looks good

PR #544, opened 2020-09-10
Use new ras rm notify

akssinhaamit:
# Motivation and Context

# What has changed

# How to test?

# Links


response_operations_ui/controllers/notify_controller.py
```diff
@@ -1,59 +1,67 @@
...
-            "emailAddress": email,
+            'notify': {
+                'email_address': email,
+                'template_id': template_id,
+                'personalisation': {}
+            }
         }
         if personalisation:
-            notification.update({"personalisation": personalisation})
-        if reference:
```
jterryons, line 38: Do we want to delete reference or is that not relevant anymore?

akssinhaamit: I don't think its relevant as I could not find the reference to the attribute being used ever. 
`_send_message` is referred from `request_to_notify` which is referred from the following functions which never passed reference as an attribute.
- `send_confirm_created_email`
- `send_create_account_email`
- `send_confirm_change_email`
- `send_password_change_email`

jterryons (approved):

PR #546, opened 2020-09-15
Survey page overview

bkhan2:
# Motivation and Context
ability to add survey mode when creating survey

# What has changed
added funcitonality

# How to test?
sign in to rops, create a survey, click on the survey. It should say eQ or SEFT in the Survey Mode section.


response_operations_ui/forms.py
```diff
@@ -199,6 +199,7 @@ class CreateSurveyDetailsForm(FlaskForm):
                                                 message='Please use alphanumeric characters only.')])
     survey_ref = StringField('survey_ref', validators=[InputRequired(message="Please remove spaces in Survey ID")])
     legal_basis = SelectField('legal_basis', choices=[('', 'Select an option')])
+    survey_mode = SelectField('survey_mode', choices=[('', 'Select an option'), ('eQ', 'eQ'), ('SEFT', 'SEFT')])
```
insacuri, line 202: is one of the pairs of values the value of the field?  If so, one of them needs to be EQ.  I think you said you'd look at it, but it's important that it gets done so I'll mention it so it doesn't get forgotten about!

insacuri (approved):

PR #556, opened 2020-10-29
Bug fix for respondent being referenced after deletion

akssinhaamit:
# Motivation and Context
The responded was being referenced from reporting unit after deletion which causes 500 error
# What has changed
logic has been added to avoid 500 errors

# How to test?
Smoke test and business approval required
# Links
[Trello](https://trello.com/c/NzvpAaU3/185-bug-respondent-delete-bugs)

### DIFF
--- a/response_operations_ui/controllers/party_controller.py
+++ b/response_operations_ui/controllers/party_controller.py
@@ -61,6 +61,8 @@ def get_respondent_by_party_id(respondent_party_id):
         response.raise_for_status()
     except requests.exceptions.HTTPError:
         log_level = logger.warning if response.status_code in (400, 404) else logger.exception
+        if response.status_code == 404:
+            return ''
         log_level('Error retrieving respondent party', respondent_party_id=respondent_party_id)
         raise ApiError(response)
 
--- a/response_operations_ui/controllers/party_controller.py
+++ b/response_operations_ui/controllers/party_controller.py
@@ -61,9 +61,9 @@ def get_respondent_by_party_id(respondent_party_id):
         response.raise_for_status()
     except requests.exceptions.HTTPError:
         log_level = logger.warning if response.status_code in (400, 404) else logger.exception
-        if response.status_code == 404:
-            return ''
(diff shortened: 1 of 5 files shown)


response_operations_ui/views/case.py
```diff
@@ -110,7 +110,8 @@ def get_user_from_case_events(case_events):
         except KeyError:
             logger.info('There was no partyId in case event', case_event=case_events[first_case_event])
             return ''
-        respondent_name = respondent.get('firstName') + ' ' + respondent.get('lastName')
+        respondent_name = respondent.get('firstName') + ' ' + respondent.get('lastName') if respondent != '' \
+            else 'Respondent does not exists in the system. This may be due to the GDPR retention policy.'
```
warrenbailey, line 114: I think if you make the above an empty json I think you could make this

if not respondent_name:
   return Respondent does not exists in the system. This may be due to the GDPR retention policy.

warrenbailey (changes requested):
Looks good, just one thing

response_operations_ui/controllers/party_controller.py
```diff
@@ -61,6 +61,8 @@ def get_respondent_by_party_id(respondent_party_id):
         response.raise_for_status()
     except requests.exceptions.HTTPError:
         log_level = logger.warning if response.status_code in (400, 404) else logger.exception
+        if response.status_code == 404:
+            return ''
```
warrenbailey, line 65: I think this would be better returning an empty json object rather than a string to keep the return value consistent

response_operations_ui/controllers/party_controller.py
```diff
@@ -61,6 +61,8 @@ def get_respondent_by_party_id(respondent_party_id):
...
         log_level('Error retrieving respondent party', respondent_party_id=respondent_party_id)
```
warrenbailey, line 66: we should probably move this log line above the if statement too

warrenbailey (approved):

PR #558, opened 2020-11-06
Banner management

TomHolroyd:
# Motivation and Context

As a rOps user, I want the ability to see the current alert banner running, and have the ability to remove it so that I may swap the alert banners which are displayed to respondents.
# What has changed

# How to test?

# Links
- https://trello.com/c/Bb75vjhE


warrenbailey (changes requested):

_infra/helm/response-operations-ui/Chart.yaml
```diff
@@ -18,4 +18,4 @@ version: 1.1.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.3.6
+appVersion: 2.3.7
```
jterryons, line 21: New feature, increment minor version.

response_operations_ui/views/admin.py
```diff
@@ -44,15 +53,48 @@ def update_banner():
...
-    else:
-        logger.debug("Banner update", user=current_username(), banner=banner)
-        admin_controller.set_banner(form.banner.data)
     return redirect(url_for("admin_bp.banner_admin"))
 
 
-def remove_banner():
-    admin_controller.remove_banner()
-    return redirect(url_for("admin_bp.banner_admin"))
+def set_suffix(today):
```
jterryons, line 95: This one is an eyebrow raiser, im surprised that datetime doesn't support date suffixes. The stackoverflow post on this is from 9 years ago but i cant cant find any recent documentation. Could you briefly comment this and reference the stack overflow page https://stackoverflow.com/questions/5891555/display-the-date-like-may-5th-using-pythons-strftime

response_operations_ui/views/admin.py
```diff
@@ -21,14 +23,21 @@
...
     form = BannerAdminForm(form=request.form)
-
-    current_banner = admin_controller.current_banner()
-    if current_banner:
-        form.banner.data = current_banner
-        logger.debug("Banner set to ", banner=current_banner)
-    return render_template('banner-admin.html', current_banner=current_banner, form=form)
+    dict_of_alerts = get_alert_list()
+    banner_removed = False
+    if 'banner_removed' in session:
```
warrenbailey, line 32: Do we need this variable? The template should be able to just check if the current_banner is set or not, or have a missed something?

Also it seems a little odd to store it in the session, it's not something I'd expect as the banner is global for all users.

warrenbailey (changes requested):

response_operations_ui/views/admin.py
TomHolroyd: Looking at the prototype, once the banner is removed the user is redirected back to /admin/banner and a notification is posted confirming banner removal. The session was added to confirm this to that user and then disappear on refresh. https://rops-prototypes.netlify.app/prototypes/alert-banner-management/version-3/04-remove-alert.html

I can change this to just be a persistent banner with either 'No banner currently active' or '{{current_banner}}'?

warrenbailey: ah ok that makes sense then 👍

warrenbailey: I just had a thought, we could use flash flash for this notification and we actually do this elsewhere in rops:

https://flask.palletsprojects.com/en/1.1.x/patterns/flashing/

For example:
https://github.com/ONSdigital/response-operations-ui/blob/a0c7f3c34b9013ed6fc84fb11b429f30f572ba63/response_operations_ui/views/reporting_units.py#L176

PR #560, opened 2020-11-17
Banner management #558, workaround PR

jdfox:
# Motivation and Context
Workaround PR for https://github.com/ONSdigital/response-operations-ui/pull/558, where GitHub Actions has hit a bug.

# What has changed

# How to test?

# Links


jdfox:
Same result as #558

PR #561, opened 2020-11-20
Manage alert templates

TomHolroyd:
As a rOps user, I want the ability to see the current alert banner running, and have the ability to remove it so that I may swap the alert banners which are displayed to respondents.

# What has changed

# How to test?

# Links


response_operations_ui/controllers/admin_controller.py
```diff
@@ -41,3 +45,25 @@ def current_banner():
...
+        r = _get_redis()
+        banner = r.get('AVAILABILITY_MESSAGE_TIME_SET')
+        logger.debug("Getting time availability message was set", banner=banner)
+        return banner
+    except redis.RedisError:
+        logger.exception("Unable to retrieve current banners")
+
+
+def get_alert_list():
+    my_dict = {}
```
jterryons, line 61: As Warren suggested before, remove the my_dict stuff

response_operations_ui/controllers/admin_controller.py
```diff
@@ -15,13 +16,16 @@ def _get_redis():
...
+def set_banner_and_time(banner, time):
     try:
         r = _get_redis()
         r.set('AVAILABILITY_MESSAGE', banner)
+        r.set('AVAILABILITY_MESSAGE_TIME_SET', time)
         logger.debug("Setting availability message", banner=banner)
     except redis.RedisError:
-        logger.exception("Unable to updated banner")
+        logger.exception("Unable to update banner and time")
+        logger.exception('Ensure time parameter is correct structure if not '
```
jterryons, line 27: collapse this into the logger exception above.

response_operations_ui/controllers/admin_controller.py
```diff
@@ -15,13 +16,16 @@ def _get_redis():
     return r
 
 
-def set_banner(banner):
+def set_banner_and_time(banner, time):
```
jterryons, line 19: I think some of Warrens comments have been missed from the previous PR

TomHolroyd: This change was made, however when we passed a default datetime.now parameter in it would only record the time it was first called and would not update when a banner was removed/updated. I don't know if this is something to do with redis?

response_operations_ui/controllers/admin_controller.py
TomHolroyd: This change investigated. A list of dicts is returned instead of a dictionary, this required extra code in banner-admin.html. It was agreed to leave it like this for readability within the html file - happy to make the change though

jterryons: 👍

jterryons (approved):

PR #575, opened 2020-12-15
Use business attributes endpoint

insacuri:
# Motivation and Context
Previously the reporting unit would get the business attributes one at a time (to fill out the name, trading as and region for the business at the time of each collection exercise).

This PR makes that call happen at once to reduce the http and database connection overheads that repeated calls give.

# What has changed

# How to test?

# Links


response_operations_ui/controllers/party_controller.py
```diff
@@ -33,6 +33,37 @@ def get_party_by_ru_ref(ru_ref):
...
+
+    :param business_party_id: A business id
+    :param collection_exercise_ids: An (optional) list of collection exercise ids
+    :type collection_exercise_ids: list
+    :return: A dict of attributes with each collection exercise id as the keys.
+    :rtype: dict
+    """
+    bound_logger = logger.bind(business_id=business_party_id, collection_exercise_ids=collection_exercise_ids)
+    bound_logger.info('Retrieving business attributes')
+    if not collection_exercise_ids:
```
insacuri, line 50: Note for later: this line and the functionality in the docstring don't match

warrenbailey (approved):
One minor comment but looks good to me otherwise

PR #580, opened 2021-01-04
Update packages to fix security issue

insacuri:
# Motivation and Context
`pipenv check` caused the build to fail because cryptography <3.3 is required.  This PR fixes that and updated the other packages

# What has changed

# How to test?

# Links


jdfox:
In JWT 2.0, `jwt.encode` now returns token as string rather than byte string. Removing `.decode()` from `self.access_token.decode()` fixes new test errors.

jdfox (approved):

PR #587, opened 2021-02-04
changed redirect after message

anwilkie:
# Motivation and Context
rOps users needed to be sent back to the reporting unit page after sending a message, because it's tedious having to navigate back after it sends you to the message thread list.

# What has changed
The `/create-messages` endpoint now redirects the user back to the previously selected responding unit page after a message is sent.

# How to test?
On response-ops, find a reporting unit, click on a survey, find a respondent, and click create message. Then, type in a message and send it. It should redirect you back to the reporting unit page.
# Links
[Trello card](https://trello.com/c/XrN8aya6)

### DIFF
--- a/tests/views/test_message.py
+++ b/tests/views/test_message.py
@@ -503,7 +503,7 @@ class TestMessage(ViewTestCase):
         mock_request.get(ru_ref_url, status_code=200)
 
         with self.app.app_context():
-            response = self.client.post("/messages/create-message", data=self.message_form, follow_redirects=True)
+            response = self.client.post("/messages/create-message", data=self.message_form)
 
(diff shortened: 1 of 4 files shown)


tests/views/test_message.py
```diff
@@ -509,6 +562,16 @@ def test_form_submit_with_FDI_data(self, mock_request, mock_get_jwt):
         mock_request.get(url_get_threads_list, json=thread_list, status_code=200)
```
insacuri, line 562: Do you still need these old mocks if you're not going to the inbox page anymore?  I don't think you do, I'd try and remove some of these mock requests to narrow it down to only what you need

insacuri (approved):

PR #595, opened 2021-02-22
Use /business instead of /party urls in party

insacuri:
# Motivation and Context
We currently use `/party-api/v1/parties/type/B/ref/<ref>` to get data on a business from party.  There is another, more sensible url that does basically the same thing `/party-api/v1/business/ref/<ref>`.  Moving to the /business url lets us retire the /parties one someday and make the party api more reasonable.

# What has changed

# How to test?

# Links


mfcrocker (approved):
Would like to see a card in the backlog to move over

PR #606, opened 2021-04-14
Moving sample distribution to not be async

mfcrocker:
# What and why?
Sample distribution is currently on a cron, but there's no real reason for it to be so and we can move it to being called when we hit the set live button

# How to test?

# Trello
https://trello.com/c/HTCVFGNU/556-replace-sample-distribution-cron-job-with-a-synchronous-call-in-collection-exercise


mfcrocker:
Closing in favour of doing this as part of the sample refactor

PR #623, opened 2021-05-21
Msgs add technical category

insacuri:
# What and why?

# How to test?

# Trello


response_operations_ui/controllers/message_controllers.py
```diff
@@ -84,34 +85,38 @@ def get_all_conversation_type_counts(survey_id, conversation_tab, business_id):
         return totals
     except KeyError:
         logger.exception("Response was successful but didn't contain a 'totals' key")
-        raise NoMessagesError
+        raise
```
creativeappdevamit, line 88: shouldn't we define what to raise just for the clarity?

creativeappdevamit (approved):
Looks Ok to me, just one question very minor happy if you want to ignore it.

response_operations_ui/controllers/message_controllers.py
insacuri: The stack trace for a key error would say it's missing a 'totals' key so finding out what's wrong should still be very clear.  It's also never once happened as far as I'm aware.

PR #624, opened 2021-05-26
Removing time from ref period start and end dates

warrenbailey:
# What and why?
Ref period start, Ref period end and employment date are date fields rather than datetime fields. This PR stops the user being able to set the time element.

# How to test?
Set up a CE and check that the time field is not displayed for the fields ref period start, ref period end and employment.
Once set up check, repeat the same check for updating the fields.

# Trello
https://trello.com/c/IRR0NUpx/666-bug-ritm0581287-this-is-now-sorted-but-need-to-query-how-the-reference-period-start-date-is-picked-up-by-eq-there-is-an-issue-wi

### DIFF
--- a/_infra/helm/response-operations-ui/Chart.yaml
+++ b/_infra/helm/response-operations-ui/Chart.yaml
@@ -18,4 +18,4 @@ version: 1.1.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.4.45
+appVersion: 2.4.46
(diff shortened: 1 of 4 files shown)


mfcrocker (approved):
I'd probably prefer to remove this data from the DB entirely but this is a good quick way of doing it

warrenbailey:
unfortunately that's not possible because they share the event table and the timestamp column with the reminders and nudge events that do require time

mfcrocker (approved):

PR #629, opened 2021-06-03
Fix table ordering

insacuri:
# What and why?
Fixes both the survey ordering in the reporting unit page (was ordering on the survey's hidden uuid rather then the displayed  reference number (009, 023, etc) and the broken resend verification url link (the ru_ref wasn't being set correctly)

# How to test?

# Trello


creativeappdevamit (approved):
looks ok

PR #630, opened 2021-06-04
Fix bug in filter

insacuri:
# What and why?

The filter in the technical inbox was breaking when used.  This PR fixes it and also fixes the clear filter button that was also broken

# How to test?

# Trello


mybigopenbox (approved):
looks good to me

PR #650, opened 2021-07-16
Iteration changes for rops to view and reply on non survey messages

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/3sJKB2KU/779-user-account-creation-something-else-rops-iteration


warrenbailey:
It appears to be missing the email on the message when you view it. Apart from that the functionality looks good

warrenbailey:
looks like the respondent is wrong too, it's shows ONS.

warrenbailey:
fixed the first issue, the second one is general issue with secure message so raised a bug for it

insacuri (approved):

PR #664, opened 2021-10-05
Flask cache for survey info

warrenbailey:
# What and why?

# How to test?

# Trello


warrenbailey:
Abandoning this approach due to:
```
The Flask-Caching extension through 1.10.1 for Flask relies on Pickle for serialization, which may lead to remote code execution or local privilege escalation. If an attacker gains access to cache storage (e.g., filesystem, Memcached, Redis, etc.), they can construct a crafted payload, poison the cache, and execute Python code.
```

PR #670, opened 2021-11-22
updated pipenv packages

anwilkie:
# What and why?
Pipenv packages needed to be updated, so a `pipenv lock` command was run after pinning `black` to the most recent stable build `21.10b0` and pinning `flask-paginate` to version `0.8.1` because updating the latter caused unit tests to fail.

# How to test?
Run acceptance tests in your environment.

# Trello
[Card](https://trello.com/c/9YxwiMlX)


creativeappdevamit (changes requested):
I think the card was to remove vulnerabilities in Response Operations and unpin and update the packages.
These changes does it take out the vulnerabilities? I think all we have done is pin another package.

anwilkie:
> 

Adam Mann told me not to worry about the `npm` packages and to only update the pipenv ones to their latest versions. I only pinned one of the packages because updating it broke unit tests.

creativeappdevamit:
> > 
> 
> Adam Mann told me not to worry about the `npm` packages and to only update the pipenv ones to their latest versions. I only pinned one of the packages because updating it broke unit tests.

Hi Adam - not sure I understand the pipenv ones. I don't this will remove vulnerability. Can you please check with Adam and ask him for the review. If he is happy, I am ok with it.

insacuri (approved):
Approving, but I'm gonna make a card to unpin flask-paginate as it really wouldn't take much work to get the tests passing with the latest one

PR #673, opened 2021-11-26
Bug 404 response dashboard

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/cWQOE5je/1061-bug-responses-dashboard-404-error

### DIFF
--- a/response_operations_ui/common/filters.py
+++ b/response_operations_ui/common/filters.py
@@ -18,7 +18,7 @@ def get_current_collection_exercise(collection_exercises):  # noqa: C901
     Figures out what the most 'current' collection exercise is from a list of them.
     This is done with the following 4 steps:
       - If there are no collection exercises, return an empty dict
-      - Search for the nearest to today past scheduledStartDateTime.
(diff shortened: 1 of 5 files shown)


response_operations_ui/common/filters.py
```diff
@@ -18,7 +18,7 @@ def get_current_collection_exercise(collection_exercises):  # noqa: C901
     Figures out what the most 'current' collection exercise is from a list of them.
     This is done with the following 4 steps:
       - If there are no collection exercises, return an empty dict
-      - Search for the nearest to today past scheduledStartDateTime.
+      - Search for the nearest to today past scheduledStartDateTime in LIVE, READY FOR LIVE and ENDED state.
```
insacuri, line 21: Looks good,  but can you add a comment after this list.   Something like:
```
The reason we care about the state in the exercises in the past, but not in the future, is because if there are exercises in the past then they're likely to be set up correctly and give us useful information.  If there are none with a start date the past then it's likely that this is a new survey, and we just want whatever information we can get.

Additionally, the responses dashboard doesn't handle not fully set up collection exercises too well, so we can try and save the user from giving them a dashboard link that won't work.
```

insacuri (approved):

PR #679, opened 2021-12-07
updation WT form in response operations

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/3YI5D4K5/1187-upgrade-wtforms-to-300-in-response-operations-ui


insacuri (approved):
Bit of a faff we have to flash errors but it is what it is!  I've not tested it myself but it looks fine from a code standpoint

PR #680, opened 2021-12-07
Fix issue when EQ CEs don't have an EQ version provided

mfcrocker:
# What and why?
We were getting issues when an EQ version wasn't provided by the survey service when creating a collection exercise. I've chosen to default this to v2 rather than v3 as this feels more likely to be compatible

# How to test?
Find a survey in the survey schema that doesn't have an eqVersion set (or unset it), then create a collection exercise for that survey

# Trello
https://trello.com/c/IMh0i4O7/1188-new-eq-survey-breaks-when-creating-first-collection-exercise

### DIFF
--- a/response_operations_ui/views/collection_exercise.py
+++ b/response_operations_ui/views/collection_exercise.py
@@ -568,7 +568,9 @@ def get_create_collection_exercise_form(survey_ref, short_name):
     logger.info("Retrieving survey data for form", short_name=short_name, survey_ref=survey_ref)
     form = CreateCollectionExerciseDetailsForm(form=request.form)
     survey_details = survey_controllers.get_survey(short_name)
-    survey_eq_version = survey_details["eqVersion"] if survey_details["surveyMode"] != "SEFT" else ""
+    survey_eq_version = "v2"
+    if survey_details["surveyMode"] == "EQ" and "eqVersion" in survey_details:
+        survey_eq_version = survey_details["eqVersion"]
     return render_template(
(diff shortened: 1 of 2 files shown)


insacuri:
This kinda papers over the problem rather than actually fixing it.  If that's the intent, as a quick fix until a card is made to do a proper fix then I'll approve it.

Otherwise, the fix really should be setting a default eq version when the survey is initially created; Which would involve probably a new radio button on the survey screen (and hopefully only that?  I think the POST survey endpoint allows you to set the eq_version, otherwise that's a huge oversight!)

response_operations_ui/views/collection_exercise.py
```diff
@@ -568,7 +568,9 @@ def get_create_collection_exercise_form(survey_ref, short_name):
     logger.info("Retrieving survey data for form", short_name=short_name, survey_ref=survey_ref)
     form = CreateCollectionExerciseDetailsForm(form=request.form)
     survey_details = survey_controllers.get_survey(short_name)
-    survey_eq_version = survey_details["eqVersion"] if survey_details["surveyMode"] != "SEFT" else ""
+    survey_eq_version = "v2"
```
insacuri, line 571: Can you put a comment that this is here in the unlikely event the eqVersion field is blank in the survey?

creativeappdevamit (changes requested):
I don't think this is the fix, the UI should show radio button not selected if its "" and force user to select one.

PR #682, opened 2021-12-09
updating verification email templates

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/yk676X1g/1225-change-rops-resend-email-to-use-the-new-template


response_operations_ui/views/reporting_units.py
```diff
@@ -325,6 +326,7 @@ def search_reporting_units():
     )
 
 
+# Not sure why is this endpoint exposed : Amit Sinha
```
insacuri, line 329: Are you not sure why it's exposed because nothing uses it or something else? Do you want to make a quick investigation card so we can look into it?

insacuri (approved):

PR #685, opened 2021-12-15
Bug Fix for resend email

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/yk676X1g/1225-change-rops-resend-email-to-use-the-new-template

### DIFF
--- a/response_operations_ui/views/respondents.py
+++ b/response_operations_ui/views/respondents.py
@@ -302,24 +302,32 @@ def edit_contact_details(respondent_id):
     )
 
 
-@respondent_bp.route("/resend_verification/<respondent_id>", methods=["GET"])
+@respondent_bp.route("/resend-verification/<respondent_id>", methods=["GET"])
 @login_required
 def view_resend_verification(respondent_id):
     logger.info("Re-send verification email requested", respondent_id=respondent_id)
     respondent = party_controller.get_respondent_by_party_id(respondent_id)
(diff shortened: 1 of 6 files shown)


response_operations_ui/views/respondents.py
```diff
@@ -302,24 +302,32 @@ def edit_contact_details(respondent_id):
...
 
 
-@respondent_bp.route("/resend_verification/<respondent_id>", methods=["GET"])
+@respondent_bp.route("/resend-verification/<respondent_id>", methods=["GET"])
 @login_required
 def view_resend_verification(respondent_id):
     logger.info("Re-send verification email requested", respondent_id=respondent_id)
     respondent = party_controller.get_respondent_by_party_id(respondent_id)
-    email = respondent["pendingEmailAddress"] if "pendingEmailAddress" in respondent else respondent["emailAddress"]
+    is_new_email_verification_request = True if "pendingEmailAddress" in respondent else False
```
insacuri, line 310: This might not work as you're expecting it.  `<key> in <dict>` will return True if the key is present, even if the value is false.  Or am I misunderstanding it, and a pendingEmailAddress key will only be there if present, and it's absent if there isn't one?

creativeappdevamit: I believe pendingEmailAddress key will only be present if there is one. I just tested it locally and it seems to work.

insacuri (approved):

PR #690, opened 2022-01-28
make period dates start/end mandatory

mybigopenbox:
# What and why?
fields need to be mandatory
# How to test?
try to create /upload eq
# Trello
https://trello.com/c/iSJM76j1/1301-bug-mandatory-events-needs-to-be-filled-in-before-a-collection-exercise-can-go-live

### DIFF
--- a/response_operations_ui/views/collection_exercise.py
+++ b/response_operations_ui/views/collection_exercise.py
@@ -105,11 +105,16 @@ def view_collection_exercise(short_name, period):
     ]
 
     ce_state = ce_details["collection_exercise"]["state"]
-    show_set_live_button = (
-        ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION")
-        and ce_details["events"].get("ref_end_period_start")
-        and ce_details["events"].get("ref_end_period_end")
-    )
+    if 'surveyMode' in ce_details['survey'] and ce_details['survey']['surveyMode'] =='EQ':
+            show_set_live_button = (
+                ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION")
+                and "ref_period_start" in ce_details["events"]
+                and "ref_period_end" in ce_details["events"]
+            )
+    else:
+        show_set_live_button = (
+                ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION"))
+        
     locked = ce_state in ("LIVE", "READY_FOR_LIVE", "EXECUTION_STARTED", "VALIDATED", "EXECUTED", "ENDED")
     processing = ce_state in ("EXECUTION_STARTED", "EXECUTED", "VALIDATED")
     validation_failed = ce_state == "FAILEDVALIDATION"
--- a/response_operations_ui/views/collection_exercise.py
+++ b/response_operations_ui/views/collection_exercise.py
@@ -105,7 +105,9 @@ def view_collection_exercise(short_name, period):
     ]
 
     ce_state = ce_details["collection_exercise"]["state"]
-    show_set_live_button = ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION")
+    show_set_live_button = ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION") \
+                           and ce_details["events"].get("ref_end_period_start") \
+                           and ce_details["events"].get("ref_end_period_end")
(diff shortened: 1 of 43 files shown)


response_operations_ui/views/collection_exercise.py
```diff
@@ -105,7 +105,15 @@ def view_collection_exercise(short_name, period):
     ]
 
     ce_state = ce_details["collection_exercise"]["state"]
-    show_set_live_button = ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION")
+    if "surveyMode" in ce_details["survey"] and ce_details["survey"]["surveyMode"] == "EQ":
```
insacuri, line 108: This can be replaced with:
```python
if ce_details["survey"].get("surveyMode") == "EQ"
```

`.get(key)` returns None if the key doesn't exist so you can write it with only 1 condition instead of having the `and`

mybigopenbox: cheers.
It was the .get() that caused the failure as (null & null) returned the wrong boolean

response_operations_ui/views/collection_exercise.py
```diff
@@ -105,7 +105,15 @@ def view_collection_exercise(short_name, period):
     ]
 
     ce_state = ce_details["collection_exercise"]["state"]
-    show_set_live_button = ce_state in ("READY_FOR_REVIEW", "FAILEDVALIDATION")
+    if ce_details["survey"]["surveyMode"] == "EQ":
```
creativeappdevamit, line 108: can you check why FAILEDVALIDATION is a part of this logic?

creativeappdevamit (commented):
what is the test? It looks like you are checking everywhere that the GOLIVE is not present. Where is the test for present?

creativeappdevamit:
Why is the PR still in progress when trello card has been moved to review? Can you please make sure to take out the label

response_operations_ui/views/collection_exercise.py
insacuri: It's been like that forever.  I think the theory of it is that it would fail validation, you'd fix whatever was wrong behind the scenes then you'd try again.

Whether that ever happened in practice is another story, but I don't think it's worth removing it here just to limit the amount of change in this PR.  We're already touching something quite important (the ability to set an exercise live) so I don't want to do too much

tests/views/test_collection_exercise.py
```diff
@@ -227,6 +271,113 @@ def setup_data(self):
...
+        mock_request.get(url_get_by_survey_with_ref_start_date, json=collection_exercise_eq_ref_start_date)
+        mock_request.get(url_get_by_survey_with_ref_end_date, json=collection_exercise_eq_ref_end_date)
+
+        response = self.client.get(f"/surveys/{short_name}/{period}", follow_redirects=True)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("Business Register and Employment Survey".encode(), response.data)
+        self.assertIn("221_201712".encode(), response.data)
+        self.assertNotIn("Select eQ version".encode(), response.data)
+        self.assertNotIn("Set as ready for live".encode(), response.data)
```
insacuri, line 379: If it's an EQ AND it has both dates then the ready for live button should be appearing?

tests/views/test_collection_exercise.py
```diff
@@ -227,6 +271,113 @@ def setup_data(self):
...
+
+        mock_request.get(url_get_by_survey_with_ref_end_date, json=collection_exercise_eq_ref_end_date)
+
+        response = self.client.get(f"/surveys/{short_name}/{period}", follow_redirects=True)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("Business Register and Employment Survey".encode(), response.data)
+        self.assertIn("221_201712".encode(), response.data)
+        self.assertNotIn("Select eQ version".encode(), response.data)
+        self.assertIn("Set as ready for live".encode(), response.data)
```
insacuri, line 351: If theres only the end date and not the start date then it shouldn't be appearing

insacuri (changes requested):

response_operations_ui/views/collection_exercise.py
mybigopenbox: So, Do I need to do something with FAILEDVALIDATION ?

creativeappdevamit (approved):

insacuri (approved):

PR #691, opened 2022-02-01
Add my account page

insacuri:
# What and why?

This is the basis for the new 'my account' page in response ops.  This includes a page that displays the data held on an internal user in uaa.  It also moves the sign out link to the top header instead of the middle header.

Note:  This PR doesn't include the link to the page and aborts as soon as you try and use it.  This will be removed and the page 'switched on' in a following story once we have the ability to change any of those details.

# How to test?

# Trello

### DIFF
--- a/response_operations_ui/controllers/uaa_controller.py
+++ b/response_operations_ui/controllers/uaa_controller.py
@@ -94,6 +94,31 @@ def get_user_by_email(email, access_token=None):
     return response.json()
 
 
+def get_user_by_id(user_id, access_token=None):
+    if access_token is None:
+        access_token = login_admin()
+
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+        "Authorization": f"Bearer {access_token}",
+    }
+
+    url = f"{app.config['UAA_SERVICE_URL']}/Users/{user_id}"
+    response = requests.get(url, headers=headers)
+    try:
+        response.raise_for_status()
+    except HTTPError:
+        logger.error(
+            "Error retrieving user from UAA",
+            status_code=response.status_code,
+            user_id=user_id
+        )
+        return
+
+    return response.json()
+
+
 def retrieve_user_code(access_token, username):
     headers = {
         "Content-Type": "application/json",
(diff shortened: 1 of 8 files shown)


response_operations_ui/controllers/uaa_controller.py
```diff
@@ -94,6 +94,27 @@ def get_user_by_email(email, access_token=None):
     return response.json()
 
 
+def get_user_by_id(user_id, access_token=None):
+    if access_token is None:
+        access_token = login_admin()
```
creativeappdevamit, line 99: I am not sure I get this can you please let me know what does this do?

response_operations_ui/templates/layouts/base.html
```diff
@@ -8,27 +8,49 @@
...
-    <!-- End Google Tag Manager (noscript) -->	
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ config.GOOGLE_TAG_MANAGER }}&gtm_auth={{ config.GOOGLE_TAG_MANAGER_PROP }}&gtm_cookies_win=x"
+    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+    <!-- End Google Tag Manager (noscript) -->
    {% endif %}
 {% endblock %}
-
-{% set pageConfig = { 
+{% if current_user.is_authenticated %}
```
creativeappdevamit, line 30: do we have test for this section? If not is it worth adding it?

creativeappdevamit (changes requested):

response_operations_ui/controllers/uaa_controller.py
insacuri: I've tweaked this, but all that login_admin function does is get the token that allows response ops (as the client) do stuff in uaa.  It's not perfect, but it's lifted from the other stuff that seems to work.

response_operations_ui/templates/layouts/base.html
insacuri: I added a test for this by tacking on a few extra checks in the login view tests.  Testing for 'my account' and 'sign out' when logged in, and the opposite when not authenticated

creativeappdevamit (approved):

PR #692, opened 2022-02-01
Change name

JacobHarrisonONS:
# What and why?
Enable users of rOps to change their name on their account

# How to test?

# Trello
https://trello.com/c/oeMOw3ax/1260-rops-ui-self-service-change-name


JacobHarrisonONS:
/deploy harrisonj configBranch=add-account-name-change-notify-template

JacobHarrisonONS:
/deploy harrisonj configBranch=add-account-name-change-notify-template

JacobHarrisonONS:
/deploy harrisonj configBranch=add-account-name-change-notify-template

JacobHarrisonONS:
/deploy harrisonj configBranch=add-account-name-change-notify-template

arroyoAle:
/deploy arroya configBranch=add-account-name-change-notify-template

arroyoAle:
/deploy arroya configBranch=add-account-name-change-notify-template

Pipfile
```diff
@@ -53,6 +53,7 @@ Jinja2 = "*"
 MarkupSafe = "*"
 WTForms = "*"
 email_validator = "*"
+self = "*"
```
insacuri, line 56: what is this for?

_infra/helm/response-operations-ui/templates/deployment.yaml
```diff
@@ -135,6 +135,8 @@ spec:
             value: {{ .Values.notify.requestCreateAccountTemplate }}
           - name: NOTIFY_REQUEST_PASSWORD_CHANGE_TEMPLATE
             value: {{ .Values.notify.requestPasswordChangeTemplate }}
+          - name: NOTIFY_CHANGE_ACCOUNT_DETAILS_TEMPLATE
```
insacuri, line 138: Kinda picky, but the value should more closely match the key name.  So this should be:
`NOTIFY_UPDATE_ACCOUNT_DETAILS_TEMPLATE`

I won't comment against everything that uses this value ,but everything that uses this value needs to be updated as well!

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -200,3 +200,33 @@ def create_user_account(email, password, user_name, first_name, last_name):
             )
 
     return errors
+
+
+def update_user_account(payload):
```
insacuri, line 205: Can you put a docstring here describing what the function does and what the payload should be?  With context, it's clear the payload is whatever you got from `GET /users/<id>` then tweaked.  But someone who has never touched uaa wouldn't know that.

response_operations_ui/forms.py
```diff
@@ -495,3 +512,15 @@ class BannerEditForm(FlaskForm):
...
     banner_id = StringField("id")
     delete = BooleanField("Delete banner", default=False)
+
+
+class Form(FlaskForm):
+    def validate(self):
+        return "option" in self.data and self.data["option"]
+
+
+class OptionsForm(Form):
```
insacuri, line 522: This has a generic name, but is really a form for a specific page.  Something like 'MyAccountForm' or something like that would be better I think.

response_operations_ui/forms.py
```diff
@@ -495,3 +512,15 @@ class BannerEditForm(FlaskForm):
     )
     banner_id = StringField("id")
     delete = BooleanField("Delete banner", default=False)
+
+
+class Form(FlaskForm):
```
insacuri, line 517: Why does this exist?  Can't you have the OptionsForm inherit from FlaskForm, then add a validate function in there?

insacuri (changes requested):

response_operations_ui/views/accounts.py
```diff
@@ -40,6 +49,58 @@ def get_my_account():
         return jsonify(e)
 
 
+@account_bp.route("/my-account", methods=["POST"])
```
insacuri, line 52: I think this could be merged into the GET my-account above.  Instead of having to do a redirect on a fail, it'd be quicker, and look less weird if you were looking at the network traffic, if you put the POST code at the top of that function, and if it fails, you just flash an error then just do all the other stuff you'd normally do in the GET.

response_operations_ui/views/accounts.py
```diff
@@ -40,6 +49,58 @@ def get_my_account():
...
+def change_account_name():
+    try:
+        form = ChangeAccountName()
+        user_id = session["user_id"]
+        user_from_uaa = uaa_controller.get_user_by_id(user_id)
+        user = {
+            "first_name": f"{user_from_uaa['name']['givenName']}",
+            "last_name": f"{user_from_uaa['name']['familyName']}",
+        }
+        payload = user_from_uaa
```
insacuri, line 73: No need for this line.  Just use `user_from_uaa` in the lines below this, or call it something else that you'd be happy using everywhere

response_operations_ui/views/accounts.py
```diff
@@ -40,6 +49,58 @@ def get_my_account():
...
+                        )
+                        return logout.logout("Your name has been changed")
+                    except NotifyError as e:
+                        logger.error(
+                            "Error sending change of name acknowledgement email to Notify Gateway", msg=e.description
+                        )
+            else:
+                return redirect(url_for("account_bp.get_my_account"))
+        return render_template("account/change-account-name.html", user=user, form=form, errors=form.errors)
+    except Exception as e:
```
insacuri, line 100: Don't do this in a view.  You'd go from a styled html page to rendering a blob of json which would be suuuuper jarring for whoever is using the site.

Remove the try/except from this.  If an exception is thrown that isn't catched, it'll bubble up and eventually get caught by errorhandlers (that are found in view/errors.py) and depending on the type of error it'll be handled in a certain way.

tests/views/test_accounts.py
```diff
@@ -14,13 +14,21 @@
...
+with open(f"{project_root}/test_data/uaa/updated_name.json") as json_data:
+    updated_name_json = json.load(json_data)
 
 test_email = "fake@ons.gov.uk"
 user_id = "fe2dc842-b3b3-4647-8317-858dab82ab94"
 url_uaa_token = f"{TestingConfig.UAA_SERVICE_URL}/oauth/token"
 url_uaa_get_accounts = f"{TestingConfig.UAA_SERVICE_URL}/Users?filter=email+eq+%22{test_email}%22"
 url_uaa_get_user_by_id = f"{TestingConfig.UAA_SERVICE_URL}/Users/{user_id}"
 url_uaa_create_account = f"{TestingConfig.UAA_SERVICE_URL}/Users"
+url_uaa_update_name = f"{TestingConfig.UAA_SERVICE_URL}/Users/{user_id}"
```
insacuri, line 26: instead of adding another one, remove this and make the one above have a more generic name like `url_uaa_user_by_id` then have everything use that instead

response_operations_ui/views/accounts.py
```diff
@@ -40,6 +49,58 @@ def get_my_account():
...
+                        "value_name": "name",
+                        "changed_value": full_name,
+                    }
+                    try:
+                        NotifyController().request_to_notify(
+                            email=user_from_uaa["emails"][0]["value"],
+                            template_name="change_account_details",
+                            personalisation=personalisation,
+                        )
+                        return logout.logout("Your name has been changed")
```
insacuri, line 92: For this, I think it'd be better to copy what the logout does then to use the logout view as a function.
it's tricky, I could be wrong about this so let me know if you think I am.  
The way I see it, while you CAN use other views as functions, I think it should be avoided so there aren't any gotchas in the future if you were to change how a view work and accidentally break another page.

If we needed that 'logout with custom flashed message' functionality for a 3rd page, I'd argue we should factor that out into it's own function and have all 3 call it.

response_operations_ui/views/accounts.py
```diff
@@ -11,33 +11,82 @@
...
-            "email": user_from_uaa["emails"][0]["value"],
-            "password_last_changed": formatted_date,
-        }
-        return render_template("account/my-account.html", user=user)
-    except Exception as e:
-        return jsonify(e)
+    form = MyAccountOptionsForm()
+    form_valid = form.validate()
+    user_id = session["user_id"]
+    user_from_uaa = uaa_controller.get_user_by_id(user_id)
```
insacuri, line 37: This still isn't quite right.  You want the order to be:
- get form
- if POST and valid form
  - do stuff...
- do GET stuff

response_operations_ui/views/accounts.py
```diff
@@ -11,33 +11,83 @@
...
+                    NotifyController().request_to_notify(
+                        email=user_from_uaa["emails"][0]["value"],
+                        template_name="update_account_details",
+                        personalisation=personalisation,
+                    )
+                    return redirect(url_for("logout_bp.logout", message="Your name has been changed"))
+                except NotifyError as e:
+                    logger.error(
+                        "Error sending change of name acknowledgement email to Notify Gateway", msg=e.description
+                    )
```
insacuri, line 87: You'll want to actually do something if theres a notifyError that isn't just logging something out.  I don't think it has to be much as the update has already happened in uaa.  Maybe `flash("Something went wrong when sending an email to confirm the change.  Your details have been updated in the system")

insacuri (approved):

PR #693, opened 2022-02-01
Change username

arroyoAle:
# What and why?

# How to test?

# Trello


arroyoAle:
/deploy arroya configBranch=add-account-name-change-notify-template

response_operations_ui/templates/account/change-username.html
```diff
@@ -0,0 +1,93 @@
...
+{% if errors %}
+  {% set errorTitle = 'There is 1 error on this page' %}
+    {% call
+            onsPanel({
+                "type": "error",
+                "classes": "u-mb-s",
+                "title":  errorTitle
+            })
+      %}
+<p xmlns="http://www.w3.org/1999/html">These <strong>must be corrected</strong> to continue.</p>
```
insacuri, line 22: `xmlns="http://www.w3.org/1999/html` wtf is this >_<

response_operations_ui/templates/account/change-username.html
```diff
@@ -0,0 +1,93 @@
...
+                onsButton({
+                "text": "Save",
+                "id": "btn-option-save",
+                "classes": 'btn-group__btn',
+                "submitType": "timer"
+                })
+            }}
+        </div>
+        <div class="grid__col u-ml-m btn-group__btn secondary">
+          <a href="{{ url_for('account_bp.get_my_account') }}" role="button" class="btn btn--link btn--secondary" id="btn-who-ons-option-cancel">
```
insacuri, line 87: that id for this link was obviously copied from somewhere else ;)  can you change it

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -227,6 +227,9 @@ def update_user_account(payload):
         if response.status_code == 404:
             # User id not found
             errors = {"user_id": ["User id not found"]}
+        elif response.status_code == 400:
+            # Username already exists
+            errors = {"response": [400, "Username already in use. Please try another"]}
```
insacuri, line 232: For this, why can't we just use the format that the other ones use? It could look something like:
```python
errors = {"status_code": response.status_code, "message": "Username already in use. Please try another"}
```

response_operations_ui/views/accounts.py
```diff
@@ -94,6 +95,53 @@ def change_account_name():
...
+            try:
+                NotifyController().request_to_notify(
+                    email=user_from_uaa["emails"][0]["value"],
+                    template_name="update_account_details",
+                    personalisation=personalisation,
+                )
+                uaa_errors = uaa_controller.update_user_account(user_from_uaa)
+                if uaa_errors is None:
+                    return redirect(url_for("logout_bp.logout", message="Your username has been changed"))
+                elif "response" in uaa_errors.keys() and uaa_errors["response"][0] == 400:
```
insacuri, line 124: This might need to be tweaked if you go with my suggestion for the change in format for the uaa error.

response_operations_ui/views/accounts.py
```diff
@@ -94,6 +95,53 @@ def change_account_name():
...
+@account_bp.route("/change-username", methods=["GET", "POST"])
+@login_required
+def change_username():
+    form = UsernameChangeForm()
+    user_id = session["user_id"]
+    user_from_uaa = uaa_controller.get_user_by_id(user_id)
+    username = user_from_uaa["userName"]
+    username_exists = False
+    if request.method == "POST" and form.validate():
+        user_from_uaa["userName"] = form["username"].data
```
insacuri, line 107: put this inside the if statement below.  No point in doing this assignment until you've checked that they're not the same

tests/views/test_accounts.py
```diff
@@ -14,16 +14,15 @@
...
-    updated_name_json = json.load(json_data)
 
 test_email = "fake@ons.gov.uk"
 user_id = "fe2dc842-b3b3-4647-8317-858dab82ab94"
 url_uaa_token = f"{TestingConfig.UAA_SERVICE_URL}/oauth/token"
 url_uaa_get_accounts = f"{TestingConfig.UAA_SERVICE_URL}/Users?filter=email+eq+%22{test_email}%22"
 url_uaa_user_by_id = f"{TestingConfig.UAA_SERVICE_URL}/Users/{user_id}"
 url_uaa_create_account = f"{TestingConfig.UAA_SERVICE_URL}/Users"
-max_char_name = (
+url_uaa_update_account = f"{TestingConfig.UAA_SERVICE_URL}/Users/{user_id}"
```
insacuri, line 24: You can just use `url_uaa_user_by_id` instead of creating a new one.

tests/views/test_accounts.py
```diff
@@ -14,16 +14,15 @@
...
+max_char = (
```
insacuri, line 25: How many characters is this?  It might be worth calling this something like `256_character_string` so it's clear at a glance what it is

insacuri (changes requested):
Looks good, just a few bits that could be tweaked

tests/views/test_accounts.py
JacobHarrisonONS: Missed this change, we did fix in other PR

insacuri (approved):
Not tested it locally but the code looks good!

PR #701, opened 2022-02-22
Check user id change email

arroyoAle:
# What and why?
Making sure that only the original user who requested the email change can use the change email link
# How to test?

# Trello

### DIFF
--- a/response_operations_ui/views/accounts.py
+++ b/response_operations_ui/views/accounts.py
@@ -321,7 +321,8 @@ def send_update_account_email(token_dict, first_name):
         )
     else:
         logger.info(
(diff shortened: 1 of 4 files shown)


response_operations_ui/views/accounts.py
```diff
@@ -115,19 +116,24 @@ def change_email():
                 "changed_value": form.data["email_address"],
             }
             try:
+                logger.info("Sending verification email")
```
insacuri, line 119: is there any metadata you can add to this log line?  Like a user_id or something that isn't the email in plain text?

insacuri (changes requested):
A few small changes but it looks good

PR #704, opened 2022-02-25
Adding ready for live seft test

arroyoAle:
# What and why?
adding a unit test to check when a seft survey is 'ready to be set to live'
# How to test?

# Trello
https://trello.com/c/wQcoM2dA/1357-unit-testing-on-seft-collection-exercises-in-response-ops

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -121,19 +121,19 @@
         },
         "charset-normalizer": {
             "hashes": [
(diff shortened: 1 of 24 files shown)


creativeappdevamit (commented):
should we add a negative test to show `Ready for Live` doesn't appear if the collection exercise state is not ready for live?

PR #707, opened 2022-03-14
Improve process to update DS version

rmccar:
# What and why?
This PR updates, improves and documents the process to update the Design System macros in response-operations-ui

- a `.design-system-version` file has been created where the Design System version can now be set
- the Design System version has been parameterised in the `load-templates` script and that script has been renamed to `load-design-system-templates` to better explain what it does
- the Design System templates have been added to the `.gitignore` file and been removed from the ras-frontstage repo
- the `load-design-system-templates` script has been added to run when the `start` script runs

Also upgrades to v33 of the DS to fix headMeta errors in tests

# How to test?
- remove the Design System macros from your local copy of response-operations-ui (delete the `response-operations-ui/templates/components` and `response-operations-ui/templates/layout` folders)
- run `make start` and see that the Design System macros are loaded and the app runs correctly
- Docs make sense

# Trello


rmccar:
[Improve and document process to update Design System version in Rops](https://trello.com/c/BhIEQ8wI/1502-improve-and-document-process-to-update-design-system-version-in-rops)

.design-system-version
```diff
@@ -0,0 +1 @@
+33.1.4
```
mfcrocker, line 1: I'm not quite sure how well this would work with respect to the Github workflow, but this would be much better stored in the helm Chart/Values.yaml and probably referenced in the deployment as an environment variable. This is where we tend to store this sort of thing.

mfcrocker (changes requested):
Generally looks good but I think we should probably be avoiding having a specific file for a config item

mfcrocker (approved):
I think this is practically the best way of doing this after a discussion - getting stuff out of YAML for local spin-ups is something we've moved away from in recent years and I don't really want to force folks to install `yq` for a simple `make start`

PR #716, opened 2022-03-25
Adding event status on survey screen

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/rjpW5h7r/1415-rops-event-status-survey-screen


mfcrocker (approved):
LGTM, though I would ideally like to see some acceptance tests to accompany this piece - I don't think it's super easy for us to get a consistent state for a collection exercise, but if you can think of a way then please add some :)

PR #719, opened 2022-03-29
role based access respondent changes

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/TUogZ7sY/1402-role-based-access-setup-user-groupings-respondent-tab

### DIFF
--- a/response_operations_ui/templates/respondent.html
+++ b/response_operations_ui/templates/respondent.html
@@ -3,8 +3,12 @@
 {% set page_title = "Respondents" %}
 
 {% block main %}
-
+    
     {% include 'partials/flashed-messages.html' %}
+
+    
+    {% set is_respondent_edit_permission = hasPermission('respondents.edit') %}
+    {% set is_respondent_delete_permission = hasPermission('respondents.delete') %}
 
     <section>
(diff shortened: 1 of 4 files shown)


response_operations_ui/templates/respondent.html
```diff
@@ -3,18 +3,22 @@
 {% set page_title = "Respondents" %}
 
 {% block main %}
-
+    
     {% include 'partials/flashed-messages.html' %}
 
+    
+    {% set is_respondent_edit_permission = hasPermission('respondents.edit') %}
```
mfcrocker, line 10: This is still really inefficient. If we're going this way, we should get the permissions once and return a dict, and then just do 

`{% set permissions = getPermissions() %}`
`{% set is_respondent_edit_permission = 'respondents.edit' in permissions %}`

mfcrocker (changes requested):
This doesn't reflect the changes to the permissions management we discussed - even if we're not moving to caching them, we should absolutely only get them once per template

response_operations_ui/templates/respondent.html
creativeappdevamit: Until I discuss this with Dan, let's keep it this way and as a part of the caching/ changes to has_permission() we will modify this.

mfcrocker (approved):

PR #731, opened 2022-05-19
Update Design system to 49.1.0

LJBabbage:
# What and why?

# How to test?

# Trello


response_operations_ui/templates/layouts/base.html
```diff
@@ -39,8 +39,13 @@
...
             }
-        ]
-    } %}
+        ],
+        "toggleServicesButton": {
+            "text": "Menu",
+            "ariaLabel": "Toggle services menu"
+        }
+    } 
+%}
```
arroyoAle, line 48: Do we need both of these toggles? Only one show up on the pages
` "serviceLinks": serviceLinks,
    "toggleButton": {
        "text": 'Menu',
        "ariaLabel": 'Toggle main menu'
    }` (lines 58-62)

LJBabbage: I hadn't seen that one, I would think not, but will check, the DS added a requirement for toggleServicesButton as I think the other one is optional

LJBabbage: Both have a place and have been added, one for service links and one for navigation. As there is no intention to use it in mobile/small screen mode in rops the fact that they both would say menu has deemed unimportant

PR #733, opened 2022-06-09
Fixing collection exercise page to deal with sample load error

arroyoAle:
# What and why?
At the moment, if the sample is in the process of being loaded, the `ingestDateTime` value is `None` which causes a 500 error since one of the tables is trying to subscript a NoneType. This adds a check to ensure that the sample upload date time is only being accessed once the sample is fully loaded.
# How to test?
Go into the database and change the `state_fk` column in the `sample_summary` table between `INIT`, `ACTIVE`, `FAILED` and `COMPLETE`, to check that none of the states cause an error.
# Trello
https://trello.com/c/i1CkfY6p/1739-500-on-rops-when-sample-is-being-loaded-causing-acceptance-tests-to-fail

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -55,7 +55,7 @@ WTForms = "*"
 email_validator = "*"
 
 [dev-packages]
-black = "==22.3.0"
+black = "*"
 flake8 = "*"
 fakeredis = "*"
 freezegun = "*"
--- a/Pipfile
+++ b/Pipfile
@@ -55,7 +55,7 @@ WTForms = "*"
 email_validator = "*"
 
 [dev-packages]
-black = "==21.12b0"
+black = "==22.3.0"
 flake8 = "*"
 fakeredis = "*"
 freezegun = "*"
--- a/response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
+++ b/response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
@@ -76,11 +76,11 @@
         {% if 'INIT' == sample.state %}
               <div class="ons-u-mb-m">
                   <div class="ons-u-fs-r">
-                      Loading&hellip; <span class="ons-u-ml-xs"><a href="/surveys/{{survey.shortName}}/{{ce.exerciseRef}}/view-sample-ci?show_msg=true">Refresh to see progress</a></span>
+                      Loading&hellip; <span class="ons-u-ml-xs"><a href="/surveys/{{survey.shortName}}/{{ce.exerciseRef}}?show_msg=true">Refresh to see progress</a></span>
                   </div>
               </div>
         {% elif 'FAILED' == sample.state %}
-             <div class="ons-u-mb-xl">
+             <div class="ons-u-mb-m">
                   {% call
                       onsPanel({
                           "type": "error",
--- a/response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
+++ b/response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
@@ -1,74 +1,49 @@
 <section class="ce-section ce-details">
     {% set surveyEditPermission = hasPermission('surveys.edit') %}
+    {% set tblHeaders =  
+        [ 
+            { 
+                "value": "Sample & Collection instruments (CI)",
+                "thClasses": "ons-u-fs-r--b"
+            }
+        ]
+    %}
+    {% if sample and (sample.state == 'ACTIVE' or sample.state == 'COMPLETE') %}
+        {% do tblHeaders.append(
+            {
+                "value": "",
+                "thClasses": "ons-u-ta"
+            }
+         )%}
+        {% do tblHeaders.append(
+            {
+                "value": "Uploaded " + '<em>' + sample.ingestDateTime[:-8] + '</em>' + sample.ingestDateTime[-8:],
+                "thClasses": "ons-u-fs-r",
+                "numeric": "true"
+            }
+         )%}
+    {% endif %}
+    {% set surveyTableData = {
+        "variants": ['compact', 'responsive'],
+        "id": "sample-table",
+        "caption": 'Sample summary',
+        "hideCaption": true,
+        "ths": tblHeaders,
+    } %}
     {% if sample %}
-        {% if 'ACTIVE' == sample.state %}
+        {% if 'INIT' == sample.state %}
             {{
-                onsTable({
-                    "variants": ['compact', 'responsive'],
-                    "id": "sample-table",
-                    "caption": 'Loaded sample summary',
-                    "hideCaption": true,
-                    "ths": [
-                        { 
-                            "value": "Sample & Collection instruments (CI)",
-                            "thClasses": "ons-u-fs-r--b"
-                        },
-                        {
-                            "value": "",
-                            "thClasses": "ons-u-ta"
-                        },
-                        {
-                            "value": "Uploaded " + '<em>' + sample.ingestDateTime[:-8] + '</em>' + sample.ingestDateTime[-8:],
-                            "thClasses": "ons-u-fs-r",
-                            "numeric": "true"
-                        },
-                    ],
-                    
-                    "trs": [
-                        {
-                            "tds": [
-                                {
-                                    "value": "Total businesses",
-                                    "tdClasses": "ons-u-ta"
-                                },
-                                {
-                                    "value": sample.totalSampleUnits,
-                                    "name": "total-instruments",
-                                    "tdClasses": "ons-u-ta-right"
-                                },
-                                {
-                                    "value": ' ',
-                                    "name": "upload-sample-time",
-                                }
-                            ]
-                        },
-                        {
-                            "tds": [
-                                {
-                                    "value": "Collection instruments",
-                                    "tdClasses": "ons-u-ta"
-                                },
-                                {
-                                    "value": sample.expectedCollectionInstruments,
-                                    "name": "total-instruments",
-                                    "tdClasses": "ons-u-ta-right"
-                                },
-                                {
-                                    "value": ' ',
-                                    "name": "upload-sample-time",
-                                }
-                            ]
-                        }
-                    ]
-                })
+                onsTable(surveyTableData)
             }}
-        {% elif 'INIT' == sample.state %}
             <div class="ons-u-mb-xl">
                 <div class="ons-u-fs-r">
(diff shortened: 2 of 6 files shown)


response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
```diff
@@ -1,66 +1,88 @@
...
+                        { 
+                            "value": "Sample & Collection instruments (CI)",
+                            "thClasses": "ons-u-fs-r--b"
+                        },
+                        {
+                            "value": "",
+                            "thClasses": "ons-u-ta"
+                        },
+                        {
+                            "value": "Uploaded " + '<em>' + sample.ingestDateTime[:-8] + '</em>' + sample.ingestDateTime[-8:],
```
LJBabbage, line 21: I am guessing this is the problem, a None sample.ingestDateTime? Or is it something else

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
```diff
@@ -1,66 +1,88 @@
...
+                {% call
+                    onsPanel({
+                        "type": "error",
+                        "classes": "ons-u-mb-l"
+                    })
+                %}
+                    {{ sample.notes }}
+                {% endcall %}
+                {% if not locked and surveyEditPermission %}
+                    <a class="ons-u-ml-xs" id="remove_sample_btn_2" href="{{ url_for('collection_exercise_bp.get_confirm_remove_sample', short_name=survey.shortName, period=ce.exerciseRef) }}">Remove</a>
```
LJBabbage, line 82: I think we need to be more explicit then just remove the sample page we say "Replace sample file" (although that is wrong anyway as it doesn't replace anything just takes you to a delete confirmation)

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
```diff
@@ -1,66 +1,88 @@
...
+            </div>
+        {% elif 'FAILED' == sample.state %}
+            <div class="ons-u-mb-xl">
+                {% call
+                    onsPanel({
+                        "type": "error",
+                        "classes": "ons-u-mb-l"
+                    })
+                %}
+                    {{ sample.notes }}
```
LJBabbage, line 79: I didn't go to the extent of finding the text for sample.notes, is it consumable for a respondent.

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
```diff
@@ -1,66 +1,88 @@
...
+                    ]
+                })
+            }}
+        {% elif 'INIT' == sample.state %}
+            <div class="ons-u-mb-xl">
+                <div class="ons-u-fs-r">
+                    Loading&hellip; <span class="ons-u-ml-xs"><a href="/surveys/{{survey.shortName}}/{{ce.exerciseRef}}/view-sample-ci?show_msg=true">Refresh to see progress</a></span>
+                </div>
+            </div>
+        {% elif 'FAILED' == sample.state %}
```
LJBabbage, line 71: How many sample states do we have? Should one not be default, i.e else?

LJBabbage (changes requested):
It does stop the 500, but I think we need to check this is how we want it to look, if it be business, designer, product manager. The lack of any tests, even thought that seems normal, does make me hesitant, but I think that perhaps is one for ways of working

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
arroyoAle: Yeah this was the problem, it didn't have the `sample.state == "ACTIVE"` check so it was trying to subscript a None since the sample.ingestDateTime was None at the time (still in INIT phase)

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
arroyoAle: I'll just remove this on both pages as it doesn't make sense to have it since the first page button won't take you to the remove sample page and the second page already has a button that does that

Pipfile
```diff
@@ -55,7 +55,7 @@ WTForms = "*"
 email_validator = "*"
 
 [dev-packages]
-black = "==21.12b0"
+black = "*"
```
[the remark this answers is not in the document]
arroyoAle: The issue was that another package that was updated had a dependency that broke unless black was at 22.3.0 which is the latest version

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
arroyoAle: There are 4, it seems that if the state is 'COMPLETE' nothing is being rendered. For now I've set the else condition to if the state is 'ACTIVE' or 'COMPLETE', I will check with Gareth what each state should look like and add that

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
arroyoAle: It shows the user where and what the error is:
`Line 2753: This sample unit ref 49907527428 is duplicated in the file.`
`Line 1: Error in [11110000001, H, 75110.0, 75110.0, 84110, 84110, 3333, 77777, 1199000001, 5, E, WW, 08/06/2001, ONS Ent Name 1, ONS Ent Name 2, ONS Ent Name 3, ONS RU Name 1, ONS RU Name 2, ONS RU Name 3, , , , C, , 1, 102, S] due to field(s) [frosic92, rusic92]`

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
arroyoAle: The database does constrain it to the 4 states but just in case the else at the moment will show just the header I will check with Gareth what we want it to say

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
```diff
@@ -1,80 +1,97 @@
 <section class="ce-section ce-details">
     {% set surveyEditPermission = hasPermission('surveys.edit') %}
+    {% set tblHeaders =
```
LJBabbage, line 3: We have a bit of repetition of if statements, we should try to reduce them to the minimum, something like 
```  
   {% set tblHeaders =  
        [ 
            { 
                "value": "Sample & Collection instruments (CI)",
                "thClasses": "ons-u-fs-r--b"
            }
        ]
    %}
  
    {% set surveyTableData = {
      "variants": ['compact', 'responsive'],
      "id": "sample-table",
      "caption": 'Sample summary',
      "hideCaption": true,
      "ths": tblHeaders,
    } %}
  
    {% if sample %}
      {% if sample.ingestDateTime %}
        {% do tblHeaders.append(
            {
                "value": "",
                "thClasses": "ons-u-ta"
            }
         )%}
        {% do tblHeaders.append(
            {
                "value": "Uploaded " + '<em>' + sample.ingestDateTime[:-8] + '</em>' + sample.ingestDateTime[-8:],
                "thClasses": "ons-u-fs-r",
                "numeric": "true"
            }
         )%}
        {% set surveyTableDataRows = [
                {
                    "tds": [
                        {
                            "value": "Total businesses",
                            "tdClasses": "ons-u-ta"
                        },
                        {
                            "value": sample.totalSampleUnits,
                            "name": "total-instruments",
                            "tdClasses": "ons-u-ta-right"
                        },
                        {
                            "value": ' ',
                            "name": "upload-sample-time",
                        }
                    ]
                },
                {
                    "tds": [
                        {
                            "value": "Collection instruments",
                            "tdClasses": "ons-u-ta"
                        },
                        {
                            "value": sample.expectedCollectionInstruments,
                            "name": "total-instruments",
                            "tdClasses": "ons-u-ta-right"
                        },
                        {
                            "value": ' ',
                            "name": "upload-sample-time",
                        }
                    ]
                }
            ] %}
            
        {% do surveyTableData | setAttribute("trs", surveyTableDataRows) %}
      {% endif %}
  
      {{ onsTable(surveyTableData) }}
  
      {% if 'INIT' == sample.state %}
            <div class="ons-u-mb-m">
                <div class="ons-u-fs-r">
                    Loading&hellip; <span class="ons-u-ml-xs"><a href="/surveys/{{survey.shortName}}/{{ce.exerciseRef}}/view-sample-ci?show_msg=true">Refresh to see progress</a></span>
                </div>
            </div>
      {% elif 'FAILED' == sample.state %}
           <div class="ons-u-mb-xl">
                {% call
                    onsPanel({
                        "type": "error",
                        "classes": "ons-u-mb-l"
                    })
                %}
                    {{ sample.notes }}
                {% endcall %}
            </div>
      {% endif %}
    {% else %}
      {{ onsTable(surveyTableData) }}
    {% endif %}

Comment (LJBabbage) [reply response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html:3]: I did take a bit of liberty with ingestDateTime, which can be changed back to live/init

File: response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html [hunk 7]
```diff
@@ -58,24 +67,32 @@
...
+                ] %}
+            {% do surveyTableData | setAttribute("trs", surveyTableDataRows) %}
+        {% endif %}
+    
+        {{ onsTable(surveyTableData) }}
+    
+        {% if 'INIT' == sample.state %}
+              <div class="ons-u-mb-m">
+                  <div class="ons-u-fs-r">
+                      Loading&hellip; <span class="ons-u-ml-xs"><a href="/surveys/{{survey.shortName}}/{{ce.exerciseRef}}/view-sample-ci?show_msg=true">Refresh to see progress</a></span>
```
LJBabbage, line 79: It's not really refreshing, more taking you to view-sample-ci

response_operations_ui/templates/collection_exercise/ce-sample-ci-section.html
```diff
@@ -58,24 +67,32 @@
...
+                  </div>
+              </div>
+        {% elif 'FAILED' == sample.state %}
```
LJBabbage, line 82: This does produce an extra line between the error banner and the button, which makes it look a bit odd

LJBabbage (approved):
Works as expected, a few minor things, but will not block as I have dragged my heals a touch with this review

PR #734, opened 2022-06-10
Role based access messages

anwilkie:
# What and why?
Role-based access needed to be implemented for the secure message-related pages. The functionality to reply to messages, as well as re-open/close conversations is now limited to users who have the `message.edit` permission.

Additionally, pipenv dependencies have been updated to fix build issues, and `black` has been bumped to version `22.3.0` in order to fix an issue that was occurring with the `click` dependency. You can read more about this issue [here](https://github.com/psf/black/issues/2964).

# How to test?
* Deploy to your environment.
* Run acceptance tests.
* Go to R-ops and click on the reporting units tab, find an RU and click on a survey for that RU. If role-based access is enabled (you can toggle this by editing the deployment config), you should see the button to send messages. If not, the button shouldn't appear.
* Go to the secure messages tab and find a conversation. You should see the links/buttons to assign a survey, reply to a message, and close conversation when role-based access is enabled. Likewise, you shouldn't see these when it's disabled.
* Click the button to close the conversation. On the next page, you should see the button to confirm closing the conversation when RBA is enabled, and it shouldn't be there when RBA is disabled.
* After closing the conversation, find the closed conversation. On that page, you should see the button to re-open the conversation when RBA is enabled, and you shouldn't see it when RBA is disbled.
* As a bonus, run the command `kubectl get secret uaa-secret -o json --namespace=<YOUR_NAMESPACE> | jq -r '.data."uaa.yml"' | base64 -d` and you should see the `messages.delete` group listed among the other groups.

# Trello
[Card](https://trello.com/c/ZicKcJj8)


response_operations_ui/templates/reporting-unit-survey.html
```diff
@@ -134,9 +135,6 @@ <h3 id="RESPONDENTS" class="ons-u-fs-m">Respondents</h3>
                             },
                             {
                                 "value": "Enrolment status"
-                            },
```
arroyoAle, line 137: If you have `messages.edit` permission, the `Create message` button still appears but there isn't a table header, and in the design it seems that this column is still present it only removes the button when the user does not have permissions

response_operations_ui/templates/close-conversation.html
```diff
@@ -4,6 +4,7 @@
 {% from "components/fieldset/_macro.njk" import onsFieldset %}
```
arroyoAle, line 4: Should a user without `messages.edit` be able to access this page at all?

response_operations_ui/templates/conversation-view/cv-reopen-form.html
```diff
@@ -1,4 +1,5 @@
 {% from "components/input/_macro.njk" import onsInput %}
```
arroyoAle, line 1: Is someone with the right endpoint and information able to reopen a conversation without the `messages.edit` permission?

arroyoAle (changes requested):
Also, if a user has the right endpoint, are they able to create a message even without the `messages.edit` permissions?

response_operations_ui/templates/reporting-unit-survey.html
arroyoAle: Screenshot 2022-06-13 at 12 00 51

response_operations_ui/templates/close-conversation.html
anwilkie: No, but I felt that having a check here was worth it just in case a user somehow reaches this page without the right permissions.

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -293,6 +293,7 @@ def user_has_permission(permission, user_id=None) -> bool:
             "respondents.edit",
             "respondents.delete",
             "messages.edit",
+            "messages.delete",
```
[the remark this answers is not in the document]
anwilkie: The card said that the group needed to be created, so even though this PR doesn't use it, we'll eventually need it to be here.

tests/views/test_message.py
```diff
@@ -792,6 +806,7 @@ def test_conversation(self, mock_request, has_permission, mock_get_jwt):
     @requests_mock.mock()
     @patch("response_operations_ui.controllers.message_controllers._get_jwt")
     def test_conversation_reply_fail(self, mock_request, mock_get_jwt):
+        sign_in_with_permission(self, mock_request, user_permission_messages_edit_json)
         mock_get_jwt.return_value = "blah"
         mock_request.get(url_get_thread, json=thread_json)
         mock_request.get(url_get_surveys_list, json=survey_list)
```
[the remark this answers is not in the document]
anwilkie: Done.

response_operations_ui/templates/reporting-unit-survey.html
arroyoAle: I'd just check with Gareth or someone to make sure how they want it

response_operations_ui/templates/close-conversation.html
arroyoAle: Can we have it throw the 500 error for now until the new design for the errors come through?

arroyoAle: The same with the other links that a user without permission should not be able to access, just have an abort(500) if they don't have the permissions

arroyoAle (approved):

arroyoAle (approved):

PR #736, opened 2022-06-21
Improve edit user roles

insacuri:
# What and why?

# How to test?

# Trello


insacuri:
Replaced with another PR as the commit history here is messy and I can't remember how to make it tidy off the top of my head ^^

PR #740, opened 2022-06-28
Fix visual issues with edit user permissions

insacuri:
# What and why?

# How to test?

# Trello


LJBabbage (approved):
As this will go to design review I have only briefly checked the UI changes, the review in dev is more about it working, which it still does. 
The only thing I did notice and not for this story, is that if you select a letter from the alphabetical options and it doesn't have a result, you lose the ability to select another letter, only the create new user is available which seems wrong

PR #741, opened 2022-06-29
Add create user page

insacuri:
# What and why?

This PR adds the create user with permissions page

# How to test?

- Set up environment and deploy this PR into it
- Run acceptance tests to seed data
- log in with the uaa_users_admin user or give the uaa_user permissions via the database
- test the create account functionality in the 'manage user accounts' tab

# Trello

https://trello.com/c/GDuekVRm


response_operations_ui/templates/admin/user-create.html
```diff
@@ -0,0 +1,311 @@
...
+                <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+                <h1 class="ons-u-mt-l">Create your account</h1>
+
+                    {{
+                        onsInput({
+                            "id": "first_name",
+                            "name": "first_name",
+                            "type": "text",
+                            "label": {
+                                "description": "Enter the new users first name",
```
LJBabbage, line 91: Not convinced you need these descriptions

response_operations_ui/templates/admin/user-create.html
```diff
@@ -0,0 +1,311 @@
...
+{% from "components/button/_macro.njk" import onsButton %}
+{% from "components/lists/_macro.njk" import onsList %}
+{% from "components/password/_macro.njk" import onsPassword %}
+{% from "components/input/_macro.njk" import onsInput %}
+{% from "components/fieldset/_macro.njk" import onsFieldset %}
+{% from "components/table/_macro.njk" import onsTable %}
+{% from "components/checkboxes/_checkbox-macro.njk" import onsCheckbox %}
+
+{% set page_title = 'Create New Account' %}
+
```
LJBabbage, line 12: Screenshot 2022-07-11 at 13 17 43

Seems like there is a bigger gap then normal at the top and it's a bit squashed at the bottom

response_operations_ui/templates/admin/user-create.html
```diff
@@ -0,0 +1,311 @@
...
+                    {% for message in messages %}
+                        <p id="flashed-message-{{ loop.index }}">{{ message }}</p>
+                    {% endfor %}
+                {% endcall %}
+            {% endif %}
+        {% endwith %}
+          {% if form.errors %}
+            {% set errorData = [] %}
+                {% set error_count = form.errors | length %}
+                {% set plural_suffix = 's' if error_count > 1 else '' %}
```
LJBabbage, line 34: Screenshot 2022-07-11 at 13 22 04
1 error?

LJBabbage: Probably can shift this to another story as I reckon this is what we have in frontstage, but I am not sure the value of 3 errors which all link to the same place.

response_operations_ui/templates/account/verify-account.html
```diff
@@ -0,0 +1,132 @@
...
+                        },
+                        "attributes": {
+                            "autocomplete": 'off'
+                        },
+                        "showPasswordText": "Show password",
+                        "error": passwordError
+                    })
+                }}
+                {{
+                    onsInput({
```
LJBabbage, line 103: I am thinking this should also have show password?

response_operations_ui/views/admin/manage_user_accounts.py
```diff
@@ -80,6 +103,85 @@ def manage_user_accounts():
...
+    token = token_decoder.generate_token(user_id)
+    internal_url = current_app.config["RESPONSE_OPERATIONS_UI_URL"]
+    verification_url = f"{internal_url}{url_for('account_bp.get_verify_account', token=token)}"
+    if not current_app.config["SEND_EMAIL_TO_GOV_NOTIFY"]:
+        logger.info("Verification url for new user", verification_link=verification_url)
+
+    try:
+        NotifyController().request_to_notify(
+            email=email,
+            template_name="create_user_account",
```
LJBabbage, line 165: Might want to check this as I messed with the templates on _get_template_id to simulate someone deleting it from notify, it didn't give me an error

tests/controllers/test_uaa_controller.py
```diff
@@ -40,6 +55,8 @@ def setUp(self):
         self.access_token = jwt.encode(payload, self.app.config["UAA_PRIVATE_KEY"], algorithm="RS256")
         self.client = self.app.test_client()
 
+    # user_has_permission
```
LJBabbage, line 58: This looks strange, not sure why linting hasn't picked this up?

tests/controllers/test_uaa_controller.py
```diff
@@ -140,3 +167,96 @@ def test_get_groups_failure(self, mock_request):
...
+        mock_request.get(url_uaa_user_by_email, status_code=404)
+
+        with self.app.test_request_context():
+            self.assertIsNone(uaa_controller.change_user_password("some.one@ons.gov.uk", user_password))
+
+    @requests_mock.mock()
+    def test_change_user_password_retrieve_code_failure(self, mock_request):
+        mock_request.post(url_uaa_token, json={"access_token": self.access_token}, status_code=201)
+        mock_request.get(url_uaa_user_by_email, json=uaa_user_search_email, status_code=200)
+        # Docs aren't clear about failure codes for this endpoint so 500 is a safe one we can assume it can throw.
```
LJBabbage, line 207: We need to migrate away from 500, but I realise this is everywhere

LJBabbage (changes requested):
I haven't gone through the tests at present (really wished we had pytest and fixtures) but made some initial observations. It does seem to work, but could do with some tweaks. I do have a concern that the verification email is active for 6 weeks even after I used it. I used it once  logged in then came back and used it again successfully changing the password, this potentially is a security risk

tests/controllers/test_uaa_controller.py
insacuri: This is just the error code from uaa I'm talking about.  The API isn't clear what error status codes it can throw so I had to guess a 500.  But I do agree that anything front facing shouldn't be 500!

response_operations_ui/controllers/notify_controller.py
```diff
@@ -14,8 +14,12 @@ class NotifyController:
     def __init__(self):
         self.request_password_change_template = app.config["NOTIFY_REQUEST_PASSWORD_CHANGE_TEMPLATE"]
         self.confirm_password_change_template = app.config["NOTIFY_CONFIRM_PASSWORD_CHANGE_TEMPLATE"]
+        self.create_user_account_template = app.config["NOTIFY_CREATE_USER_ACCOUNT_TEMPLATE"]
+
+        # request and confirm create account templates are the pre-RBA ones and will be removed at some point.
```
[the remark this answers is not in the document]
insacuri: Updated the comment to be clearer.  The point was for it to be a marker for something that's going to be deleted once this is live.

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -146,7 +147,17 @@ def retrieve_user_code(access_token, username):
     return response.json().get("code")
 
 
-def change_password(access_token, user_code, new_password):
+def change_password(access_token: str, user_code: str, new_password: str) -> requests.Response:
+    """
+    Resets the password for a user using a user_code that was given to us by uaa via the password reset functionality
```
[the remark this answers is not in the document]
insacuri: Yeah that's fair, changed it.

tests/controllers/test_uaa_controller.py
insacuri: I think because it's valid as you can have a comment above a function with an extra line.  But you're right, it does look weird so I'll move it closer together

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -179,9 +190,70 @@ def change_user_password(email, password):
...
+    password_reset_code = retrieve_user_code(access_token=access_token, username=username)
+    if password_reset_code is None:
+        return
+
+    return change_password(access_token=access_token, user_code=password_reset_code, new_password=password)
+
+
+def create_user_account_with_random_password(email: str, first_name: str, last_name: str) -> dict:
+    """
+    Creates a user in uaa with a 64 character length password.  This is designed to be used when a user with admin
```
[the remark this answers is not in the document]
insacuri: Yeah agreed, removed.

response_operations_ui/templates/admin/user-create.html
insacuri: It was part of the design 🤷‍♂️

response_operations_ui/common/token_decoder.py
```diff
@@ -22,11 +22,11 @@ def decode_access_token(access_token):
     return decoded_jwt
 
 
-def generate_email_token(email):
-    """Creates a token based on a provided email address
+def generate_token(data):
+    """Creates a token based on a provided data
```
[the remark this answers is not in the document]
insacuri: Sure, fixed.

response_operations_ui/views/accounts.py
```diff
@@ -437,6 +437,38 @@ def post_create_account(token):
...
+
+
+@account_bp.route("/verify-account/<token>", methods=["POST"])
+def post_verify_account(token):
+    duration = app.config["CREATE_ACCOUNT_EMAIL_TOKEN_EXPIRY"]
+    user_id = token_decoder.decode_email_token(token, duration)
+    user = uaa_controller.get_user_by_id(user_id)
+    if user is None:
+        raise Exception("User does not exist")
+    form = VerifyAccountForm(request.form)
```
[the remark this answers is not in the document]
insacuri: Agreed, though would you be okay if I put it as a point in the old account creation code deletion card?  It'd be easier to know what's needed and what isn't once we stop changing everything

response_operations_ui/templates/admin/user-create.html
insacuri: Created a card to tackle this specifically as it should probably be fixed but doesn't need to hold up this PR getting merged.

response_operations_ui/templates/account/verify-account.html
insacuri: Agreed, fixed.

response_operations_ui/templates/admin/user-create.html
LJBabbage: No worries, just seemed a bit overkill

response_operations_ui/templates/admin/user-create.html
LJBabbage: It does seem like there are multiple issues here, the strange error count, the error message duplication and the linking. If your passwords matched, you only have one password with doesn't meet the requirements

response_operations_ui/views/admin/manage_user_accounts.py
insacuri: We talked about this and it's fine.  If the value is bad then we can't do anything about it as the failure will happen in the cloud function that sends the request to gov notify.
If the key is missing, it'll raise a keyError that'll bubble all the way up and cause a 500 error.  It's a bit messy but this should never ever happen so I'm okay with letting it break messy.

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -179,9 +190,70 @@ def change_user_password(email, password):
     return change_password(access_token=access_token, user_code=password_reset_code, new_password=password)
 
 
-def create_user_account(email, password, user_name, first_name, last_name):
+def reset_user_password_by_id(user_id: str, password: str) -> requests.Response | None:
```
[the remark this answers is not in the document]
insacuri: I changed the names so they're more consistent but I'm really happy with that suggested code change as I don't think it's much better then what we have here.  Would you be okay with just the name change or do you feel strongly that we should change this bit?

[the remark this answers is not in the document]
LJBabbage: It's only an opinion and this does works, so I have no issue if you don't want to change it. Absorbing retrieve_user_code (password_reset_code) into change password does make sense to me, because you always need to do it to change a password and if we had 5+ places that called change password would we want to call and pass it 5 times. Up to you, but won't block

response_operations_ui/templates/admin/user-create.html
insacuri: All fixed

LJBabbage: Padding at the bottom looks fine 👍 , but mine still seems to have a bigger gap at the top then norma
Screenshot 2022-07-14 at 09 09 38
l

LJBabbage (approved):
Functionally fine, visuals deferred to content designer

PR #744, opened 2022-07-01
Remove EQ_VERSION_ENABLED flag

insacuri:
# What and why?

When we removed the feature flag for the eq v3 work, we removed the flag from the code, but forgot to remove it from the helm chart.  This PR removes the remaining config for it.

# How to test?

# Trello


LJBabbage (approved):
I can only see it being used in these 2 places

PR #745, opened 2022-07-04
Refactor get_users_list to only have one return type

mfcrocker:
# What and why?
Having a method return >1 type is confusing and a pretty quick way to get errors

# How to test?

# Trello
https://trello.com/c/po2v9sgc/1788-update-getuserslist-in-responseops-to-not-have-multiple-return-types

### DIFF
--- a/response_operations_ui/controllers/uaa_controller.py
+++ b/response_operations_ui/controllers/uaa_controller.py
@@ -414,7 +414,7 @@ def user_has_permission(permission, user_id=None) -> bool:
 
 def get_users_list(
     start_index: int, max_count: int, query: str = None, sort_by: str = "email", sort_order: str = "ascending"
-) -> dict | str:
+) -> dict:
     """
     Gets all users in uaa.  A query can be provided to refine the search.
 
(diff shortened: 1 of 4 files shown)


LJBabbage (commented):
Not this PR but the code needs a look at. I am not sure how we get an "Unauthorised Access"  with step access_token = login_admin() just above it. To me unless I have read it wrong, this would prevent an authorised? Meaning the error would be inaccurate to both us and the user. Calling to _get_refine_user_list and instantiation Pagination on error doesn't make sense either. 

On the PR the doc string return on get_users_list could do with a tweak and I think we might consider changing the error message to someone different for the user as even if it does somehow produce an unauthorised it isn't exclusive to that. I think we should have a message which gets them to try again, rather than them thinking they need to request more rights.

response_operations_ui/controllers/uaa_controller.py
```diff
@@ -432,13 +432,12 @@ def get_users_list(
...
     url = f"{app.config['UAA_SERVICE_URL']}/Users"
     response = requests.get(url, params=param, headers=headers)
-    # TODO Update this function to only have one return type (dict).  Multiple return types isn't great.
     try:
         response.raise_for_status()
         return response.json()
     except HTTPError:
         logger.error("Unauthorised attempt to get user list.", status_code=response.status_code)
-        raise "Unauthorised Access"
+        raise {"error": "Unauthorised Access", "totalResults": 0, "resources": []}
```
insacuri, line 440: This is raising a dict.  I didn't know you could and I'm surprised this is working >_<  This should probably be replaced with return?

insacuri (approved):

LJBabbage (approved):
Nice one 👍 You don't need to return "totalResults": 0, "resources": [] in the error anymore, but it doesn't need me to review again.

PR #749, opened 2022-07-08
Improve test performance and layout

insacuri:
# What and why?

PR attached shaves about 40-50 seconds off the whole test suite.

The view tests are (time) expensive to run as each one seems to have a slow startup.  By combining some of the error tests and using the mocks between sets of them, we get the same coverage but the suite runs.

As well as this, we tidied up some of the tests and removed a few that were providing 0 value.

# How to test?

# Trello


LJBabbage (approved):
Does seem to make things a bit quicker and looks ok. 👍 I do think we need to consider pytest as this kind of file would be hugely benefit from it. I will put I card up for discussion

PR #750, opened 2022-07-14
Bug fix missing alphabetical selection

SteveScorfield:
# What and why?
The alphabetical selection menu was missing when a letter was selected and there were no users with this letter. This needed to be fixed so a user didn't have to keep clicking back to the previous selection. I have also added a section for the no user message so that there is even spacing when no users are found.

# How to test?
1. Log in as an admin user
2. Click the 'Manage user accounts' button
3. Click on a letter until the 'No results found' message is returned
4. The alphabetical menu should be present

# Trello
https://trello.com/c/j7eVOjwg/1804-manage-user-accounts-does-not-display-alphabetical-selection-if-result-is-blank

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM python:3.10-slim
 WORKDIR /app
 RUN rm -rf node_modules
 COPY . /app
-EXPOSE 8085
+EXPOSE 8086
 RUN apt-get update -y && apt-get install -y python3-pip unzip curl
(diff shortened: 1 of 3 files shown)


JacobHarrisonONS (approved):
Tested and works as it should. the table change is a little hard to read. Steve confirmed his changes were moving {% if user_list is defined and user_list|length > 0 %} to line 107 and added the section around the output on else.

PR #751, opened 2022-07-18
bump design system

JacobHarrisonONS:
# What and why?
Bump design system

# How to test?

# Trello
https://trello.com/c/AP5L9Kap/1812-upgrade-design-system-to-most-recent-version


arroyoAle (commented):
The only thing I see is that the top banner where the service links are has changed colour, not sure which one its supposed to be but its now dark blue and on main it's white (it's also white on frontstage but it's already like that on the main branch

arroyoAle (approved):

PR #754, opened 2022-07-22
Fix double 'not started' radio bug on response status page

insacuri:
# What and why?

When going to view the response status of a case, if the case was in the NOTSTARTED state then there would be 2 visually identical 'Not started' radio buttons in addition to the 2 that set it into a completed state.  This PR fixes the issue and provides extra clarity in the form of a comment around the code to hopefully prevent it happening again.

For context, it happened because we recently added 2 new valid transitions from the NOTSTARTED state.  These states looped back to NOTSTARTED, but because they resulted in a NOTSTARTED end state, they were deemed 'allowed' and were put onto the screen.  This is because the responses page blindly takes the list of possible transitions and regardless of what state the case is currently in, if any possible transition ends up in the NOTSTARTED, COMPLETEDBYPHONE or NOLONGERREQUIRED states then it'll put it in the list of radio buttons for this page.

If we ever decide to offer UI pages that do more than:
- NOT STARTED -> COMPLETED BY PHONE
- NOT STARTED -> NO LONGER REQUIRED
then we'd want to be a bit smarter about checking the current state and creating the list of radio buttons based on that, instead of the basic way we're doing it now.

# How to test?

- Run acceptance tests
- Log in and go to the reporting units tab, pick any reporting unit, then click on the survey
- On the one case for the survey, click view.  The response chasing page should only have 2 radio buttons.

# Trello

### DIFF
--- a/response_operations_ui/views/case.py
+++ b/response_operations_ui/views/case.py
@@ -25,7 +25,7 @@ from response_operations_ui.controllers.party_controller import (
 )
 from response_operations_ui.forms import ChangeGroupStatusForm
 
-COMPLETE_STATE = ["COMPLETE"]
+COMPLETE_STATE = ["COMPLETE", "COMPLETEDBYPHONE"]
 COMPLETED_CASE_EVENTS = ["OFFLINE_RESPONSE_PROCESSED", "SUCCESSFUL_RESPONSE_UPLOAD", "COMPLETED_BY_PHONE"]
 SUCCESSFUL_CASE_EVENTS = ["OFFLINE_RESPONSE_PROCESSED", "SUCCESSFUL_RESPONSE_UPLOAD", "ONLINE_QUESTIONNAIRE_RESPONSE"]
 case_bp = Blueprint("case_bp", __name__, static_folder="static", template_folder="templates")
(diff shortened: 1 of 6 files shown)


tests/views/test_change_response_status.py
```diff
@@ -68,6 +68,9 @@ def setup_data(self):
             "EQ_LAUNCH": "INPROGRESS",
             "SUCCESSFUL_RESPONSE_UPLOAD": "COMPLETE",
             "COMPLETED_BY_PHONE": "COMPLETEDBYPHONE",
+            "RESPONDENT_ENROLED": "NOTSTARTED",
+            "ACCESS_CODE_AUTHENTICATION_ATTEMPT": "NOTSTARTED",
+            "NO_LONGER_REQUIRED": "NOLONGERREQUIRED",
```
LJBabbage, line 73: Does this need an assert somewhere  (NO_LONGER_REQUIRED)?

response_operations_ui/views/case.py
```diff
@@ -39,21 +39,25 @@ def get_response_statuses(ru_ref, error=None):
...
 
+    # The statuses variable is a dict that contains the action as the key (EQ_LAUNCH for example) and the value is
+    # the new state it will be in (e.g., COMPLETE, INPROGRESS, etc).
+    # The available_statuses are the statuses that we allow a case to switch to.  We currently only allow changing to
+    # COMPLETEDBYPHONE and NOLONGERREQUIRED, and no other, states via the UI.
+    # Historically we've not allowed changing back to NOTSTARTED as case wasn't smart enough to reset the case to a
+    # state where it would start sending communications out (e.g., reminder emails), if this changes then we can add it
+    # to this list though it would need to be written to be a bit more clever about which radio buttons to present as
+    # this implementation is very basic.
+    allowed_statuses = {"COMPLETEDBYPHONE", "NOLONGERREQUIRED"}
```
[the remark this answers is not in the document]
insacuri: No reason, other then the person who first wrote it did something weird >_<  Changing and moving it now!

tests/views/test_change_response_status.py
insacuri: Yeah that's fair.  Added one.

LJBabbage (approved):

PR #756, opened 2022-08-01
Switch on Role Based Access

arroyoAle:
# What and why?
Sets the role based access flag to true
# How to test?

# Trello

### DIFF
--- a/_infra/helm/response-operations-ui/values.yaml
+++ b/_infra/helm/response-operations-ui/values.yaml
@@ -80,4 +80,4 @@ test:
   enabled: false
   
 features:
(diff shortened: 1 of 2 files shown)


_infra/helm/response-operations-ui/values.yaml
```diff
@@ -80,4 +80,4 @@ test:
   enabled: false
   
 features:
-  isRoleBasedAccessEnabled: false
+  isRoleBasedAccessEnabled: ture
```
insacuri, line 83: You misspelt true in a one line PR >_<

insacuri (approved):

PR #757, opened 2022-08-05
BUG: Changing case from a completed state to NOTSTARTED state in response ops

SteveScorfield:
# What and why?
The functionality to turn a case to 'not started' when the status has been set to 'completed by phone' and 'no longer required' had been removed. This was required by the business users. 

The functionality has been added back in. The issues of  'not started' radio buttons appearing when the case was already in the 'not started' status has been resolved. They no longer show.

# How to test?
Change to completed by phone/no longer required:
Click reporting unit -> enter a reporting unit (I used 0001) -> Click search -> Click the number under RU reference -> Click the number under Survey -> Click 'Change' under the Status of the Collection Exercise table -> Click either 'Completed by phone' or 'No longer required' (these should be the only radio buttons) - > Confirm - > Redirect back to reporting unit page

Change back to not started:
Click the number under Survey -> Click 'View' under the Status of the Collection Exercise table -> Click 'Not started' - > Confirm

Testing can also be achieved by running the unit tests specific to this

# Trello
https://trello.com/c/sYxpWZPN/1898-bug-changing-case-from-a-completed-state-to-notstarted-state-in-response-ops-track-time


SteveScorfield:
[BUG: Changing case from a completed state to NOTSTARTED state in response ops (track time)](https://trello.com/c/sYxpWZPN/1898-bug-changing-case-from-a-completed-state-to-notstarted-state-in-response-ops-track-time)

response_operations_ui/views/case.py
```diff
@@ -64,12 +64,17 @@ def get_response_statuses(ru_ref, error=None):
...
+    case_events = get_case_events_by_case_id(case_id=case_id)
 
     if case_group_status == "COMPLETE":
-        case_events = get_case_events_by_case_id(case_id=case_id)
         case_event = get_case_event_for_seft_or_eq(case_events)
         completed_respondent = get_user_from_case_events(case_event)
 
+    # Below code was added to remove duplicated and not needed 'Not started' radio buttons when the case is set to
+    # 'NOTSTARTED'
+    if case_group_status == "NOTSTARTED":
```
LJBabbage, line 75: We should try to absorb this into initial allowed status, so we don't have to step through the results again, it should also be explicit rather than removing anything that has a value of not started. I have had a quick play and come up with this, I have done a very quick refactor as well, because we twice checked to see if the case_group is complete and have a redundant case events call if it is

```
def allowed_statuses(case_group_status):
    return {"COMPLETEDBYPHONE", "NOLONGERREQUIRED"} if case_group_status == "NOTSTARTED" else {"NOTSTARTED"}
```

```
    case_groups = case_controller.get_case_groups_by_business_party_id(reporting_unit["id"])
    case_group = case_controller.get_case_group_by_collection_exercise(case_groups, exercise["id"])
    case_group_status = case_group["caseGroupStatus"]
    case_id = get_case_by_case_group_id(case_group["id"]).get("id")
    is_case_complete = case_group_status in COMPLETE_STATE
    completed_timestamp = get_timestamp_for_completed_case_event(case_id) if is_case_complete else None
    
    if is_case_complete:
        case_events = get_case_events_by_case_id(case_id=case_id)
        case_event = get_case_event_for_seft_or_eq(case_events)
        completed_respondent = get_user_from_case_events(case_event)

    case_statuses_allowed = case_controller.allowed_statuses(case_group_status)
    available_statuses = {
        event: map_ce_response_status(status)
        for event, status in statuses.items()
        if status in case_statuses_allowed
```

LJBabbage (changes requested):

response_operations_ui/views/case.py
LJBabbage: You might have to tweak this as I am think we might also have INPROGRESS as a state

response_operations_ui/views/case.py
```diff
@@ -40,22 +39,23 @@ def get_response_statuses(ru_ref, error=None):
...
+
     reporting_unit = party_controller.get_business_by_ru_ref(ru_ref)
 
-    # The possible_transitions_for_case variable is a dict that contains the action as the key (EQ_LAUNCH for example)
-    # and the value is the new state it will be in (e.g., COMPLETE, INPROGRESS, etc).
-    # The allowed_transitions_for_case are the statuses that we allow a case to switch to.  We currently only allow
-    # a subset of them via the UI as case can't handle switching between certain states correctly.
-    possible_transitions_for_case = case_controller.get_available_case_group_statuses_direct(exercise["id"], ru_ref)
-    allowed_transitions_for_case = {
+    statuses = case_controller.get_available_case_group_statuses_direct(exercise["id"], ru_ref)
```
insacuri, line 52: Why'd you change the names back?  I made them more wordy to actually explain what they are.  `statuses` and `available_statuses` don't mean as much as `possible_transitions_for_case` and `allowed_transitions_for_case` I think

insacuri (changes requested):

response_operations_ui/views/case.py
insacuri: Agreed, I was going to make a comment similar to this.  We could/should maybe go a bit further and for all the states that we allow changes to (notstarted, nolongerrequired and completedbyphone I think) we get the list of all the allowed transitions, filter out all the ones which just go back to themselves (like NOTSTARTED -> action -> NOTSTARTED) then display the list.

The way the PR is now probably works as expected but is still a bit hard to follow I think.

response_operations_ui/views/case.py
SteveScorfield: I can change it back

tests/views/test_change_response_status.py
```diff
@@ -289,3 +296,45 @@ def test_respondent_name_not_in_metadata_for_completed_case_event(self, mock_req
...
+        mock_request.get(url_update_case_group_status, json=case_groups_not_started)
+        mock_request.get(url_get_case_by_case_group_id, json=[case])
+
+        response = self.client.get(f"/case/{ru_ref}/response-status?survey={short_name}&period={period}")
+
+        data = response.data
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(b"19000001", data)
+        self.assertIn(b"Bolts and Ratchets", data)
+        self.assertIn(b"221 BLOCKS", data)
```
LJBabbage, line 317: Not major, but we really should start to assert against just what we are testing, so this can be removed. The whole test class really needs a refactor, but definitely not for this story

LJBabbage (approved):
Works ok, some minor stuff that doesn't need me to block

tests/test_data/case/case_group_statuses_completed_to_notstarted.json
```diff
@@ -0,0 +1,3 @@
+{
+  "COMPLETED_TO_NOTSTARTED":"NOTSTARTED"
```
insacuri, line 2: What is this COMPLETED_TO_NOTSTARTED event?

insacuri (approved):

PR #758, opened 2022-08-09
Delete seft collection instrument

LJBabbage:
# What and why?
This PR updates SEFT collection instruments to call the delete end point in the collection instrument service rather than unlink
 
# How to test?
Deploy this [PR](https://github.com/ONSdigital/ras-collection-instrument/pull/282) to your environment and add and remove seft collection instruments. The other PR has more detailed instructions
# Trello

### DIFF
--- a/response_operations_ui/__init__.py
+++ b/response_operations_ui/__init__.py
@@ -6,12 +6,12 @@ import redis
 from flask import Flask, flash, redirect, session, url_for
 from flask_assets import Environment
 from flask_login import LoginManager
+from flask_session import Session
 from flask_talisman import Talisman
 from flask_wtf.csrf import CSRFError, CSRFProtect
 from structlog import wrap_logger
 
 from config import Config
-from flask_session import Session
 from response_operations_ui.common.jinja_filters import filter_blueprint
 from response_operations_ui.controllers.uaa_controller import user_has_permission
 from response_operations_ui.logger_config import logger_initial_config
--- a/response_operations_ui/__init__.py
+++ b/response_operations_ui/__init__.py
@@ -6,12 +6,12 @@ import redis
 from flask import Flask, flash, redirect, session, url_for
 from flask_assets import Environment
 from flask_login import LoginManager
-from flask_session import Session
 from flask_talisman import Talisman
 from flask_wtf.csrf import CSRFError, CSRFProtect
 from structlog import wrap_logger
 
 from config import Config
+from flask_session import Session
 from response_operations_ui.common.jinja_filters import filter_blueprint
 from response_operations_ui.controllers.uaa_controller import user_has_permission
 from response_operations_ui.logger_config import logger_initial_config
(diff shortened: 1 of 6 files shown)


response_operations_ui/__init__.py
```diff
@@ -6,12 +6,12 @@
 from flask import Flask, flash, redirect, session, url_for
 from flask_assets import Environment
 from flask_login import LoginManager
-from flask_session import Session
 from flask_talisman import Talisman
 from flask_wtf.csrf import CSRFError, CSRFProtect
 from structlog import wrap_logger
 
 from config import Config
+from flask_session import Session
```
LJBabbage, line 14: Linting decided to move this and failed when moved back :man_shrugging:

response_operations_ui/controllers/collection_instrument_controllers.py
```diff
@@ -182,6 +182,24 @@ def unlink_collection_instrument(ce_id, ci_id):
     return True
 
 
+def delete_seft_collection_instrument(ci_id: str) -> bool:
+    """Deletes a SEFT collection instrument
```
LJBabbage, line 186: I have reduced logging compared to others, I just don't think we need the noise, we just need to know the cause if it fails. There will also be a request call in the logs anyway

response_operations_ui/templates/ce-seft-instrument.html
```diff
@@ -10,7 +10,6 @@
 
 {% block main %}
     {% include 'collection_exercise/ce-success-panel.html' %}
-    {% include 'collection_exercise/ce-handle-errors.html' %}
```
LJBabbage, line 13: No idea why we had 2 * {% include 'partials/error-box.html' %}  and a {% include 'collection_exercise/ce-handle-errors.html' %} On error I was getting multiple error boxes appearing, for this and other errors this page could generate

response_operations_ui/views/collection_exercise.py
```diff
@@ -975,13 +988,25 @@ def get_seft_collection_instrument(short_name, period):
 @login_required
 def post_seft_collection_instrument(short_name, period):
     verify_permission("surveys.edit", session)
-    if "unselect-ci" in request.form:
-        return _unselect_seft_collection_instrument(short_name, period)
+    if "delete-ci" in request.form:
```
LJBabbage, line 991: Confirmed with Gareth, that unlinking a SEFT is now invalid and should be just delete to avoid orphaned data

response_operations_ui/views/collection_exercise.py
```diff
@@ -425,7 +425,21 @@ def _upload_collection_instrument(short_name, period):
 
 
 def _unselect_collection_instrument(short_name, period):
-    success_panel = _unlink_collection_instrument()
+    success_panel = "Collection instrument removed"
```
LJBabbage, line 428: Because both eQ and SEFT was using the unlink functionality it was split, now SEFT doesn't use it I put it back into one

JacobHarrisonONS (approved):

PR #760, opened 2022-08-12
Add optional param for number of cases

insacuri:
# What and why?

Currently it's only possible to see the 12 most recent cases, if you want more (say, to change a status) then we have to go to the database for this.  This PR adds a 'hidden' url parameter, as in there is no UI element for it, `max-number-of-cases` that you can add to the `/reporting-units/<ru_ref>/surveys/<survey_id>` url to make it display more.

This is only really a temporary 'fix' until the work is done to make it possible to easily see all the cases for a survey.  It should, however, reduce the ops burden for us until it's implemented.

# How to test?

# Trello

### DIFF
--- a/_infra/helm/response-operations-ui/Chart.yaml
+++ b/_infra/helm/response-operations-ui/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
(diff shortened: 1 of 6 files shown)


anwilkie (approved):
Looks okay to me. Though, I noticed that if you set the `max-number-of-cases` to 0 in the URL, you get a 500 error. I thought that was worth noting.

PR #761, opened 2022-08-19
increment design system version to 55.1.0

anwilkie:
# What and why?
The design system needed to be upgraded to the latest version of 55.1.0. There have been a few breaking changes since the previous version we were on, although fortunately, they don't seem to affect this service. Here are the breaking changes examined:

In version 54, the `collapsible` component was updated to remove the `variants` variable. We never use this component anywhere, so this shouldn't affect us.
In version 54, the `input` component was updates so that any inputs that use `params.prefix` or `params.suffix` must set additional variables to render. While we do use `input` in places, we never set `params.prefix` or `params.suffix`, so this shouldn't affect us.
In version 55, the `hero` component was updated with a breaking change. We never use this component anywhere, so this shouldn't affect us.

# How to test?
Deploy to your environment and navigate around to make sure things are rendered correctly.

# Trello
[Card](https://trello.com/c/41urU4MG)


insacuri:
Superceded by another PR that bumped it to 55.2.0

PR #762, opened 2022-08-22
Refactor and combine password forms.  Also set min password length to 12.

insacuri:
# What and why?

# How to test?

# Trello


LJBabbage (approved):
Deployed and tested before and after this change and both work as expected

PR #764, opened 2022-08-30
Update username tests to be emails

LJBabbage:
# What and why?
This PR updates the usernames to email address to make it more consistent with what our user now use
# How to test?
Make sure nothing has been missed.
# Trello


tests/controllers/test_uaa_controller.py
```diff
@@ -24,13 +24,13 @@
 with open(f"{project_root}/test_data/uaa/email_search_user.json") as fp:
     uaa_user_search_email = json.load(fp)
 
-user_id = "fe2dc842-b3b3-4647-8317-858dab82ab94"
+user_id = "some.one@ons.gov.uk"
```
insacuri, line 27: Does the user_id really change with this?  Each user should still have a uuid that identifies them that's separate from the username and email

tests/test_data/uaa/create_user_already_exists.json
```diff
@@ -1,5 +1,5 @@
 {
-  "user_id": "0620883d-83f4-4e3e-87a4-7e7bcec5b71a",
+  "user_id": "some.one@ons.gov.uk",
```
insacuri, line 2: Same with the other comment above (and I won't do it for each other one as it'd be repetitive).  Does the id change?  I don't think it does

tests/test_data/uaa/updated_name.json
```diff
@@ -1,11 +1,11 @@
 {
-  "id": "fe2dc842-b3b3-4647-8317-858dab82ab94",
+  "id": "ons@ons.fake",
```
insacuri, line 2: This definitely isn't right!

tests/views/test_accounts.py
```diff
@@ -16,16 +16,18 @@
 with open(f"{project_root}/test_data/uaa/user_by_id.json") as json_data:
     uaa_user_by_id_json = json.load(json_data)
 
-test_email = "fake@ons.gov.uk"
-user_id = "fe2dc842-b3b3-4647-8317-858dab82ab94"
+user_id = "some.one@ons.gov.uk"
```
insacuri, line 19: Same user id comment

insacuri (changes requested):
Pretty much the same comment, the username and email is okay to change but i'm almost certain the id should stay as a uuid

tests/controllers/test_uaa_controller.py
LJBabbage: Oh yeah, I have mistaken user_id and user_name on a few, darn

response_operations_ui/__init__.py
```diff
@@ -6,12 +6,12 @@
 from flask import Flask, flash, redirect, session, url_for
 from flask_assets import Environment
 from flask_login import LoginManager
-from flask_session import Session
 from flask_talisman import Talisman
 from flask_wtf.csrf import CSRFError, CSRFProtect
 from structlog import wrap_logger
 
 from config import Config
+from flask_session import Session
```
insacuri, line 14: ah, this'll fail the linting btw.  The reason it's happening is that I think running the unit tests at the top level creates a `flask_session` folder to store the session locally.  isort sees this and goes 'oh flask_session folder, that must mean it's a local, not a 3rd party package' and puts it with all the other local ones.

LJBabbage: Bit of an odd one that, and naturally it passes local. Thanks

tests/controllers/test_uaa_controller.py
```diff
@@ -25,12 +25,12 @@
     uaa_user_search_email = json.load(fp)
 
 user_id = "fe2dc842-b3b3-4647-8317-858dab82ab94"
+user_email = "some.one@ons.gov.uk"
 group_id = "9da7cfd5-95d0-455b-9005-02ce638e56c9"
 fake_group_id = "eaf2988b-99b4-423b-9a09-63b1d6f07677"
-fake_user_id = "56e97a1b-2188-4989-8342-199b83c505ce"
+fake_user_id = "fake@ons.gov.uk"
```
insacuri, line 31: Minor, but the fake user id should also be a uuid

tests/views/test_accounts.py
```diff
@@ -16,16 +16,18 @@
 with open(f"{project_root}/test_data/uaa/user_by_id.json") as json_data:
     uaa_user_by_id_json = json.load(json_data)
 
-test_email = "fake@ons.gov.uk"
 user_id = "fe2dc842-b3b3-4647-8317-858dab82ab94"
+user_email = user_id
```
insacuri, line 20: Sorry I missed this one.  The user email should be an actual email, not an id.   If you revert the change back to `test_email = "fake@ons.gov.uk"` then it should be fine.

insacuri (approved):
LGTM after that last comment is addressed

PR #765, opened 2022-08-30
Call new sample summary status endpoint

insacuri:
# What and why?

Prior to this PR, loading a sample meant the sample service would hit the database once to load the sample unit, then again to get the count of all the sample units to see if it was done yet.  If it wasn't then it'd not do anything else, if all the sample units had been loaded then it would transition the state of the sample summary to ACTIVE (which let the things that cared know that the sample load was complete).

This new PR changes that.  During the sample load it does the 1 call per unit that the current one does to save it to the database, but now there is a separate http endpoint that needs to be called to check if all the sample units for a sample summary have been loaded, and then change the status. 

This means that response-operations-ui needs to call that endpoint and find it out if it's done by itself.  We only call the endpoint if the collection exercise is still being created (so basically in the CREATED or SCHEDULED states) AND while the sample summary is still in the INIT state.  If both these are true then we need to keep checking if it's active yet.  This check is only done on 2 pages, the collection exercise page and the sample & CI page for that collection exercise.

This sample PR (https://github.com/ONSdigital/rm-sample-service/pull/232) is tied to this one, and they need to go in as a pair.

# How to test?

- Deploy both PRs to your enviornment
- Create a collection exercise and load a large-ish sample. While it's loading the state should stay as INIT and you should see calls in response-ops hitting the new check endpoint.
- Once they're all loaded, you should see the check endpoint transition the state of the sample summary to ACTIVE.
- Once the state is active, response-ops shouldn't hit that endpoint anymore until you delete the sample.
- Also test deleting the sample and adding it again, the calls should resume.
- Set the exercise live.  Everything should be the same and there should be no more calls to this new endpoint.

# Trello


response_operations_ui/controllers/sample_controllers.py
```diff
@@ -28,6 +28,38 @@ def get_sample_summary(sample_summary_id):
...
+    :param sample_summary_id:
+    :return:
+    """
+    logger.info("Checking sample summary state", sample_summary_id=sample_summary_id)
+    url = f'{app.config["SAMPLE_URL"]}/samples/samplesummary/{sample_summary_id}/check-all-units-present'
+    response = requests.get(url, auth=app.config["BASIC_AUTH"])
+
+    try:
+        response.raise_for_status()
+    except HTTPError:
```
pricem14pc, line 45: This does highlight the need to re-work the user journey around this process. An explicit button click "check sample load status" for example (rather than the page load) to transition the sample summary state, would be a nicer user experience. Presumably the CE page would throw a 500 if this exception was raised?

insacuri: Yeah that's fair.  I've just changed it so that it'll flash an error telling the user to refresh the page to try again.  Refreshing might not help every error, but if it's the error wasn't transient (like if the uuid for the sample summary was wrong for some reason) then we'd be in the same situation now needing manual intervention.

pricem14pc (approved):
Functionally this is behaving as expected with the changes introduced in sample-service https://github.com/ONSdigital/rm-sample-service/pull/232.

While the new tests cover some niche scenarios in handling particular errors, these are very difficult to replicate and I've been unable to do so and see the impact of the new flash message in the UI.

Following these changes we should review the wider user sample load journey and whether an explicit Rops action should be introduced to check and transition the status summary.

PR #766, opened 2022-09-07
Add sample count

insacuri:
# What and why?

# How to test?

# Trello


anwilkie (approved):
Ran acceptance test, and they work. Also, I tried loading a large sample file with 2000 sample units and I watched the sample count go up each time I refreshed.

PR #767, opened 2022-09-07
Update design system version

insacuri:
# What and why?

# How to test?

# Trello


JacobHarrisonONS (approved):
Passing acceptance tests run with helm

PR #768, opened 2022-09-09
added show password field to sign-in page

anwilkie:
# What and why?
Unlike frontstage/elsewhere on response-ops, the sign-in page on response-ops did not have an option to show the password. This PR fixes it to make it more accessible and consistent.

# How to test?
Access response-ops and check that the show password option is there and that it works. Also, run acceptance tests.

N.B 
This was extended to remove attributes from all password fields and correctly display errors, because of that activate-account and change password also need testing as well as the login page
# Trello
[Card](https://trello.com/c/FRHWS9Sc)


LJBabbage (approved):
Any reason we are mixing speech marks?

response_operations_ui/templates/sign_in.html
```diff
@@ -106,17 +107,17 @@ <h2>Sign in</h2>
...
-                        "text": "Password"
+                        "text": 'Password'
                     },
                     "attributes": {
-                        "required": true
+                        "autocomplete": 'off'
                     },
-                    "error": error
+                    "showPasswordText": "Show password",
+                    "error": passwordError
```
insacuri, line 120: The error changed but the code didn't change with it?  and the required true is gone?  Did you test this as my gut tells me this is wrong >_<

insacuri (changes requested):

response_operations_ui/templates/sign_in.html
LJBabbage: It looks like you might be right, the form still blocks it and the correct error comes up, however that is server side as this would deal with it client side. We also lose the red box, I will update it for Adam

LJBabbage: Argg, onsPassword doesn't pass attributes, so we can use them

insacuri (approved):

response_operations_ui/templates/account/activate-account.html
```diff
@@ -88,13 +88,10 @@ <h1>Activate your account</h1>
...
                         "label": {
-                            "text": 'Password'
-                        },
-                        "attributes": {
-                            "autocomplete": 'off'
+                            "text": "Password"
                         },
                         "showPasswordText": "Show password",
-                        "error": passwordError
+                        "error": {"text": errorData[0]["text"] } if errorData else None
```
LJBabbage, line 94: This isn't ideal and I will raise a card as this page has a number of issues. At present everything gets returned via form.password.errors and because of that there will never be an error from form.password_confirm.errors which means there is redundancy in this file. Because nothing is attributed to confirm password having an error field on the component is also never going to work and why I removed it. 
For the moment the first password field will deal with errors and show the first elements text

response_operations_ui/templates/account/activate-account.html
```diff
@@ -88,13 +88,10 @@ <h1>Activate your account</h1>
                         "id": "password",
                         "name": "password",
                         "label": {
-                            "text": 'Password'
-                        },
-                        "attributes": {
```
LJBabbage, line 93: Password fields for DS don't current support attributes, so these are redundant

insacuri (approved):

PR #771, opened 2022-09-13
Remove change of username, update username on email change

LJBabbage:
# What and why?
This PR removes the ability for user to change their username, instead merging the functionality in with email change (as the 2 field are now tied together)

# How to test?
From the UI login as normal then go to My Account, make sure you can change your email but not username and the order matches the card. Select email change and then change it to something new, make sure the verification email is sent and when you click on it both the user_name and email are updated. Now login with your new username/email to make sure it works. Once confirmed try some unhappy paths, i.e change your email to one that already exists, hit the old end point for changing username etc

N.B I have reduced logging for successful paths, I only really want to know on error
# Trello


response_operations_ui/controllers/uaa_controller.py
```diff
@@ -71,11 +71,10 @@ def login_admin():
         abort(response.status_code)
 
 
-def get_user_by_email(email: str, access_token=None) -> dict | None:
+def get_user_by_filter(user_filter: str, access_token=None) -> dict | None:
```
LJBabbage, line 74: Instead of enter another load of boiler plate using the filter for something else I rationalised the call

response_operations_ui/views/accounts.py
```diff
@@ -104,66 +101,97 @@ def change_account_name():
...
                 )
                 return redirect(url_for("logout_bp.logout"))
-            except NotifyError as e:
-                logger.error("Error sending 'email change' email to Notify Gateway", msg=e.description)
-                flash("Something went wrong while updating your email. Please try again", category="error")
+
+            flash("Something went wrong while updating your email. Please try again", category="error")
         else:
-            return redirect(url_for("account_bp.get_my_account"))
+            logger.info(f"Unable to change username and email for {user_id}, email already used")
```
LJBabbage, line 128: Feasibly you could use the new email here, but because I would have to obfuscate I didn't feel it added much

response_operations_ui/views/accounts.py
```diff
@@ -104,66 +101,97 @@ def change_account_name():
...
+
 @account_bp.route("/verify-email/<token>", methods=["GET"])
-@login_required
 def verify_email(token):
     try:
         duration = app.config["UPDATE_ACCOUNT_EMAIL_TOKEN_EXPIRY"]
         json_token = token_decoder.decode_email_token(token, duration)
         token_dict = json.loads(json_token)
-        user_id = session["user_id"]
-        if token_dict["user_id"] == user_id:
```
LJBabbage, line 153: This has been checked with IA, we don't need them to login on confirming an email/username change

tests/views/test_accounts.py
```diff
@@ -171,219 +172,141 @@ def test_name_uaa_error(self, mock_request):
...
-                data={"username": "uaa_user!"},
-            )
-            self.assertIn(
-                b"Username can only contain lowercase letters, numbers, and special characters (`.`, `@`, and `_`)",
-                response.data,
-            )
 
-    @requests_mock.mock()
-    def test_username_max_length_and_wrong_character_errors(self, mock_request):
-        with patch("response_operations_ui.views.accounts.NotifyController") as mock_notify:
```
LJBabbage, line 253: There is loads of unnecessary mocking in these tests, which just adds noise and makes them harder to read. This would fail on form validation and never get near notify, so there is no need to mock it. I think I removed 40 lines of code in 4 tests :upside_down_face:

response_operations_ui/views/accounts.py
```diff
@@ -104,66 +101,101 @@ def change_account_name():
...
+
+        user_from_uaa = uaa_controller.get_user_by_id(token_dict["user_id"])
+
+        if user_from_uaa is None:
+            return render_template("request-new-account-error.html")
+
+        user_from_uaa["emails"][0]["value"] = token_dict["email"]
+        user_from_uaa["userName"] = token_dict["email"]
+
+        errors = uaa_controller.update_user_account(user_from_uaa)
```
LJBabbage, line 190: You could call _can_change_username_and_email however the likelihood of it happening are very slim and you increase the amount of calls. There is error catching here anyway if there is an issue, so decide to not to use it

JacobHarrisonONS (approved):

PR #773, opened 2022-09-14
Update jinja2 to match new DS spec

mfcrocker:
# What and why?
Design System 55.2 tidied up their jinja2 if statements to not need to null check, and we should do the same

# How to test?

# Trello
https://trello.com/c/NiBGbXy3/1980-spike-investigate-chainable-jinja-changes-introduced-by-ds-1-day


insacuri (approved):
LGTM.  I'm just assuming you've tested it!

PR #775, opened 2022-09-16
Remove survey ci classifiers

insacuri:
# What and why?

It was noticed that nothing actually used the `ci_selectors` that was created in every `build_collection_exercise_details` call.  To save on extra calls to services that don't need it, we removed it if nothing was using it.

This includes all the test data that had it and all the mock calls in the tests that were made.

# How to test?

# Trello


anwilkie (approved):
Acceptance tests ran okay in my environment

PR #776, opened 2022-09-20
Change username to email on sign-in

LJBabbage:
# What and why?
Update Username on signing page to email address
# How to test?
Make sure the change has been made when logging in
# Trello


response_operations_ui/templates/sign_in.html
```diff
@@ -93,7 +93,7 @@ <h2>Sign in</h2>
                     "name": "username",
```
LJBabbage, line 93: This is still the username (email and username are now the same value) and why it hasn't been changed. It's purely to make things easier for the user to change it to Email address

insacuri (approved):

PR #779, opened 2022-09-22
Adding link to ru ref

arroyoAle:
# What and why?
This pr adds a link to the ru ref already present in a secure message thread and allows users to go directly to see that ru from the message.
# How to test?
Make sure the link takes a user to the correct ru
# Trello


LJBabbage (approved):
Works as expected. It really should be in the backend, but as the rest of the file is like that :see_no_evil:

PR #780, opened 2022-09-22
Allow messages to be close when respondent has been deleted

LJBabbage:
# What and why?
A rops user should be able to close a conversation even if the respondent has been removed, this PR updates the code base to do this. This was leading to 500 in some situations
# How to test?
This realistically needs access to the DB to test properly as we have no UI to delete a respondent. Create a collection exercise as you normally would and make sure you login to frontstage with a respondent. Start a conversation with them using rops, send a few messages between the 2 making sure it works as normal. Now go to party in the db and find your user and delete the respondent (I changed their UUID so I could toggle without clearing out all the associated tables, which is effectively the same, but did delete as well). Now check you can close the conversation successfully. Also check that once a conversation has been closed it can't be re-opened
# Trello


response_operations_ui/views/messages.py
```diff
@@ -188,6 +188,17 @@ def view_conversation(thread_id):
                 thread_data=thread_conversation,
             )
 
+    close_conversation_url = url_for(
```
LJBabbage, line 191: I don't like this function, however also hate it on the frontend, the whole function definitely needs a refactor as it's difficult to follow

response_operations_ui/views/messages.py
```diff
@@ -627,14 +639,16 @@ def close_conversation(thread_id):
         )
 
     thread_conversation = message_controllers.get_conversation(thread_id)
-    refined_thread = [_refine(message) for message in reversed(thread_conversation["messages"])]
+    last_message = _refine(thread_conversation["messages"][-1])
```
LJBabbage, line 642: We only need the last message

tests/test_data/message/thread_missing_respondent.json
```diff
@@ -1,196 +1,106 @@
 {
-    "category": "SURVEY",
```
LJBabbage, line 2: This file was bloated and a bit of a mess, we only need 2 messages, one from the rops user and one from the respondent

response_operations_ui/views/messages.py
```diff
@@ -921,6 +935,7 @@ def _get_to_name(message: dict) -> str:
         return f"{message.get('@msg_to')[0].get('firstName')} {message.get('@msg_to')[0].get('lastName')}"
     except (IndexError, TypeError):
         logger.info("Failed to retrieve message to name", message_id=message.get("msg_id"))
+        return "Deleted respondent"
```
LJBabbage, line 938: This isn't ideal but does align with get_from_name

tests/views/test_message.py
```diff
@@ -415,20 +417,29 @@ def test_get_thread_success(self, mock_request, mock_get_jwt):
         self.assertIn("MISC".encode(), response.data)
 
     @requests_mock.mock()
-    @patch("response_operations_ui.controllers.message_controllers._get_jwt")
```
LJBabbage, line 418: The name has been changed but I have also remove unnecessary mocking

tests/views/test_message.py
```diff
@@ -415,20 +417,29 @@ def test_get_thread_success(self, mock_request, mock_get_jwt):
...
-        response = self.client.get("/messages/threads/fb0e79bd-e132-4f4f-a7fd-5e8c6b41b9af")
-        self.assertIn("Cannot reply or modify conversation as the respondent has been deleted".encode(), response.data)
+    def test_open_thread_with_deleted_respondent(self, mock_request):
+        response = self._get_thread_with_deleted_respondent(mock_request, missing_user_json)
+        self.assertIn("You can no longer send a message as this respondent has been deleted".encode(), response.data)
+        self.assertIn("Close conversation".encode(), response.data)
+
+    @requests_mock.mock()
+    def test_closed_thread_with_deleted_respondent(self, mock_request):
+        missing_user_json_closed_thread = missing_user_json.copy()
```
insacuri, line 427: Does this work as expected? If you make a copy then modify it, doesn't the original get modified too?  I've been using deepcopy to make sure they're completely brand new, but if this works in the same way then great.

LJBabbage: Worth checking

LJBabbage: Copy is fine https://docs.python.org/3/library/copy.html as there isn't any compound objects I don't need to copy them as well, which would be deepcopy

anwilkie (commented):
Upon deploying and testing this, I was able to confirm that deleting a respondent/changing the respondent's ID allowed me to close a conversation for the user (and not be able to open it), so that part works.

However, I encountered an oddity when sending messages back and forth between the r-ops user and the respondent where I could initiate a new conversation thread (and have the respondent respond to it) but when I tried adding more replies to the conversation on r-ops, I got a popup about the message not sending, and looking at the logs, I got a `CSRF token is missing` log line, followed by:

`{"event": "Message sending failed due to API Error", "severity": "error", "level": "error", "service": "response-operations-ui", "exception": "Traceback (most recent call last):\n  File \"/app/response_operations_ui/controllers/message_controllers.py\", line 163, in send_message\n    response.raise_for_status()\n  File \"/usr/local/lib/python3.10/site-packages/requests/models.py\", line 1022, in raise_for_status\n    raise HTTPError(http_error_msg, response=self)\nrequests.exceptions.HTTPError: 400 Client Error: BAD REQUEST for url: http://secure-message.wilkia.svc.cluster.local:8080/messages", "created_at": "2022-09-27T09:241664270675"}`

I don't encounter this issue on the main branch (or other PRs in-flight), so this PR is causing it.

LJBabbage:
> Upon deploying and testing this, I was able to confirm that deleting a respondent/changing the respondent's ID allowed me to close a conversation for the user (and not be able to open it), so that part works.
> 
> ...

CSRF is missing in many places on rops and not related to this issue, but naturally something else happened so I have dug into the logs (yours) so the issue comes from secure message not being able to validate the respondent and their details
image This means you must of sent a message to a respondent that was deleted, not what was changed but still in the same area so happy to look, now I can only think of 2 ways that can happen, from the conversation page or from the ru page. From your comment it sounds like this was a new message and not the old conversation which worked as expected? Now if it was the ru page and you did just change just the uuid of the user but not the respondents enrolment they will be still connected, at which point if you do sent a message (to a respondent now with a different id) you will go down a dark alley. However this is not a scenario you will ever encounter as if you delete a respondent properly the db will also force the enrolments to be removed as well, removing this path. 

If you can test again, fully deleting the respondent from the db I reckon it will be ok.

The only other way I think this could happen is if a page was left open with the text box and send button available, the respondent then removed and then the send button clicked.

anwilkie:
I retested it today and I'm encountering the issue with not being able to send the message before I change anything about the respondent's data in the database. I'm concerned that the code changes in this PR might have accidentally introduced a new bug to response-ops (that occurs regardless of whether the respondent is deleted). I've done the exact same steps before deploying the PR and the message sending is okay; it's only after I deploy this PR that I run into this issue.

anwilkie:
I've tested this again and I'm still getting the "CSRF token is missing" error and I can't send message replies to the respondent.

anwilkie (approved):
Tested this out again, and the message sending works okay now.

PR #782, opened 2022-09-27
fix url on case page

anwilkie:
# What and why?
There was a bug on response-ops where, if a user goes to the case details page for a reporting unit's survey and then refreshes the page, the `cancel` and `close` hyperlinks wouldn't work because they would redirect the user back to the same page. This PR fixes this by explicitly sending the user back to the reporting unit survey details page.

# How to test?
* Run acceptance tests.
* Access the case details page for a reporting unit's survey, refresh the page, and click on `cancel` or `close`. You should get redirected to the previous reporting units survey page without any errors.

# Trello
[Card](https://trello.com/c/xsTGUDP2)

### DIFF
--- a/response_operations_ui/templates/response-status.html
+++ b/response_operations_ui/templates/response-status.html
@@ -93,7 +93,7 @@
                 })
             }}
             {% if is_case_complete %}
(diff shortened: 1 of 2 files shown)


insacuri (approved):
LGTM 👍

PR #783, opened 2022-10-07
BUG - rOps sampled businesses display 500 error in RU tab

SteveScorfield:
# What and why?
An error has been found, where when a sample is loaded with a business that does not exist yet for a survey (before go live), the user searches for the RU ref and clicks the RU ref link. This is an unhelpful 500 error page. 

To create a better user experience, a new error page will be shown for this particular case.

# How to test?
1. Create a new CE
2. Load a sample file with a company RU ref that does not exist yet for a survey  
3. Don't set survey to live
4. Go into Reporting Units tab
5. Load RU of sampled, but not set ready for live business
6. Click on RU to load business details
7. A specific error page should be present for the failed RU ref

# Trello
https://trello.com/c/wB7Vtrxr/2071-bug-rops-sampled-businesses-display-500-error-in-ru-tab-track-time

### DIFF
--- a/response_operations_ui/templates/errors/ru-error.html
+++ b/response_operations_ui/templates/errors/ru-error.html
@@ -1,0 +1,17 @@
+{% extends "layouts/base.html" %}
+{% set page_title = "Failed to retrieve RU Ref - ONS Business Surveys" %}
+
+{% block main %}
+    <h1 class="ons-u-mt-m">Failed to retrieve RU Ref</h1>
+    {% call
+        onsPanel({
+            "type": "error",
+            "classes": "ons-u-mb-l"
+        })
+    %}
+      <p>We were unable to retrieve RU Ref {{ ru_ref }}. </p>
+      <p>Please try again after the survey has gone live.</p>
+      <p>You may want to visit the <a href="/">homepage</a></p>
+      <p>If you still encounter problems please <a href="/contact-us">contact us</a>. We apologise for any inconvenience this may have caused.</p>
+    {% endcall %}
+{% endblock main %}
--- a/response_operations_ui/controllers/party_controller.py
+++ b/response_operations_ui/controllers/party_controller.py
@@ -10,7 +10,8 @@ from response_operations_ui.controllers.survey_controllers import get_survey_by_
 from response_operations_ui.exceptions.exceptions import (
     ApiError,
     SearchRespondentsException,
-    UpdateContactDetailsException,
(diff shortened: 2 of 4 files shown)


SteveScorfield:
[BUG - rOps sampled businesses display 500 error in RU tab (track time)](https://trello.com/c/wB7Vtrxr/2071-bug-rops-sampled-businesses-display-500-error-in-ru-tab-track-time)

response_operations_ui/templates/errors/ru-error.html
```diff
@@ -0,0 +1,16 @@
...
+{% block main %}
+    <h1 class="ons-u-mt-m">Reporting unit - not loaded</h1>
+    {% call
+        onsPanel({
+            "type": "error",
+            "classes": "ons-u-mb-l"
+        })
+    %}
+      <p>This reporting unit has been loaded using a sample, but the collection exercise is not yet ready for live. 
+        Please refer to your Survey Processing Centre contact if this is not corrected.</p>
```
arroyoAle, line 13: Should this not be `correct` instead of `corrected`, are we wanting them to do something or just wait until someone else to do something? Should we put a time for them to wait until they should contact someone if we want to keep it like this?

arroyoAle (changes requested):

insacuri (changes requested):
I think this is the wrong approach for this change.  With this change, if anything calls `get_business_by_ru_ref` for any reason, if it comes back with a 404 then you'll get that error page that's specific only to this one situation.

I think you need to:
- Not have a separate exception and remove all mentions of it
- In the view (the /reporting-units/<ru_ref> page where the error is happening), if you call that function and an API error is raised, you check if it's a 404, if so you render that template then and there instead of letting the exception bubble its way to the top of the app.

response_operations_ui/templates/errors/ru-error.html
SteveScorfield: This has been corrected. I used the wrong word when writing out the agreed message.

insacuri (changes requested):
Missing a unit test.  Otherwise looks fine

insacuri (approved):

response_operations_ui/templates/errors/ru-error.html
SteveScorfield: Adam M has checked this. I will dismiss this to get it into our branch.

PR #795, opened 2022-11-09
Upgrade python to 3.11 and update action versions

insacuri:
# What and why?

Upgrades python version to 3.11 and upgrades checkout and python-setup versions to something more current.

# How to test?

# Trello


matthew-robinson-ons (approved):
Upgrade functions correctly with acceptance tests passing.

PR #796, opened 2022-11-10
collection exercise event status display

anwilkie:
# What and why?

The purpose of this PR is to display collection exercise event status to a user via the interface on the survey list and collection exercise details page.

# How to test?

* Create a CE and fill out event dats for MPS, go live, reminders and nudge emails.
* In the database, edit some of the event statuses from `SCHEDULED` to `RETRY`, `PROCESSING`, and/or `FAILED`.
* Make sure event status indicators are appearing as intended (look at scenarios on Trello card).

NOTE: Because of a bug unrelated to this card, you can't currently create a fifth nudge email for a CE without causing a 500 error, but you can create the first four, and HTML code has been supplied to handle a fifth nudge email, so that one should be okay assuming the other events are okay.

# Trello
[Card](https://trello.com/c/qV7f3B9M)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ lint:
 	pipenv run flake8 --exclude ./node_modules
 
 lint-check: load-design-system-templates
-	pipenv check ./response_operations_ui ./tests -1 51457 -i 51499
+	pipenv check ./response_operations_ui ./tests -1 51457-i 51499
 	pipenv run isort . --check-only
 	pipenv run black --line-length 120 --check .
 	pipenv run djlint . 
--- a/Makefile
+++ b/Makefile
@@ -13,14 +13,14 @@ load-design-system-templates:
 	pipenv run ./scripts/load_templates.sh $(DESIGN_SYSTEM_VERSION)
 
 lint:
-	pipenv check ./response_operations_ui ./tests -i 51499
+	pipenv check ./response_operations_ui ./tests -i 51457 -i 51499
 	pipenv run isort .
 	pipenv run black --line-length 120 .
 	pipenv run djlint .
 	pipenv run flake8 --exclude ./node_modules
 
 lint-check: load-design-system-templates
-	#pipenv check ./response_operations_ui ./tests
+	pipenv check ./response_operations_ui ./tests -1 51457 -i 51499
 	pipenv run isort . --check-only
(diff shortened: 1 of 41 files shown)


Makefile
```diff
@@ -13,14 +13,14 @@ load-design-system-templates:
 	pipenv run ./scripts/load_templates.sh $(DESIGN_SYSTEM_VERSION)
 
 lint:
-	pipenv check ./response_operations_ui ./tests 
+	#pipenv check ./response_operations_ui ./tests
```
LJBabbage, line 16: We don't want to do this, it should only ignore the issue, see frontstage

pyproject.toml
```diff
@@ -2,4 +2,4 @@
 profile = "black"
 
 [tool.djlint]
-ignore="H021,H026"
+ignore="H021,H025,H026"
```
LJBabbage, line 5: What is H025 and why are we ignoring now?

LJBabbage (commented):
There is a lot of duplication in this PR with slightly different params, we should look to rationalise with loops and dicts. We should also try to do any logic in the backend

pyproject.toml
insacuri: H025 is an orphaned tag one.  With one of the recent updates to djlint, it's marking a few as orphaned even though they're actually fine; Those ones are concatenating strings together to make a complete`<a>` tag but the linter isn't recognising it

insacuri:
How do you test a RETRY status without just hard coding it for an event?

insacuri (changes requested):
Missing a unit test for it.  Probably needs one for the exericse list page and one for the exercise page itself where it shows the text appearing when the events have one or more of those states

anwilkie:
> Missing a unit test for it. Probably needs one for the exericse list page and one for the exercise page itself where it shows the text appearing when the events have one or more of those states

Done.

insacuri:
Not sure why this was closed >_<

LJBabbage (approved):
Taking a pragmatic approach on this, this does satisfy the requirements and works as expected. However the amount of logic on this page is an issue and should be prepared in the  backend before being passed to the HTML, saying that the code was already like it long before this PR and this hasn't made it any worse, in fact it is an improvement

PR #798, opened 2022-11-11
Allow events to be editable

insacuri:
# What and why?

The `mps`, `go_live`, `return_by` and `exercise_end` fields couldn't be edited once the exercise was set to a live state.  There might've been a reason for this in the past but that doesn't seem to be the case now.  This PR makes it so that the events are editable if the event hasn't elapsed yet.  As soon as the event has happened then the edit button disappears.

I also removed a bunch of packages from the Pipfile that we don't directly use.  These ones I've removed are transient dependencies (which pipenv would bring in anyway if something needed it, i.e., flask bringing in certifi in its requirements) 

# How to test?

# Trello


LJBabbage (approved):
Works as expected, set one up initially and then changed it and it obeyed the change

PR #802, opened 2022-12-06
Increase logging around RBA

LJBabbage:
# What and why?
This PR adds extra logging for IA to record who requests a RBA change
# How to test?
Do the steps on the story and make sure the correct log info are written out

N.B Happy to change the name of things, couldn't really think of anything at the time


SteveScorfield (approved):
Looks fine to me.

PR #804, opened 2022-12-08
Sample to ce page

insacuri:
# What and why?

This PR moves the sample functionality out of the 'sample and ci page' and puts it on the main page for a collection exercise.
A count of the collection instruments linked for an EQ exercise (or uploaded for a SEFT exercise) was added to the instruments part of the collection exercise screen.

Finally it tidies up some of the tests, combining multiple tests into a single case.  This is useful for us as each view test (that craetes a `test_client` is expensive to start, but cheap to tweak once you're running the test)

# How to test?

- Run both EQ and SEFT and BRES exercises and make sure they work
- Upload a large sample (1-3k in dev should do it) and make sure the sample load progress still works

# Trello

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -57,11 +57,11 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:0d9c601124e5a6ba9712dbc60d9c53c21e34f5f641fe83002317394311bdce14",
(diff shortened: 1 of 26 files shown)


SteveScorfield (approved):
Looks good to me

PR #806, opened 2022-12-13
Collection Exercise design improvements

mfcrocker:
# What and why?

# How to test?

# Trello


response_operations_ui/templates/create-collection-exercise.html
```diff
@@ -3,8 +3,10 @@
 {% from "components/metadata/_macro.njk" import onsMetadata %}
 {% from "components/fieldset/_macro.njk" import onsFieldset %}
 {% from "components/input/_macro.njk" import onsInput %}
+{% from "components/accordion/_macro.njk" import onsAccordion %}
 
 {% set page_title = 'Create collection exercise' %}
+{% set previous_period = request.args.get('previous_period') %}
```
insacuri, line 9: hmm, we've only done this in one place so I'm inclined so say to use template variables in the python code instead of doing request.args.get here... but then again I don't know if it's actually bad practice or not?  I think keeping as much code out of the templates is probably right?

insacuri (dismissed):
The acceptance tests fail with this change deployed

response_operations_ui/views/collection_exercise.py
```diff
@@ -588,7 +588,7 @@ def edit_collection_exercise_details(short_name, period):
 
 @collection_exercise_bp.route("/<survey_ref>/<short_name>/create-collection-exercise", methods=["GET"])
 @login_required
-def get_create_collection_exercise_form(survey_ref, short_name):
+def get_create_collection_exercise_form(survey_ref, short_name, previous_period=None):
```
insacuri, line 591: I think this carries on from my previous comment.  I think function variables are meant to be used for parts of the url (like survey_ref and short_name here).  If you want to get the previous_period from the url parameters, removing this and adding `request.args.get('previous_period')` here would be better

response_operations_ui/templates/create-collection-exercise.html
```diff
@@ -21,76 +23,77 @@ <h2 id="error-creating-ce-message" class="ons-u-fs-r--b">Error creating collecti
...
             </form>
+            <div class="ons-grid__col ons-col-3@xs">
+                {{
+                    onsAccordion({
+                        "id": "infoAccordion",
+                        "open": true,
+                        "itemsList": [
+                            {
+                                "title": "What is a Period ID",
+                                "content": "<p>The period ID is a two, four or six digit number that represents the month or year that a survey runs until.</p><p>'202212' or '2212' would represent December 2022.</p><p>'22' would represent the year 2022.</p>"
```
insacuri, line 91: The text in the accordion needs to be allowed to take up more slots in the grid as it's a bit narrow in comparison to the figma design.  Also, should it be open by default?  That's more of a question for Dina, but I think it's a bit ugly having accordions open by default... at that point you might as well always display it instead of bothering with an accordion

response_operations_ui/templates/create-collection-exercise.html
```diff
@@ -21,76 +23,77 @@ <h2 id="error-creating-ce-message" class="ons-u-fs-r--b">Error creating collecti
...
+                        onsPanel({
+                            "body": period_information,
+                        })
+                    }}
+                {% endif %}
+                <label class="ons-label ons-u-fs-m">Period</label>
+                {% if previous_period is not none %}
+                    <span class="ons-label__description">The most recent Period ID added was {{ previous_period }}</span>
+                {% else %}
+                    <span class="ons-label__description">The Period ID is a two, four or six digit number.</span>
```
insacuri, line 44: These textboxes need a little bit of padding I think as the text is right up against the input box

JacobHarrisonONS (approved):

PR #808, opened 2022-12-15
Sign out of user account on permission change

SteveScorfield:
# What and why?
There is a requirement for the user to be signed out when the users permission has been changed. This code change should accomplish this.

# How to test?
1. In a browser profile (Chrome) sign in with an account you want to change the permissions for (test user)
2. In one browser profile sign in with an admin user
3. Using the admin account, goto manage users and select the test user that was used to sign in
4. Change the permissions (either ticket a new permission or untick a permission)
5. Click save changed
6. Go back to the test user and either refresh the screen or click on another tab
7. The user should be signed out a redirected to the sign in page

# Trello
https://trello.com/c/frmUHnXZ/1798-spike-sign-out-of-user-account-on-permission-change-1-2-days


SteveScorfield:
[Spike: Sign out of user account on permission change (2 days)](https://trello.com/c/frmUHnXZ/1798-spike-sign-out-of-user-account-on-permission-change-2-days)

LJBabbage (commented):
As a spike it works, but we might want to be more specific in how we use all_sessions.get(user_in_session). I reckon that will dump out a large amount of data, without knowing (as it might be fine) what all the bits are, because of that doing a user_id in might not work consistent and the scope too wide. If this was a PR I would also say the function names needs to be connected to the job it is doing and not what caused it to happen, something like delete_user_from_redis_session is probably more appropriate. I would also be interested in how multiple sessions from the same account are handled, is it one session in redis and therefore you can escape that loop early.

SteveScorfield:
@LJBabbage I've improved the performance by using scan() instead of key() (it's more efficient). I've changed the get() to mget(), again this method increases performance.  I've replied to your comments below:

"As a spike it works, but we might want to be more specific in how we use all_sessions.get(user_in_session). I reckon that will dump out a large amount of data, without knowing (as it might be fine) what all the bits are, because of that doing a user_id in might not work consistently and the scope be to wide."

- mget() will only retrieve the values related to that particular session id. From my tests, user_id will always be present within the session data and I believe that this would work consistently. 

"If this was a PR I would also say the function names needs to be connected to the job it is doing and not what caused it to happen, something like delete_user_from_redis_session is probably more appropriate."

- I've updated this as you are correct it should be renamed to something more appropriate. I've updated it as part of the spike so that if we decide to go down this route, anyone picking it up can understand better.

"I would also be interested in how multiple sessions from the same account are handled, is it one session in redis and therefore you can escape that loop early."

- For every new browser session the user uses to login to rOps, a new session id is created in the DB along with the user id. So the code will keep looping through all the sessions until the end.

LJBabbage (approved):

response_operations_ui/views/admin/manage_user_accounts.py
```diff
@@ -281,6 +282,16 @@ def post_manage_account_groups(user_id):
...
 
 
+def delete_user_from_redis_session(user_id):
+    all_sessions = current_app.config["SESSION_REDIS"]
+    get_all_sessions_keys = all_sessions.scan(cursor=0, match="*")[1]
+
+    for user_in_session in get_all_sessions_keys:
+        user_session_information = str(all_sessions.mget(user_in_session))
+        if user_id in user_session_information:
+            all_sessions.delete(user_in_session)
```
insacuri, line 292: Can you put a comment in this loop to say why we're not exiting as soon as we find the user?  I didn't realise you could have multiple sessions at once and that's a fact that's easily forgotten by the team!

SteveScorfield:
[Sign out ROps user when permissions change](https://trello.com/c/iXX98VVD/2446-sign-out-rops-user-when-permissions-change)

LJBabbage (changes requested):
Works as expected. Did we want a test or is it something difficult to test?

PR #810, opened 2022-12-22
Remove eQ version selection from UI

LJBabbage:
# What and why?
This PR removes the option from the UI to select between eQ v2 and v3. I have also done a little bit of refactoring as we should try to avoid hidden fields being passed around
# How to test?
Make sure current collection exercises still work and creation of new ones also work as expected. Basically new ones should get a v3 in the db if eQ and nothing if a SEFT
# Trello


response_operations_ui/forms.py
```diff
@@ -274,8 +274,6 @@ def validate_day(form, field):
 class CreateCollectionExerciseDetailsForm(FlaskForm):
     user_description = StringField("user_description")
     period = IntegerField("period")
-    hidden_survey_id = HiddenField("hidden_survey_id")
```
LJBabbage, line 277: We should try to avoid doing this, passing around hidden fields which naturally can be changed on the UI isn't a good idea. No extra calls have been made as I have moved it in the backend

response_operations_ui/templates/create-collection-exercise.html
```diff
@@ -75,42 +75,6 @@ <h1 class="ons-u-fs-xl">Create collection exercise</h1>
                             }
                         })
                     }}
-                    {{
```
LJBabbage, line 78: As above, no longer required

response_operations_ui/views/collection_exercise.py
```diff
@@ -649,16 +642,16 @@ def create_collection_exercise(survey_ref, short_name):
...
             )
 
     logger.info(
         "Creating collection exercise for survey",
         survey=short_name,
         survey_ref=survey_ref,
-        survey_eq_version=survey_eq_version,
     )
 
+    survey_eq_version = "v3" if survey_details["surveyMode"] == "EQ" else ""
```
LJBabbage, line 653: This should only be temporary, but I can't remove this until the column has been removed from the db, which can't be done at present

tests/views/test_collection_exercise.py
```diff
@@ -1277,26 +1276,6 @@ def test_create_collection_exercise_success(self, mock_request):
         self.assertEqual(response.status_code, 200)
         self.assertIn("123456".encode(), response.data)
 
-    @requests_mock.mock()
-    def test_create_collection_exercise_failure(self, mock_request):
```
LJBabbage, line 1281: :thumbsdown: This test is not good, and is testing bad practise in our code. We should not just let a 500 from another service bubble up through this service as well and produce a 500, which is what it is testing. The generic assert request_history should also not be used like this as multiply mock_requests are in play and therefore renders a generic history count pointless. The good news is this doesn't actually cover a single line of code above the happy path and can be removed.

response_operations_ui/controllers/survey_controllers.py
```diff
@@ -276,7 +276,7 @@ def create_survey(survey_ref, short_name, long_name, legal_basis, survey_mode):
             {"name": "COLLECTION_INSTRUMENT", "classifierTypes": ["FORM_TYPE"]},
             {"name": "COMMUNICATION_TEMPLATE", "classifierTypes": ["LEGAL_BASIS", "REGION"]},
         ],
-        "eqVersion": "v2" if survey_mode != "SEFT" else "",
+        "eqVersion": "v3" if survey_mode != "SEFT" else "",
```
LJBabbage, line 279: Strictly speaking not part of this PR and in the future will be redundant, however new surveys should always be v3 not v2 now

insacuri: Good job catching this.  Could've lead to some interesting issues if we hadn't done this and the similar collection exercise change below >_<

tests/views/test_collection_exercise.py
```diff
@@ -1337,25 +1316,21 @@ def test_failed_create_ce_validation_period_exists(self, mock_request):
         )
 
     @requests_mock.mock()
-    def test_failed_create_ce_letters_fails_in_period_validation(self, mock_request):
+    def test_create_collection_exercise_invalid_period(self, mock_request):
```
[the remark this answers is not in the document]
insacuri: Improving the tests is a beast.... just doing this gets is a nudge in the right direction!

insacuri (approved):
Code looks good and the database values look sensible when running it in my environment

PR #812, opened 2023-01-17
Design system update

anwilkie:
# What and why?
Upgrading design system to latest version at time of writing (61.0.4). There was a small breaking change introduced in version 60 where certain components associated with `pageConfig` needed to be moved into `pageConfig.header`.

Unrelated to the design system upgrade, I also noticed that the monochrome ONS logo wasn't appearing at the bottom of the page like it does on frontstage. This was because invalid parameters were being put in the footer. I removed these, which causes the ONS logo to appear properly.

# How to test?
* Deploy to environment and run acceptance tests with `make helm`. Note that you will need to set the `appVersion` in `values.yaml` to `pr-108` so that it uses the acceptance test fix I made. If not, you'll receive errors about the tests not being able to find certain elements on the page.
* Test out response-ops navigation and UI. There might be a few instances of elements changing size ever so slightly, but nothing significant should have changed.

# Trello
[Card](https://trello.com/c/hIMCYrXm)


insacuri (dismissed):
This fails the acceptance tests on the change name step for...some reason that I can't figure out at a glance.

```
And they change their name to ONSname                                                                # uk.gov.onsdigital.rops.steps.UserAccountManagementSteps.changeName(java.lang.String)
      org.openqa.selenium.NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"#first_name"}
```

I took a look and those ids are there on each page as expected, so maybe something is different about how it clicks input boxes now or something?

insacuri:
Besides the test failure, everything else seems to work just fine and this can be approved once that's sorted!

JacobHarrisonONS (approved):
Acceptance tests are passing fine. Response Ops looks good to me, see nothing that looks out of place

PR #813, opened 2023-01-18
Mock Redis in response ops unit tests

SteveScorfield:
# What and why?
The unit tests within rOps would only run locally when the local machine was pointing to a redis session in GCP. Otherwise, the majority of tests would fail. Fakeredis was added that creates a fake server instance that the tests can run against. This fake server is only ran if the configuration is test or debug.

# How to test?
Run the unit tests without the local machine without a redis session being pointed to.

# Trello
https://trello.com/c/jZqS2whT/2103-mock-redis-in-response-ops-unit-tests


SteveScorfield:
[Mock Redis in response ops unit tests](https://trello.com/c/jZqS2whT/2103-mock-redis-in-response-ops-unit-tests)

Makefile
```diff
@@ -13,14 +13,14 @@ load-design-system-templates:
 	pipenv run ./scripts/load_templates.sh $(DESIGN_SYSTEM_VERSION)
 
 lint:
-	pipenv check ./response_operations_ui ./tests -i 51499
+	pipenv check ./response_operations_ui ./tests -i 51499 -i 51668
```
LJBabbage, line 16: 51668? I am only getting one now, the new one for certifi which can be fixed with a bump, see frontstage. It might be your version of pipenv

LJBabbage (commented):
Works as expected, I had no redis and the tests still worked. The only though is if I did have redis deployed through docker, brew or forwarded would fake redis take precedent, as it sits on the same port. I actually tried this with docker and it did work, but I didn't dig into the detail of which one it was using

LJBabbage (approved):

PR #826, opened 2023-02-02
Add page for no permission

insacuri:
# What and why?

# How to test?

# Trello

### DIFF
--- a/response_operations_ui/views/errors.py
+++ b/response_operations_ui/views/errors.py
@@ -47,8 +47,12 @@ def handle_authentication_error(error):
 
 
 @error_bp.app_errorhandler(NoPermissionError)
-def handle_no_permission_error(_):
-    logger.error("User attempted to access page they don't have permission to view", user_id=session["user_id"])
+def handle_no_permission_error(error):
+    logger.error(
+        "User attempted to access page they don't have permission to view",
+        user_id=session["user_id"],
+        permission=error.missing_permission,
+    )
     return render_template("errors/no-permission-error.html")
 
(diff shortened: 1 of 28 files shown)


response_operations_ui/views/errors.py
```diff
@@ -45,6 +46,12 @@ def handle_authentication_error(error):
     return redirect(url_for("sign_in_bp.sign_in"))
 
 
+@error_bp.app_errorhandler(NoPermissionError)
```
LJBabbage, line 49: :thumbsup: We need to do this more often

response_operations_ui/views/messages.py
```diff
@@ -53,8 +56,8 @@
 @login_required
 def create_message():
     if not user_has_permission("messages.edit"):
-        logger.exception("No message edit role")
-        abort(401)
+        logger.error("No message edit role")
```
LJBabbage, line 59: Without any biddings (i.e user_id) and the volume of log message we get do these add anything? With it raising to NoPermissionError which also has an error do we need them?

LJBabbage (approved):
Created a new secure message with admin account, when I tried to access the close page with uaa_reporting_units_edit_user@ons.gov.uk it gave me the new message. Minor comment about logging

response_operations_ui/views/messages.py
insacuri: Agreed.  Just changed it so we add the missing permission to the exception, and the NoPermissionError handler logs both the users id and the missing permission

PR #830, opened 2023-02-07
Add missing NoPermissionError to remaining pages

insacuri:
# What and why?

In https://github.com/ONSdigital/response-operations-ui/pull/826 there was a 'no access' page added for users accessing pages they didn't have permission for.  Because we had 2 different ways to check permissions, only half of the permission locked pages got this.  With this PR, all of the pages now function in the same way and all of them check permissions in the same way too for consistency.

# How to test?

# Trello

### DIFF
--- a/response_operations_ui/views/admin/manage_user_accounts.py
+++ b/response_operations_ui/views/admin/manage_user_accounts.py
@@ -14,7 +14,8 @@ from flask_paginate import Pagination
 from requests import HTTPError
 from structlog import wrap_logger
 
-from response_operations_ui.common import token_decoder, uaa
+from response_operations_ui.common.token_decoder import generate_token
+from response_operations_ui.common.uaa import verify_permission
 from response_operations_ui.controllers.notify_controller import NotifyController
 from response_operations_ui.controllers.uaa_controller import (
     add_group_membership,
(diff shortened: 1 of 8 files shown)


response_operations_ui/views/admin/manage_user_accounts.py
```diff
@@ -14,7 +14,7 @@
 from requests import HTTPError
 from structlog import wrap_logger
 
-from response_operations_ui.common import token_decoder
+from response_operations_ui.common import token_decoder, uaa
```
LJBabbage, line 17: Do we need to import all of uaa? do we not just need the function like the others. I think token_decoder might be the same

LJBabbage (commented):
The ones changed look good and I tested a couple with the correct result, however I am unsure that we have all permissions correct, the one that caught my eye is updating case statuses (there were a couple in respondents.py I also saw) However I reckon that is out of scope for this card, similar returning 200 might be considered the wrong http response code. Only one minor comment in scope

response_operations_ui/views/admin/manage_user_accounts.py
insacuri: Agreed.  Changed it.

LJBabbage (approved):

PR #833, opened 2023-02-09
Update CE page to allow multimode

LJBabbage:
# What and why?
This PR enables multimode in CE. I have taken the opportunity to remove the logic from the frontend and move it into the code for ce-ci-section.html where realistically it should be. Although I would have liked to moved build_collection_exercise_details and the new ci table context builder I have left where it is at the moment, it shouldn't really be in the view. I have made the code more explicit (eq/seft function called as such) and efficient (i.e not calling collection instrument on every pass; moving eq_ci_selectors; refactoring test etc) where it has been on my route. I have also moved redundancy for the same reason 

N.B 

The test change tab is minimised due to the amount of changes, rather than deleted as it might appear on the files changed section

Ideally ce-collection-instrument-section.html should be refactored as well, but time restrictions meant I have left mainly as if and changed to be controlled from the URL. We should aim for a clear logic/display separation where possible. 

# How to test?
Whether on GCP or local you will need this branch as well https://github.com/ONSdigital/ras-collection-instrument/pull/295 
Initially SEFT/EQ should be tested separately to make sure they still work as normal. Once that works you need to add in a EQ_AND_SEFT survey, this can be done by updating the survey_mode (survey scheme) in the DB or setting the env MULTI_MODE_ENABLED=True allowing the creation/edit of a new EQ_AND_SEFT survey on the UI. Once that is done create a CE and make sure both eQ and SEFT CI options are available and also work as expected.

N.B This doesn't match the image on the card as I haven't done the total CI count, I missed it initially and decided I needed some clarification on when it should appear, with that in mind and down to the blocking nature of this card I have split it out 

# Trello


response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -204,29 +204,6 @@ def execute_collection_exercise(collection_exercise_id):
     logger.info("Successfully began execution of collection exercise", collection_exercise_id=collection_exercise_id)
 
 
-def update_collection_exercise_eq_version(collection_exercise_id, eq_version):
```
LJBabbage, line 207: Had to refactor higher up and realised this is now redundant

response_operations_ui/views/collection_exercise.py
```diff
@@ -61,10 +60,14 @@
...
 
+CI_TABLE_TEXT = {
+    "EQ": {"no_instrument": "Add", "restricted": "View", "has_permission": "Select or Add"},
+    "SEFT": {"no_instrument": "Upload", "restricted": "View", "has_permission": "View or Upload"},
+}
 
-def build_collection_exercise_details(short_name: str, period: str, include_ci=True) -> dict:
-    """
 
+def build_collection_exercise_details(short_name: str, period: str, include_ci: bool = False) -> dict:
```
LJBabbage, line 69: include_ci=True was incorrectly being used, meaning most of the calls made to build_collection_exercise_details were processing cis unnecessarily

response_operations_ui/views/collection_exercise.py
```diff
@@ -92,26 +96,30 @@ def build_collection_exercise_details(short_name: str, period: str, include_ci=T
...
 
     if include_ci:
-        collection_instruments = collection_instrument_controllers.get_collection_instruments_by_classifier(
-            collection_exercise_id=collection_exercise_id, survey_id=survey_id
+        exercise_dict["collection_instruments"] = _build_collection_instruments_details(
+            collection_exercise_id, survey_id
         )
+    return exercise_dict
 
-        eq_ci_selectors = collection_instrument_controllers.get_collection_instruments_by_classifier(
```
LJBabbage, line 99: Used in one place, but call on everything including SEFT which is clearly wrong. Moved higher

response_operations_ui/views/collection_exercise.py
```diff
@@ -230,14 +267,12 @@ def post_collection_exercise(short_name, period):
...
-    elif "unselect-ci" in request.form:
-        return _unselect_collection_instrument(short_name, period)
-    if "eq-version" in request.form:
-        return _update_eq_version(short_name, period)
+    if "load-seft-ci" in request.form:
+        return _upload_seft_collection_instrument(short_name, period)
+    elif "select-eq-ci" in request.form:
+        return _select_eq_collection_instrument(short_name, period)
+    elif "unselect-eq-ci" in request.form:
+        return _unselect_eq_collection_instrument(short_name, period)
```
LJBabbage, line 275: Moved to more explicit to avoid issue seen above

response_operations_ui/views/collection_exercise.py
```diff
@@ -308,42 +343,18 @@ def _select_collection_instrument(short_name, period):
...
             "collection_exercise_bp.get_view_sample_ci",
             short_name=short_name,
             period=period,
             success_panel=success_panel,
+            survey_mode="EQ",
         )
     )
 
 
-def _update_eq_version(short_name, period):
```
LJBabbage, line 322: Redundant now

response_operations_ui/views/collection_exercise.py
```diff
@@ -230,14 +267,12 @@ def post_collection_exercise(short_name, period):
 @collection_exercise_bp.route("/<short_name>/<period>/view-sample-ci", methods=["POST"])
 @login_required
 def post_sample_ci(short_name, period):
-    if "load-ci" in request.form:
-        return _upload_collection_instrument(short_name, period)
-    elif "select-ci" in request.form:
-        return _select_collection_instrument(short_name, period)
-    elif "unselect-ci" in request.form:
-        return _unselect_collection_instrument(short_name, period)
-    if "eq-version" in request.form:
```
LJBabbage, line 239: redundant now

response_operations_ui/views/collection_exercise.py
```diff
@@ -1010,54 +1025,3 @@ def _delete_seft_collection_instrument(short_name, period):
             success_panel=success_panel,
         )
     )
-
-
-def _upload_seft_collection_instrument(short_name, period):
```
LJBabbage, line 1015: Most random prize of the year goes to this one, we have the identical function already, called _upload_collection_instrument. I even used a string comparator on it. The only difference is in the url_for, the params are in a slightly different order, but they are name so that becomes irrelevant

tests/views/test_collection_exercise.py
```diff
@@ -222,14 +199,6 @@ def setup_data(self):
             "ref_period_start": "2017-05-15T00:00:00Z",
             "ref_period_end": "2017-05-15T00:00:00Z",
         }
-        self.survey = {
```
LJBabbage, line 225: We already has this once

tests/views/test_collection_exercise.py
```diff
@@ -353,12 +351,6 @@ def setup_data(self):
                 "surveyId": survey_id,
             }
         ]
-        self.linked_sample = {
```
LJBabbage, line 356: Not used

tests/views/test_collection_exercise.py
```diff
@@ -513,19 +577,10 @@ def test_collection_exercise_view_event_statuses(self, mock_request):
...
-    @requests_mock.mock()
     @patch("response_operations_ui.views.collection_exercise.build_collection_exercise_details")
-    def test_collection_exercise_view_seft_exercise_states(self, mock_request, mock_details):
-        mock_request.get(url_get_survey_by_short_name, json=self.seft_survey)
-        mock_request.get(url_ces_by_survey, json=self.collection_exercises)
-        mock_request.get(url_ce_by_id, json=collection_exercise_details["collection_exercise"])
-        mock_request.get(url_get_collection_exercise_events, json=self.collection_exercise_events)
-        mock_request.get(url_link_sample, json=[sample_summary_id])
-        mock_request.get(url_get_sample_summary, json=self.sample_summary)
-        mock_request.post(url_collection_instrument, status_code=201)
```
LJBabbage, line 525: Over mocked

tests/views/test_collection_exercise.py
```diff
@@ -928,22 +1046,6 @@ def test_upload_sample_csv_too_few_columns(self, mock_request, mock_details):
         self.assertIn("Sample summary".encode(), response.data)
         self.assertIn("Too few columns in CSV file".encode(), response.data)
 
-    @requests_mock.mock()
-    @patch("response_operations_ui.views.collection_exercise.build_collection_exercise_details")
-    def test_eq_version_change_success(self, mock_request, mock_details):
```
LJBabbage, line 933: redundant

tests/views/test_collection_exercise.py
```diff
@@ -1019,9 +1121,8 @@ def test_post_ready_for_live_service_fail(self, mock_request):
         self.assertEqual(len(request_history), 2)
         self.assertEqual(response.status_code, 500)
 
-    @requests_mock.mock()
     @patch("response_operations_ui.views.collection_exercise.build_collection_exercise_details")
-    def test_get_processing(self, mock_request, mock_details):
```
LJBabbage, line 1024: unused

tests/views/test_collection_exercise.py
```diff
@@ -1031,9 +1132,8 @@ def test_get_processing(self, mock_request, mock_details):
         self.assertEqual(response.status_code, 200)
         self.assertIn("Processing collection exercise".encode(), response.data)
 
-    @requests_mock.mock()
     @patch("response_operations_ui.views.collection_exercise.build_collection_exercise_details")
-    def test_failed_execution(self, mock_request, mock_details):
```
LJBabbage, line 1036: Unused

tests/views/test_collection_exercise.py
```diff
@@ -535,39 +591,101 @@ def test_collection_exercise_view_seft_exercise_states(self, mock_request, mock_
...
+        # Then the collection exercise has 1 key (SEFT) and 2 values in the collection instrument
+        expected_output = {"SEFT": eq_and_seft}
+
+        self.assertEquals(expected_output, exercise_dict["collection_instruments"])
+        self.assertEquals(len(exercise_dict["collection_instruments"]["SEFT"]), 2)
+        self.assertEquals(exercise_dict["collection_instruments"]["SEFT"][0]["id"], collection_instrument_id)
+        self.assertEquals(exercise_dict["collection_instruments"]["SEFT"][1]["id"], collection_instrument_id_2)
+
+    @requests_mock.mock()
+    def test_collection_exercise_details_eq_and_seft_collection_instruments(self, mock_request):
```
LJBabbage, line 666: You could argue that some of these tests aren't needed and we could just test the endpoint, but that isn't particularly reliable, for things like count we have nothing really to assert against bar a td class which isn't great, we should always aim to assert dict contents and not response.data, unless naturally test the display

insacuri: Sounds good to me.  I was going to say this should probably be in `controllers/test_collection_exercise_controllers.py` but then realised this function lives in the view so I'd say this test is in the right place

insacuri (changes requested):
A few small bits but considering you changed so much, it looks good!

insacuri (approved):
LGTM

PR #835, opened 2023-02-14
Split the CE dates edit feature into 2

mfcrocker:
# What and why?
We have two links that go to the same page to edit the Period ID and Period as shown to user. This splits it into two edit screens.

# How to test?

# Trello
https://trello.com/c/THv1aQuZ/2473-split-out-the-edit-id-and-edit-dates-functionality-on-the-collection-exercise-screen


JacobHarrisonONS (approved):
Looks good. tested with EQ, SEFT, BRES and EQANDSEFT tags

PR #844, opened 2023-02-23
Extend collection instrument error message and add tests

LJBabbage:
# What and why?
This PR allows the message from CI upload to be displayed, it also adds a missing test
# How to test?
Test in conjunction with https://github.com/ONSdigital/ras-collection-instrument/pull/297
# Trello

### DIFF
--- a/response_operations_ui/controllers/collection_instrument_controllers.py
+++ b/response_operations_ui/controllers/collection_instrument_controllers.py
@@ -44,14 +44,16 @@ def upload_collection_instrument(collection_exercise_id, file, form_type=None):
             form_type=form_type,
             status=response.status_code,
         )
-        return False
+        if response.headers["Content-Type"] == "application/json":
+            return False, response.json().get("errors")[0]
+        return False, None
 
     logger.info(
         "Successfully uploaded collection instrument",
         collection_exercise_id=collection_exercise_id,
         form_type=form_type,
     )
-    return True
+    return True, None
 
 
 def upload_ru_specific_collection_instrument(collection_exercise_id, file, ru_ref):
--- a/tests/views/test_collection_exercise.py
+++ b/tests/views/test_collection_exercise.py
@@ -840,22 +840,6 @@ class TestCollectionExercise(ViewTestCase):
         self.assertEqual(response.status_code, 200)
         self.assertIn("test_collection_instrument.xlxs".encode(), response.data)
 
-    @requests_mock.mock()
-    @patch("response_operations_ui.views.collection_exercise.build_collection_exercise_details")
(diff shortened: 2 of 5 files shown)


tests/views/test_collection_exercise.py
```diff
@@ -840,22 +840,6 @@ def test_view_seft_collection_instrument_after_upload(self, mock_request, mock_d
         self.assertEqual(response.status_code, 200)
         self.assertIn("test_collection_instrument.xlxs".encode(), response.data)
 
-    @requests_mock.mock()
-    @patch("response_operations_ui.views.collection_exercise.build_collection_exercise_details")
```
LJBabbage, line 844: This is a duplication

tests/views/test_collection_exercise.py
```diff
@@ -1852,11 +1836,35 @@ def test_seft_upload_collection_instrument_supports_xls(self, mock_request, mock
...
+        response = self.client.post(
+            f"/surveys/{short_name}/{period}/load-collection-instruments", data=post_data, follow_redirects=True
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn("Error: Failed to upload collection instrument".encode(), response.data)
+        self.assertIn("Error message passed".encode(), response.data)
+
+    @requests_mock.mock()
+    @patch("response_operations_ui.controllers.collection_instrument_controllers.upload_collection_instrument")
```
LJBabbage, line 1860: The hierarchy in this test is not quite right. The responsibility of testing the different scenarios of the upload should be with test_collection_instrument_controller.py (which has been updated above) This test shouldn't stretch down that far as the combinations as you go further down into the code will only exponentially grow, we should only be concerned with the returns that come from upload_collection_instrument. With that in mind there was actually a missing test, which was added above

response_operations_ui/controllers/collection_instrument_controllers.py
```diff
@@ -44,14 +44,16 @@ def upload_collection_instrument(collection_exercise_id, file, form_type=None):
             form_type=form_type,
             status=response.status_code,
         )
-        return False
+        if response.headers["Content-Type"] == "application/json":
```
insacuri, line 47: Is this code like this to handle 400 vs 500 errors?  If so, can you put a comment next to it as it's definitely different to how we normally do it.

LJBabbage: I pulled it from the one below upload_ru_specific_collection_instrument, its very debatable if we need it, but we did have a test that didn't provide json, perhaps its that which needs updating

insacuri (approved):
LGTM

PR #845, opened 2023-02-24
Remove intermediary seft upload page

matthew-robinson-ons:
# What and why?
When clicking on the 'Upload' or 'View or upload' CI SEFT on the CE information page, an intermediary summary page is shown, this PR removes that page and takes the user directly to the load collection instrument page.
# How to test?
For a SEFT and an EQ and SEFT go through the seft upload journey on a collection exercise to ensure the intermediary page is no longer present and the user is taken directly to the load-collection-instruments page. The eQ journey should not have changed. 
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-682

### DIFF
--- a/response_operations_ui/templates/collection_exercise/ce-ci-section.html
+++ b/response_operations_ui/templates/collection_exercise/ce-ci-section.html
@@ -61,7 +61,7 @@
                     "tdClasses": "ons-u-ta-centre"
                   }, 
                   {
-                    "value": '<a id="seft-upload-link" href="' + url_for('collection_exercise_bp.get_seft_collection_instrument', period=ce.exerciseRef, short_name=survey.shortName.replace(' ', '')) + '">' + ci_type.link_text + '</a>',
+                    "value": '<a href="' + url_for('collection_exercise_bp.get_seft_collection_instrument', period=ce.exerciseRef, short_name=survey.shortName.replace(' ', '')) + '" id="seft-upload-link">' + ci_type.link_text + '</a>',
                     "tdClasses": "ons-u-ta-right"
                   }
                 ] 
--- a/response_operations_ui/templates/collection_exercise/ce-ci-section.html
+++ b/response_operations_ui/templates/collection_exercise/ce-ci-section.html
@@ -28,9 +28,8 @@
         %}
     {% do collectionInstrumentTableData | setAttribute("ths", tblHeaders) %}
   {% endif %}
-  
     {% for ci_type in ci_table_context["ci_details"] %}
-      {% if survey["surveyMode"] == "EQ" %}
+      {% if ci_type.type == "eq" %}
         {% do collectionInstrumentTableDataRows.append(
             {
                 "tds": [
@@ -50,7 +49,7 @@
         )
         %}
       {% endif %}
-      {% if survey["surveyMode"] == "SEFT" %}
+      {% if ci_type.type == "seft" %}
         {% do collectionInstrumentTableDataRows.append(
             {
                 "tds": [
--- a/response_operations_ui/templates/collection_exercise/ce-ci-section.html
+++ b/response_operations_ui/templates/collection_exercise/ce-ci-section.html
@@ -30,24 +30,46 @@
   {% endif %}
   
     {% for ci_type in ci_table_context["ci_details"] %}
-      {% do collectionInstrumentTableDataRows.append(
-          {
-              "tds": [
-                {
-                  "value": ci_type.title
-                },
-                {
-                  "value": ci_type.count,
-                  "tdClasses": "ons-u-ta-centre"
-                }, 
-                {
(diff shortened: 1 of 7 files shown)


response_operations_ui/templates/collection_exercise/ce-ci-section.html
```diff
@@ -28,26 +28,25 @@
...
+                "tds": [
+                  {
+                    "value": ci_type.title
+                  },
+                  {
+                    "value": ci_type.count,
+                    "tdClasses": "ons-u-ta-centre"
+                  }, 
+                  {
+                    "value": '<a href="' + url_for('collection_exercise_bp.get_seft_collection_instrument', period=ce.exerciseRef, short_name=survey.shortName.replace(' ', '')) + '" id="seft-upload-link">' + ci_type.link_text + '</a>' if ci_type.type == "seft" else '<a href="' + ci_type.url + '" id="view-add-upload-ci-' + ci_type.type + '">' + ci_type.link_text + '</a>',
```
matthew-robinson-ons, line 43: Tried a few options to achieve this. Don't like how long the line of code is but seemed to involve the least code reuse.

response_operations_ui/templates/collection_exercise/ce-view-sample-ci.html
```diff
@@ -16,11 +16,7 @@
     {% include 'partials/flashed-messages.html' %}
     <div class="ons-grid ons-grid--gutterless">
         {% include 'partials/error-box.html' %}
-        {% if survey.surveyMode == 'EQ' %}
-          <h1 name="page-ce-title">Select collection instruments</h1>
-        {% else %}
-          <h1 name="page-ce-title">Load collection instruments</h1>
-        {% endif %}
+        <h1 name="page-ce-title">Select collection instruments</h1>
```
matthew-robinson-ons, line 19: This page no longer handles SEFT

response_operations_ui/templates/collection_exercise/ce-ci-section.html
LJBabbage, line 43: This could and probably should be done in the backend. The href is only a url, id and link text, you just need to do the logic in the backend and pass them through (url and link_text is already passed) In fact you can get rid of type from the dict if you do, or just use the same id for seft (view-add-upload-ci-)

response_operations_ui/templates/collection_exercise/ce-collection-instrument-section.html
```diff
@@ -1,55 +1,5 @@
...
-                                    "tdClasses": "ons-u-ta-right"
-                                }
-                            ]
-                        }
-                    ]
-                })
-            }}
-        {% endif %}
-      
         {% if request.args.get('survey_mode') == "EQ" %}
```
LJBabbage, line 3: If it is now just eQ, can't this if be removed

LJBabbage: Taking it a step further back can't you removed the url arg survey_mode as well? https://github.com/ONSdigital/response-operations-ui/blob/main/response_operations_ui/views/collection_exercise.py#L198

LJBabbage: And here on the route back https://github.com/ONSdigital/response-operations-ui/blob/main/response_operations_ui/views/collection_exercise.py#L353

LJBabbage: Me personally I would then change the template name to be explicit and name it as if is just for eQ.

LJBabbage (changes requested):
A few suggestions, I was also expecting to see a few acceptance test changes, worrying if removing a page doesn't break at least one.

response_operations_ui/templates/collection_exercise/ce-collection-instrument-section.html
matthew-robinson-ons: Code updated and template also renamed

response_operations_ui/templates/collection_exercise/ce-ci-section.html
matthew-robinson-ons: Have moved this logic to the backend. Used the same id for seft and eq however acceptance tests cannot find the id when run as a seft so working through that now.

LJBabbage (approved):
Nice one, works as expected and code is really easy to follow, I tried all survey mode combinations (eQ, SEFT and EQ_AND_SEFT). Saw a very minor change, but no need for another review pass after the change, so approving. Obviously wait for acceptance tests before merging. You might also consider taking out the Pipfile.lock, as this could feasibly break something else in the code base unless there was an intended change in it

PR #846, opened 2023-02-24
Fix broken details element

insacuri:
# What and why?

The 'what is the period ID' part of the create-collection-exercise page was broken and just showing an arrow with no text.  This PR fixes it so it shows the title and help text

# How to test?

# Trello


matthew-robinson-ons (approved):
Looks good

PR #849, opened 2023-03-03
Add delete survey page

insacuri:
# What and why?

Now that there's the survey delete permission and survey delete endpoint setup, this PR adds the UI elements so the survey can be deleted if there are no exercises for the survey AND no secure messages.

TODO:
- Query secure message to ensure there are no messages for the survey (so those messages don't become orphaned)
- Fix tests with updated mock requests in edit pages

# How to test?

# Trello

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -236,32 +236,32 @@
         },
         "cryptography": {
             "hashes": [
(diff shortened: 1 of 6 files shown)


pricem14pc:
We wont do this via the UI

PR #861, opened 2023-04-24
Remove flask-assets

mfcrocker:
# What and why?
`flask-assets` hasn't been updated in years, and is starting to give warnings about methods it uses being deprecated. We're not actually using it for anything, so I've pulled it out.

# How to test?
Spin up and have a poke around. You shouldn't see any CSS errors.

# Trello
RAS-748


SteveScorfield (approved):
Looks good to me

PR #865, opened 2023-05-19
Adding ref period dates validation before 'Set ready for live' appears

arroyoAle:
# What and why?
This PR adds a requirement for both reference period dates to be added to any multimode survey to ensure that when respondent open an eQ version of the collection exercise, frontstage is able to generate a correct eQ url

NOTE: I'm not sure if this fixes it completely but it prevents the business from setting a multimode survey to live without all the required dates. I'm not sure what collection exercise does differently between the EQ, SEFT, and EQ_AND_SEFT types.

# How to test?

- create a multi mode survey
- try to create a collection exercise for this survey without adding reference period dates

# Jira
[RAS-764](https://jira.ons.gov.uk/browse/RAS-764)


LJBabbage (changes requested):
On a bug like this we should always add a test. I also find this code confusing unfortunately, why would we allow the live button to appear on failed validation? Whether that is correct or not this can be rationalised and both 'eQ' and 'eQ and SEFT' can be catered by the same if statement as using 'in' it doesn't matter if eQ can never get to FAILEDVALIDATION but EQ_AND_SEFT can

response_operations_ui/views/collection_exercise.py
```diff
@@ -133,8 +133,14 @@ def view_collection_exercise(short_name, period):
             and "ref_period_start" in ce_details["events"]
             and "ref_period_end" in ce_details["events"]
         )
+    elif survey_mode == "EQ_AND_SEFT":
```
LJBabbage, line 136: This needs to be rationalised with EQ 
```
if survey_mode in ("EQ", "EQ_AND_SEFT"):
        show_set_live_button = (
            ce_state in "READY_FOR_REVIEW"
            and "ref_period_start" in ce_details["events"]
            and "ref_period_end" in ce_details["events"]
        )
else:
        show_set_live_button = ce_state in "READY_FOR_REVIEW"

File: tests/views/test_collection_exercise.py [hunk 2]
```diff
@@ -519,6 +519,120 @@ def test_collection_exercise_view_eq_both_ref_dates(self, mock_request):
...
+        mock_request.get(url_get_collection_exercise_events, json=collex_events)
+        mock_request.get(url_link_sample, json=[sample_summary_id])
+        mock_request.get(url_get_sample_summary, json=sample_summary)
+        mock_request.get(f"{url_get_collection_instrument}?{ci_search_string}", json=eq_ci, complete_qs=True)
+        mock_request.get(
+            f"{url_get_collection_instrument}?{ci_type_search_string_eq}", json=eq_ci_selectors, complete_qs=True
+        )
+
+    @requests_mock.mock()
+    def test_collection_exercise_view_eq_and_seft_non_ref_date(self, mock_request):
```
LJBabbage, line 546: If you rationalise as above you don't need to do all these tests, just 1 for EQ_AND_SEFT, and it can be any (with, without, both )because realistically you are just testing the If in bit

LJBabbage (changes requested):

LJBabbage (approved):

PR #875, opened 2023-07-28
RAS-837 Rops Security Alerts

LJBabbage:
# What and why?

# How to test?

# Trello


Pipfile
```diff
@@ -46,7 +46,7 @@ email_validator = "*"
 [dev-packages]
 black = "*"
 flake8 = "*"
-djlint = "*"
+djlint = "==1.25.0"
```
LJBabbage, line 49: djlint is causing issues when the lock file is regenerated, it concerns html linting. It's not related to this story and should be fixed properly, so I will raise another card and pin it for the moment

SteveScorfield: Makes sense

SteveScorfield (approved):
Looks good to me

PR #878, opened 2023-08-07
update certifi

anwilkie:
# What and why?
Updating certifi to resolve a dependabot alert

# How to test?
Deploy and run acceptance tests

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-852)


LJBabbage (changes requested):
This one is not needed and not updating certifi, it was done as the card needing 14 security updates. Also the reason why there is no dependabot PR

PR #880, opened 2023-08-30
change version to v65

anwilkie:
# What and why?
Updating design system to version 65. In version 64, the `ariaLabel` variable for the `<a>` component was removed, so I removed instances of it from response-ops accordingly.

# How to test?
Deploy, run acceptance tests, and navigate around service to make sure the pages look correct.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-863)


SteveScorfield (approved):
Looks good to me

PR #882, opened 2023-09-05
fix ons-icon in html files

anwilkie:
# What and why?
I missed something in the last design system upgrade where the icons that appear near links (such as on the response chasing link on the collection exercise overview page) appeared enormous. This turned out to be because the design system v65 renamed the `ons-svg-icon` to `ons-icon` (this wasn't mentioned in the breaking changes!).

Also, the design system upgrade messed with the appearance of the tab component due to the fact that the way we use it is not in accordance with the design system's ethos. The way the tabs are currently created is by a hacky HYML code block, which isn't how it's meant to be done, but it allows us to have the appropriate functionality. To continue supporting our implementation of tabs, the .css file has had some additional config in order to maintain its current appearance.

# How to test?
Deploy, run acceptance tests, go to response-ops and navigate to one of the collection exercise overview and make sure that the icons near links (on response chasing, for example) are not way larger than they should be. Also, make sure that the tab component (seen when looking at RUs) appears correctly, with a line going across the page underneath the tabs.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-863)


matthew-robinson-ons (approved):
Looks good, icons are normal size and code is fine.

arroyoAle (approved):
Everything looks alright from a click around rOps

PR #895, opened 2023-10-27
Ras 303 change sample summary call

anwilkie:
# What and why?
There was a redundant call to get the sample summaries for a group of collection exercises that needed removal.

# How to test?
Deploy and run acceptance tests. Also, make sure you can access the surveys page, click on colllection exercises, and view their info

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-303)


tests/controllers/test_collection_exercise_controller.py
```diff
@@ -122,23 +118,10 @@ def test_get_collection_exercises_with_events_and_samples_by_survey_id(self, moc
                 ],
             }
         ]
-        sample_summary = {
```
JacobHarrisonONS, line 125: Sample summary has been moved into collection exercise in the form of 
Integer sampleSize;
List<SampleLink> sampleLinks;
Do we want to check that is the case where valid in a test?

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -440,11 +439,6 @@ def get_collection_exercises_with_samples_by_survey_id(survey_id):
 
     ce_list = get_collection_exercises_by_survey(survey_id)
 
-    for ce in ce_list:
-        sample_summary_id = get_linked_sample_summary_id(ce["id"])
-        if sample_summary_id:
-            ce["sample_summary"] = get_sample_summary(sample_summary_id)
-
```
JacobHarrisonONS, line 447: Double check for other calls to get_linked_sample_summary_id and get_sample_summary. There are calls present in build_collection_exercise_details() in collection_exercise.py are they still required ?

JacobHarrisonONS (dismissed):
Just a few things to clarify and some test improvement.

arroyoAle (changes requested):
I agree with Jacob's comments, and when you port forward, run the acceptance tests and go to `localhost:8085/surveys/QBS`, there is no value shown in the "sample size" column, which makes me guess that its either not getting passed from collection exercise, or, we need to change the way its referenced in the html when its displayed (I think its just a change to the key id

arroyoAle:
What it looks like: 

Screenshot from 2023-10-31 11-13-37


What its meant to look like:

Screenshot from 2023-10-31 11-15-27

tests/controllers/test_collection_exercise_controller.py
anwilkie: The adding of the sample size and sample links to the collection exercise seems to be done on the collection exercise service side of things and is tested out quite thoroughly there. I don't think these two variables are ever referenced to on r-ops outside of HTML files.

response_operations_ui/controllers/collection_exercise_controllers.py
anwilkie: I have tried removing the other usage of that function, but it results in a 500 error and several test failures. It doesn't seem like something that can be removed without significantly refactoring the code.

anwilkie:
> What it looks like:
> 
> Screenshot from 2023-10-31 11-13-37
> 
> What its meant to look like:
> 
> Screenshot from 2023-10-31 11-15-27

I've fixed this issue.

response_operations_ui/controllers/collection_exercise_controllers.py
arroyoAle: Looking into it I don't think we can remove any of the other calls to sample as they collect the ingest date time rather than the sample size and links

tests/controllers/test_collection_exercise_controller.py
arroyoAle: I think we still need to add the values to the collection exercise we create in the tests


```
collection_exercises = [
  {
      "id": collection_exercise_id,
      "name": "201601",
      "scheduledExecutionDateTime": "2017-05-15T00:00:00Z",
      "state": "PUBLISHED",
      "exerciseRef": "000000",
      "events": [
          {
              "collectionExerciseId": "14fb3e68-4dca-46db-bf49-04b84e07e77c",
              "eventStatus": "PROCESSED",
              "id": "b4a36392-a21f-485b-9dc4-d151a8fcd565",
              "tag": "mps",
              "timestamp": "2018-03-16T00:00:00.000Z",
          }
      ],
     "sampleSize": "8",
     "sampleLink": "UUID"
  }
]
```

Kind of like this, since this is what collection exercise would be returning, so we can test on the proper data.

The issue now is that all of the test data we have in the rOps tests is in an old format as it expects, but I'm not sure how we want to deal with that, probably question for the larger team/tech session

pricem14pc:
@anwilkie As [This method](https://github.com/ONSdigital/response-operations-ui/pull/895/files#diff-4310204905351551516d7296e7015779cfedf1a848e13f44010c7449e0b0a392R437) no longer has the responsibility of retrieving sample summary data, it would usually be renamed more appropriately `get_collection_exercises_by_survey_id`. Knowing this, have a look at what additional refactoring you could do to make your changes more efficient and remove redundancy.

anwilkie:
> @anwilkie As [This method](https://github.com/ONSdigital/response-operations-ui/pull/895/files#diff-4310204905351551516d7296e7015779cfedf1a848e13f44010c7449e0b0a392R437) no longer has the responsibility of retrieving sample summary data, it would usually be renamed more appropriately `get_collection_exercises_by_survey_id`. Knowing this, have a look at what additional refactoring you could do to make your changes more efficient and remove redundancy.

I've renamed the function and its usages accordingly.

response_operations_ui/controllers/collection_exercise_controllers.py
```diff
@@ -435,16 +434,11 @@ def link_sample_summary_to_collection_exercise(collection_exercise_id, sample_su
     return response.json()
 
 
-def get_collection_exercises_with_samples_by_survey_id(survey_id):
+def get_collection_exercises_by_survey_id(survey_id):
     logger.info("Retrieving collection exercise with samples", survey_id=survey_id)
```
arroyoAle, line 438: Is this method even needed anymore? All it does is logs and calls another method in the same controller

arroyoAle (changes requested):
Looks good other than the maybe redundant method step

response_operations_ui/controllers/collection_exercise_controllers.py
anwilkie: Fixed. I've rewritten calls of this function to use the one nested inside.

arroyoAle (approved):
Tests pass, acceptance and unit, I'm happy with this

matthew-robinson-ons (approved):
Also tested acceptance tests, unit tests and no other issues found.

PR #896, opened 2023-11-06
RAS-315 Refactor collection-exercise

LJBabbage:
# What and why?
This PR moves the logic from the frontend to the backend and puts in some best practises of type hinting and testing. There can be more done in this space, but I have had to be realistic what could be done in the time, in the end of the day we are populating a 3 column table, we should be able to do all the preparation in the backend to hand to a frontend to display

# How to test?
Elements that need to be tested.

- Edit permission - create a basic CE with an account that can, then logout and come back in with one that can't and make sure there are no links
- Check all the links are correct, i.e the text is right, edit when there is a value add when there isn't, also check they work and populate the right fields in the db
- Locking - make sure fields which should be locked, are not editable. These will be Period ID and Reporting period, or anything in the past. To lock a CE it should be in one of these states (LIVE", "READY_FOR_LIVE", "EXECUTION_STARTED", "VALIDATED", "EXECUTED", "ENDED") for this test just complete the CE and press the green button that will do it.
- Status alerts (I don't think these ever appear normal and I have no idea why they are here, but I have migrated them) The only way you can do this is go to the db and change the status of the event (SELECT * FROM collectionexercise.event) Change to FAILED or RETRY etc.
- Make sure collection instruments and sample work the same
- Make sure reminders/nudges generated as before

# Trello

### DIFF
--- a/response_operations_ui/contexts/collection_exercise.py
+++ b/response_operations_ui/contexts/collection_exercise.py
@@ -41,7 +41,7 @@ def build_ce_context(ce_details: dict, has_edit_permission: bool, locked: bool) 
         has_edit_permission,
     )
 
-    ce_info_events = _ce_events(ce, events)
+    ce_info_events = _ce_info_events(ce, events)
     action_date_events = _action_date_events(events)
     ce_info = _build_event_section(ce, has_edit_permission, locked, ce_info_events, short_name, survey_ref)
     action_dates = _build_event_section(ce, has_edit_permission, locked, action_date_events, short_name, survey_ref)
@@ -80,7 +80,7 @@ def _build_event_section(
     return section
 
 
-def _ce_events(ce: dict, events: dict) -> dict:
+def _ce_info_events(ce: dict, events: dict) -> dict:
     return {
         "period_id": {
             "label": "Period ID *",
--- a/response_operations_ui/contexts/collection_exercise.py
+++ b/response_operations_ui/contexts/collection_exercise.py
@@ -1,0 +1,265 @@
+from flask import url_for
+
+REMINDER_EMAILS = {
+    "reminder": "First reminder",
+    "reminder2": "Second reminder",
+    "reminder3": "Third reminder",
+    "reminder4": "Fourth reminder",
+    "reminder5": "Fifth reminder",
+}
+
+NUDGE_EMAILS = {
+    "nudge_email_0": "First nudge email",
+    "nudge_email_1": "Second nudge email",
+    "nudge_email_2": "Third nudge email",
+    "nudge_email_3": "Fourth nudge email",
+    "nudge_email_4": "Fifth nudge email",
+}
+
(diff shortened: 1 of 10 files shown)


anwilkie (approved):
I went through the process of creating a CE and testing out the functionality as it completes its journey, and it seems like everything (fields, permissions, etc.) works okay. The acceptance tests have been run using the cucumber PR, and those pass, as well.

response_operations_ui/contexts/collection_exercise.py
```diff
@@ -0,0 +1,265 @@
...
+    ci_table = _build_ci_table(
+        collection_instruments,
+        locked,
+        ce_details["survey"]["surveyMode"],
+        short_name,
+        ce["exerciseRef"],
+        has_edit_permission,
+    )
+
+    ce_info_events = _ce_events(ce, events)
```
arroyoAle, line 44: For consistency - _ce_info_events rather than _ce_event
```suggestion
    ce_info_events = _ce_events(ce, events)
```

response_operations_ui/contexts/collection_exercise.py
```diff
@@ -0,0 +1,265 @@
...
+NUDGE_EMAILS = {
+    "nudge_email_0": "First nudge email",
+    "nudge_email_1": "Second nudge email",
+    "nudge_email_2": "Third nudge email",
+    "nudge_email_3": "Fourth nudge email",
+    "nudge_email_4": "Fifth nudge email",
+}
+
+CI_TABLE_LINK_TEXT = {
+    "EQ": {"no_instrument": "Add", "restricted": "View", "has_permission": "Select or Add"},
```
arroyoAle, line 20: I realise this bit is copied over, but when the user does not have survey edit permissions, they cannot add, upload or modidy EQ or SEFT instruments, so it showing Add and Upload where there are no instuments indicates to the user they would be able to do that when they are not. This probably seems like another card but I've just noticed it

arroyoAle (changes requested):
Everything works and all tests pass. However, there is no test coverage over the response changing builder, and my other two comments above

response_operations_ui/contexts/collection_exercise.py
LJBabbage: I would agree, definitely a bug, but also not connected to this PR.

LJBabbage:
> Everything works and all tests pass. However, there is no test coverage for the response changing builder (https://github.com/ONSdigital/response-operations-ui/blob/67af04672e0cceb866233424e0940b180474f0c8/response_operations_ui/contexts/collection_exercise.py#L258-L266), and my other two comments above

I should of left it where it was ☹️ , arguably out of scope as it was just moved, but I did create a new file and that file should therefore be covered

arroyoAle (approved):
All tests work and pass, UI looks as it should and functions properly

PR #901, opened 2023-11-09
Correct response chasing error with status check

LJBabbage:
# What and why?
In my last story I had to move response chasing, but I didn't change it, however what I hadn't picked up was there was data mutation on the dict here https://github.com/ONSdigital/response-operations-ui/blob/main/response_operations_ui/views/collection_exercise.py#L153. Basically doing this to the state element {"LIVE": "Live", "ENDED": "Ended"} Ironically we shouldn't be doing this but I am not going to unpick it now and just match it. 

# How to test?
Create a CE and put it in live/ended and make sure response chasing is there.
# Trello


SteveScorfield (approved):
Unit and acceptance tested. Tested UI going from live to ended. All works correctly. PR approved.

PR #902, opened 2023-11-09
Test and fix case controller local development test failures

pricem14pc:
# What and why?

The recently updated `urllib3` library (from 1.X to 2.X) has introduced a local development environment (Mac Book build) test failure with an existing RequestMock. The URL mock of a `204` (no content) that returns an empty array `[ ]` throws a ChunkedEncodingError under urllib3 2.X. This doesn't happen under urllib3 1.X. This also doesn't happen in GHA under urllib3 2.X (Linux build)

By definition, a `204` should not return any content. Having tested, the case service doesn't actually return any content for these endpoints when it returns a `204` and it's only the test mock that is wrong. Hence the empty array in the mock has been removed. The test assertion still checks for an empty array `[ ]` as the internal method under test is returning an empty array when it receives the `204` from case.

# How to test

Ensure all tests pass in both GHA and in a local development (Mac Book) environment


SteveScorfield (approved):
Tests passed. PR approved

LJBabbage (approved):

PR #914, opened 2024-02-02
RAS-939 Survey Status Change

arroyoAle:
# What and why?
Adds a new case group transition from `complete` to `not started` to allow business users a new status change in rOps
# How to test?
- deploy this and https://github.com/ONSdigital/rm-case-service/pull/289 to your environment
- run the acceptance tests to populate your env
- log into frontstage and complete the survey
- go to http://localhost:8085/reporting-units/49900000001 and navigate to change the status from complete to not started
- check that the survey appears in the todo page in frontstage
- check that http://localhost:8085/reporting-units/49900000001 and http://localhost:8085/case/49900000001/response-status?survey=QBS&period=1912 look the same and act the same as they do on main including with and without permissions
# Jira
https://jira.ons.gov.uk/browse/RAS-939


response_operations_ui/templates/reporting-unit-survey.html
```diff
@@ -67,9 +67,9 @@ <h3 id="COLLECTION_EXERCISES" class="ons-u-fs-m">Collection exercises</h3>
                         {% set status_class = 'ons-status--error' %}
                     {% endif %}
 
-                    {% if ce.responseStatus in ['Not started', 'In progress'] and isReportingUnitPermission  %}
+                    {% if ce.responseStatus in ['Not started', 'In progress','Completed', 'Completed by phone', 'No longer required'] and isReportingUnitPermission  %}
                         {% set survey_link = '&nbsp;   <a href="' + url_for("case_bp.get_response_statuses", ru_ref=ru.sampleUnitRef, survey=survey.shortName, period=ce.exerciseRef) + '">Change</a>' %}
-                    {% elif ce.responseStatus in ['Completed', 'Completed by phone', 'No longer required'] and isReportingUnitPermission  %}
+                    {% elif isReportingUnitPermission  %}
```
arroyoAle, line 72: I'm unsure about this since I added the permission check in the next page as well so this might be worth just being an else and the `view` link

arroyoAle: I decided to just make this an if else and show the view to all users as that is how it is used across the rest of the site. It just made more sense to me

response_operations_ui/templates/reporting-unit-survey.html
```diff
@@ -67,12 +67,10 @@ <h3 id="COLLECTION_EXERCISES" class="ons-u-fs-m">Collection exercises</h3>
                         {% set status_class = 'ons-status--error' %}
                     {% endif %}
 
-                    {% if ce.responseStatus in ['Not started', 'In progress'] and isReportingUnitPermission  %}
+                    {% if ce.responseStatus in ['Not started', 'In progress','Completed', 'Completed by phone', 'No longer required'] and isReportingUnitPermission  %}
```
LJBabbage, line 70: This isn't going to work in conjunction with eQ, QuestionnaireState which holds the answers for eQ will still be present if Completed by phone or No longer required has been used and the respondent started the survey. eQ only deletes QuestionnaireState if it's completed by them

LJBabbage: It would be nice to take the opportunity to move to this to the backend, removing the logic from the template. It only needs the check and link generation, not everything.

tests/views/test_change_response_status.py
```diff
@@ -331,3 +356,43 @@ def test_not_started_status_is_present_for_no_longer_required(self, mock_request
...
+
+        response = self.client.get(f"/case/{ru_ref}/response-status?survey={short_name}&period={period}")
+
+        data = response.data
+        self.assertEqual(response.status_code, 200)
+        self.assertIn(b"Completed", data)
+        self.assertIn(b"Not started", data)
+
+    @requests_mock.mock()
+    def test_not_started_status_is_not_present_for_not_started_without_permission(self, mock_request):
```
LJBabbage, line 382: I find this name difficult to understand, specially with 2 not_started in it

tests/views/test_change_response_status.py
```diff
@@ -331,3 +356,43 @@ def test_not_started_status_is_present_for_no_longer_required(self, mock_request
         self.assertEqual(response.status_code, 200)
         self.assertIn(b"No longer required", data)
         self.assertIn(b"Not started", data)
+
+    @requests_mock.mock()
+    def test_not_started_status_is_present_for_completed(self, mock_request):
```
LJBabbage, line 361: We really should try to move away from this pattern or massive mocking at the highest level as it gets messy. The functionality should in the backend which then means we can unit test the link creation combinations.

LJBabbage (changes requested):

response_operations_ui/templates/reporting-unit-survey.html
pricem14pc: > This isn't going to work in conjunction with eQ, QuestionnaireState which holds the answers for eQ will still be present if Completed by phone or No longer required has been used and the respondent started the survey. eQ only deletes QuestionnaireState if it's completed by them

The issue of EQ post submission state has also been raised when changing from COMPLETED to NOT STARTED. Depending on the time-bound post submission state, and the post-submission cookie state the respondent will not necessarily be presented with a new 'blank' questionnaire as the response_id may resolve to an exiting state.

arroyoAle: I haven't changed the functionality of the Completed by Phone and No longer required rOps users were already able to change those to Not started just that before they were clicking a link that said `View` and were then presented with options to change the status - I've moved the logic to the backend

tests/views/test_change_response_status.py
arroyoAle: Updated to be more clear

tests/views/test_change_response_status.py
arroyoAle: I'll leave this in but I have moved the link creation to the backend with additional tests

SteveScorfield (commented):
I have tested these changes and reviewed the code. All looks good to me.

SteveScorfield (approved):

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+        + '<a href="'
+        + hyperlink
+        + '">'
+        + hyperlink_text
+        + "</a>"
+    )
+
+
+def _build_respondents_section(
+    respondents: list, case, collection_exercises: list, ru: dict, survey: dict, unused_iac: str, permissions: list
```
LJBabbage, line 78: This is another situation where you should pass what it needs, it will then make the assignment cleaner

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+):
+    table = []
+    for respondent in respondents:
+        row = {}
+        if permissions[0]:
+            if unused_iac:
+                row["enrolment_code"] = unused_iac
```
LJBabbage, line 85: Is enrolment_code not assigned if it has permissions and unused

LJBabbage: This also looks like a walrus operator

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+                    ru_ref=ru["sampleUnitRef"],
+                    trading_as=collection_exercises[0]["tradingAs"],
+                    survey_ref=survey["surveyRef"],
+                    survey_name=survey["shortName"],
+                )
+                row["enrolment_code_hyperlink_text"] = "Generate new enrollment code"
+        else:
+            row["enrolment_code"] = ""
+        row["contact_details"] = {
+            "Name": f"{respondent['firstName']} {respondent['lastName']}",
```
LJBabbage, line 101: Are you using the key in the frontend as I usually wouldn't expect capitalised keys

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+        }
+        row["account_status"] = (
+            '<span class="ons-status '
+            + _select_account_status_class(respondent["status"])
+            + '">'
+            + respondent["status"].capitalize()
+            + "</span>"
+        )
+        row["enrolment_status"] = _build_enrolment_status(respondent, ru, survey, permissions)
+        if permissions[1]:
```
LJBabbage, line 113: As above, not sure what permission[1] means

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+                {"name": "survey_id", "value": survey["id"]},
+                {"name": "msg_to_name", "value": row["contact_details"]["Name"]},
+                {"name": "msg_to", "value": respondent["id"]},
+            ]
+        table.append(row)
+
+    return table
+
+
+def _select_account_status_class(status):
```
LJBabbage, line 128: Feels like a one liner

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+    if status == "SUSPENDED":
+        return "ons-status--error"
+    else:
+        return "ons-status--success"
+
+
+def _build_enrollment_status_hyperlink(respondent, ru, survey, permissions):
+    if respondent["enrolmentStatus"] == "ENABLED":
```
LJBabbage, line 136: State as constant maybe?

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+            respondent_first_name=respondent["firstName"],
+            respondent_last_name=respondent["lastName"],
+            business_id=ru["id"],
+            trading_as=ru["trading_as"],
+            change_flag="ENABLED",
+            tab="reporting_units",
+        )
+        hyperlink_text = "Re-enable"
+    if not permissions[0]:
+        hyperlink = None
```
LJBabbage, line 177: This feels strange, you set it above, then clear it out if no permission, feels the wrong order

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _build_enrolment_status(respondent, ru, survey, permissions):
+    (
+        hyperlink,
+        hyperlink_text,
+        enrolment_status_class,
+    ) = _build_enrollment_status_hyperlink(respondent, ru, survey, permissions)
+    enrolment_status = (
+        '<span class="ons-status '
```
LJBabbage, line 189: As above with HTML

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+            tab="reporting_units",
+        )
+        hyperlink_text = "Re-enable"
+    if not permissions[0]:
+        hyperlink = None
+
+    return hyperlink, hyperlink_text, status_class
+
+
+def _build_enrolment_status(respondent, ru, survey, permissions):
```
LJBabbage, line 182: Little inconsistent with type hinting

tests/context/conftest.py
```diff
@@ -242,3 +242,900 @@ def ce_details_dynamic_event_deleted(ce_details_dynamic_event):
...
+    error_collection_exercise_with_details = collection_exercises_with_details.copy()
+    error_collection_exercise_with_details[0]["responseStatus"] = "Error"
+    return error_collection_exercise_with_details
+
+
+@pytest.fixture
+def multiple_collection_exercises_with_details():
+    return [
+        {
+            "id": "57e06b27-a62b-4d8e-8d57-7b6b80067e4f",
```
LJBabbage, line 377: Feels to me this chunk can be extracted

tests/context/conftest.py
```diff
@@ -242,3 +242,900 @@ def ce_details_dynamic_event_deleted(ce_details_dynamic_event):
...
+                "partyId": "258b95c5-1e48-4357-881c-ad4d544eab32",
+            },
+            {
+                "businessRespondentStatus": "ACTIVE",
+                "enrolments": [{"enrolmentStatus": "ENABLED", "surveyId": "02b9c366-7397-42f7-942a-76dc5876d86d"}],
+                "partyId": "985bf97e-4f03-4898-92ef-dd7aac23ab08",
+            },
+        ],
+        "attributes": {
+            "birthdate": "01/09/1993",
```
LJBabbage, line 622: This again feels like it can be extracted and used as a fixture import

tests/context/conftest.py
```diff
@@ -242,3 +242,900 @@ def ce_details_dynamic_event_deleted(ce_details_dynamic_event):
...
+            "surveyId": "02b9c366-7397-42f7-942a-76dc5876d86d",
+        },
+        "caseEvents": None,
+    }
+
+
+@pytest.fixture
+def multiple_cases():
+    return {
+        "state": "ACTIONABLE",
```
LJBabbage, line 794: Again this looks like the same chunk bar an odd change

tests/context/test_case_context.py
```diff
@@ -0,0 +1,129 @@
+from response_operations_ui.contexts.case import build_case_context
+
+RU_REF = "49900000001"
+SURVEY_SHORT_NAME = "QBS"
+CASE_GROUP_ID = "6fef0397-f07b-4d65-8988-931cec23057f"
+CE_PERIOD = "1912"
+SURVEY_ID = "02b9c366-7397-42f7-942a-76dc5876d86d"
+
+
+def test_case_context_with_permissions(app, transitions_for_complete_case, expected_case_context_with_all_permissions):
```
LJBabbage, line 10: You might be able to use fixture params here

tests/views/test_change_response_status.py
LJBabbage: You can, but now you have done the work to test the dict being passed, you don't need to. The idea is you should only need to test the template logic

LJBabbage (changes requested):
Looking better and nice to see some pytest fixtures use. It needs a bit or refinement with function names, structure and reuse.

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,55 @@
+from flask import url_for
+
+
+def build_case_context(
```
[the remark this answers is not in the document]
arroyoAle: Updated this

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,55 @@
...
+    ru_ref,
```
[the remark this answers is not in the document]
arroyoAle: I've added typing where i can

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,55 @@
...
+        case_group_id,
+        ce_period,
+        allowed_transitions_for_case,
+        survey_id,
+        has_reporting_unit_permission,
+    )
+    return {"change_response_status": change_response_status}
+
+
+def _build_change_response_status(
```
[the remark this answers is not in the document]
arroyoAle: You're right I feel like I was on autopilot when writing this - now fixed

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,55 @@
...
+    survey_short_name,
+    case_group_id,
+    ce_period,
+    allowed_transitions_for_case,
+    survey_id,
+    has_reporting_unit_permission,
+):
+    if has_reporting_unit_permission:
+        change_response_status = {
+            "form_action": url_for(
```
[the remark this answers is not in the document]
arroyoAle: updated this

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,55 @@
...
+            "cancel_link": url_for("reporting_unit_bp.view_reporting_unit_survey", ru_ref=ru_ref, survey_id=survey_id),
+        }
+    else:
+        change_response_status = {}
+    return change_response_status
+
+
+def _generate_radios(allowed_transitions_for_case):
+    radios = []
+    for index, (event, status) in enumerate(allowed_transitions_for_case.items()):
```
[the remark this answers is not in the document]
arroyoAle: Swapped list comprehension in

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
+from flask import url_for
+
+
+def build_reporting_units_context(
+    collection_exercises: list, ru: dict, survey: dict, respondents: list, case, unused_iac: str, permissions: list
+):
```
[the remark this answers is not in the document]
arroyoAle: added the return type

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+    respondents_section = _build_respondents_section(
+        respondents, case, collection_exercises, ru, survey, unused_iac, permissions
+    )
+
+    return {"collection_exercise_section": collection_exercise_section, "respondents_section": respondents_section}
+
+
+def _build_collection_exercise_section(collection_exercises: list, ru: dict, survey: dict, permissions: list):
+    table = []
+    for ce in collection_exercises:
```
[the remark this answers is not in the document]
arroyoAle: Added list comprehension

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+            "region": ce["companyRegion"],
+            "response_status": ce["responseStatus"],
+            "status": _build_ce_status(ru, survey, ce, permissions),
+        }
+        table.append(row)
+
+    return table
+
+
+def _select_status_class(status):
```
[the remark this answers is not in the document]
arroyoAle: updated this to use a dict

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+    hyperlink = url_for(
+        "case_bp.get_response_statuses",
+        ru_ref=ru["sampleUnitRef"],
+        survey=survey["shortName"],
+        period=ce["exerciseRef"],
+    )
+    hyperlink_text = "Change" if permissions[0] else "View"
+    status_class = _select_status_class(ce["responseStatus"])
+    response_status = ce["responseStatus"]
+    return (
```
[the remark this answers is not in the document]
arroyoAle: Updated this

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _build_ce_status(ru, survey, ce, permissions):
+    hyperlink = url_for(
+        "case_bp.get_response_statuses",
+        ru_ref=ru["sampleUnitRef"],
+        survey=survey["shortName"],
+        period=ce["exerciseRef"],
+    )
+    hyperlink_text = "Change" if permissions[0] else "View"
```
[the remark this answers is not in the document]
arroyoAle: I've updated this to a dict to make it more clear since the messages permission is used further down

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+    elif status == "No longer required":
+        return "ons-status--dead"
+    else:
+        return "ons-status--error"
+
+
+def _build_ce_status(ru, survey, ce, permissions):
+    hyperlink = url_for(
+        "case_bp.get_response_statuses",
+        ru_ref=ru["sampleUnitRef"],
```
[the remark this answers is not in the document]
arroyoAle: Updated this

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+    )
+
+
+def _build_respondents_section(
+    respondents: list, case, collection_exercises: list, ru: dict, survey: dict, unused_iac: str, permissions: list
+):
+    table = []
+    for respondent in respondents:
+        row = {}
+        if permissions[0]:
```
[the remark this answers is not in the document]
arroyoAle: Updated to a dict

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,199 @@
...
+                row["enrolment_code_hyperlink_text"] = "Generate new enrollment code"
+        else:
+            row["enrolment_code"] = ""
+        row["contact_details"] = {
+            "Name": f"{respondent['firstName']} {respondent['lastName']}",
+            "Email": respondent["emailAddress"],
+            "Tel": respondent["telephone"],
+        }
+        row["account_status"] = (
+            '<span class="ons-status '
```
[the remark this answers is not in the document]
arroyoAle: Updated as above

response_operations_ui/contexts/reporting_units.py
arroyoAle: Updated to receive only what it needs

response_operations_ui/contexts/reporting_units.py
arroyoAle: Yeah in the front end it checks the permissions again and doesn't try to display the enrolment code at all - also not sure how to put in a walrus operator here

response_operations_ui/contexts/reporting_units.py
arroyoAle: I was going to and then didn't this has been updated

response_operations_ui/contexts/reporting_units.py
arroyoAle: Updated to a dict

response_operations_ui/contexts/reporting_units.py
arroyoAle: Changed to a one liner

response_operations_ui/contexts/reporting_units.py
arroyoAle: Updated to contstant

response_operations_ui/contexts/reporting_units.py
arroyoAle: Updated this to make it make more sense

response_operations_ui/contexts/reporting_units.py
arroyoAle: Updated as above

response_operations_ui/contexts/reporting_units.py
arroyoAle: Added type hinting

tests/context/conftest.py
arroyoAle: Updated

tests/context/conftest.py
arroyoAle: Made into a fixture import

tests/context/conftest.py
arroyoAle: Updated this

tests/context/test_case_context.py
arroyoAle: Swapped for fixture params

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,35 @@
...
+def build_response_status_context(
+    ru_ref: str,
+    survey_short_name: str,
+    case_group_id: str,
+    ce_period: str,
+    allowed_transitions_for_case: dict,
+    survey_id: str,
+    has_reporting_unit_permission: bool,
+) -> dict:
+    if has_reporting_unit_permission:
```
LJBabbage, line 13: You have kind of doubled up here with the frontend and checked permissions twice ({% if isReportingUnitPermission %} in the template), so else is never reachable. However we shouldn't need to do isReportingUnitPermission in the template, as it's been done in the back end. Depends on what else uses it on how to pass it through, if it's just this then checking change_response_status is not null would work

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,231 @@
...
+        respondents,
+        case["id"],
+        collection_exercises[0]["id"],
+        collection_exercises[0]["tradingAs"],
+        ru["name"],
+        ru["sampleUnitRef"],
+        ru["id"],
+        survey["surveyRef"],
+        survey["shortName"],
+        survey["id"],
```
LJBabbage, line 31: Sorry my fault, I wasn't clear. If it's 1 or 2 elements in a list we should pass the value, more than that pass the list. It's a balance act between passing what it needs and how many vars for the function

response_operations_ui/contexts/case.py
```diff
@@ -0,0 +1,35 @@
...
+            ),
+            "radios": _generate_radios(allowed_transitions_for_case),
+            "cancel_link": url_for("reporting_unit_bp.view_reporting_unit_survey", ru_ref=ru_ref, survey_id=survey_id),
+        }
+    else:
+        change_response_status = {}
+    return {"change_response_status": change_response_status}
+
+
+def _generate_radios(allowed_transitions_for_case: dict):
```
LJBabbage, line 30: Missing the return type

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,231 @@
...
+    ce_status = {
+        "hyperlink": url_for(
+            "case_bp.get_response_statuses",
+            ru_ref=ru_ref,
+            survey=survey,
+            period=period,
+        ),
+        "hyperlink_text": "Change" if ru_permission else "View",
+        "status_class": (
+            CE_STATUS_CLASS[response_status] if response_status in CE_STATUS_CLASS.keys() else "ons-status--error"
```
LJBabbage, line 84: A get with default might be better here

response_operations_ui/contexts/reporting_units.py
```diff
@@ -0,0 +1,231 @@
...
+    survey_short_name: str,
+    survey_id: str,
+    unused_iac: str,
+    permissions: dict,
+) -> list:
+    table = []
+    for respondent in respondents:
+        row = {}
+        if permissions["reporting_unit_edit"]:
+            if unused_iac:
```
LJBabbage, line 109: I see it, the logic is a bit unusual as the frontend is also checking off the iac key being passed to it and diverting it [around](https://github.com/ONSdigital/response-operations-ui/blob/804edb9edfcb540f74c37cce6c6bd8f12de098f4/response_operations_ui/templates/reporting-unit-survey.html#L74). We should try not to do that as it kind of invalidates any unit tests in that area as there is the same logic both in the frontend and back.

SteveScorfield (approved):
I have tested this again and all works as expected. I'm also happy with the recommended code changes.

tests/context/conftest.py
```diff
@@ -242,3 +242,815 @@ def ce_details_dynamic_event_deleted(ce_details_dynamic_event):
     ce_details_dynamic_event_deleted = ce_details_dynamic_event.copy()
     del ce_details_dynamic_event_deleted["events"]["reminder2"]
     return ce_details_dynamic_event_deleted
+
+
+@pytest.fixture()
+def collection_exercises_with_details():
```
LJBabbage, line 248: This has a lot of stuff we don't need. Notice up the top there is a slim line ce_details. The idea being that is the starting point, each fixture below extents depending on what it needs to test. It doesn't need anything that it is not testing, it's not going to the frontend so dates, events, samples etc aren't needed for everything. So if it's build_reporting_units_context your testing, call your fixture appropriately extent the CE base and just make sure the values are populated to test, everything else could be None.

It will reduce the code massively and make it easier to maintain

tests/context/conftest.py
```diff
@@ -242,3 +242,815 @@ def ce_details_dynamic_event_deleted(ce_details_dynamic_event):
...
+
+@pytest.fixture
+def transitions_for_completed_by_phone_case():
+    return {"COMPLETED_TO_NOTSTARTED": "Not started"}
+
+
+@pytest.fixture
+def transitions_for_no_longer_required_case():
+    return {"COMPLETED_TO_NOTSTARTED": "Not started"}
+
```
LJBabbage, line 1016: Not sure why these are all the same

tests/context/conftest.py
```diff
@@ -242,3 +242,815 @@ def ce_details_dynamic_event_deleted(ce_details_dynamic_event):
...
+        "NO_LONGER_REQUIRED": "No longer required",
+    }
+
+
+@pytest.fixture
+def transitions_for_not_started_case():
+    return {
+        "COMPLETED_BY_PHONE": "Completed by phone",
+        "NO_LONGER_REQUIRED": "No longer required",
+    }
```
LJBabbage, line 1031: Same again

tests/context/test_case_context.py
```diff
@@ -0,0 +1,176 @@
+from response_operations_ui.contexts.case import build_response_status_context
+
+
+def test_case_context_with_permissions(
```
LJBabbage, line 4: The test name needs a tweak, we are testing the response status

LJBabbage (changes requested):
Code looking good, a few minor things, tests need a bit more work regarding naming and fixture use

response_operations_ui/contexts/case.py
arroyoAle: I've removed this permissions check from the frontend and made it just check for it existing

response_operations_ui/contexts/reporting_units.py
arroyoAle: Ahh yeah that makes more sense - updated

response_operations_ui/contexts/case.py
arroyoAle: added return type

response_operations_ui/contexts/reporting_units.py
arroyoAle: Updated to get with a default

tests/context/conftest.py
arroyoAle: Updated and slimmed down

tests/context/conftest.py
arroyoAle: removed the repeated ones

tests/context/conftest.py
arroyoAle: As above

tests/context/test_case_context.py
arroyoAle: Updated

tests/context/test_case_context.py
```diff
@@ -0,0 +1,189 @@
+from response_operations_ui.contexts.case import build_response_status_context
+
+
+def test_response_status_context_with_permissions(
```
LJBabbage, line 4: These are fine, and one for another time, but pytest.mark.parametrize would of worked well here and saved a lot of repetition, https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#parametrize. You would of had to use something like getfixturevalue in the function for the compare

LJBabbage (approved):
Looks good and works as expected, seen a small issue not connected to the card, respondent seems to be missing, but it's the same in latest. Bar that one recommendation for future tests
image

PR #916, opened 2024-02-06
RAS-999: Migrate Rops from flask-pagination to DS pagination

SteveScorfield:
# What and why?
Flask-pagination needs upgrading to a later version. This has caused some issues. I have looked at upgrading it, however, moving to the ONS DS has proven to be easier than first thought. I have decided to move to the ONS pagination. This card achieves this.

# How to test?
1. Build your env and run the acceptance tests (to generate data)
2. Pull this PR to your local env and run it locally
3. Update the following:
    1. The limit in response_operations_ui/views/admin/manage_user_accounts.py to under the amount of user accounts 
    we have
   2. The PARTY_BUSINESS_RESULTS_PER_PAGE value to under the amount of businesses we have (please note that 
   due to the data there will be 2 less than expected as there are 3 companies with the same name and only 1 is retrieved). 
   This will show the pagination for reporting units
   3. The PARTY_RESPONDENTS_PER_PAGE to under the number of respondents we have to show the pagination for the 
   respondent page
  4. In the url for messages add ?limit=2 to the end. Ensure that there are more than 2 messages to show pagination
5. Pagination should work for all these areas.

Run the unit tests and acceptance tests.

Any issues, let me know
# Jira
https://jira.ons.gov.uk/browse/RAS-999
and
https://jira.ons.gov.uk/browse/RAS-1002


response_operations_ui/views/respondents.py
```diff
@@ -447,5 +438,5 @@ def _generate_pagination_href(email_address="", first_name="", last_name=""):
     if last_name != "":
         href_string_list.append("lastname=" + last_name)
     # This is needed for custom hrefs as per the flask-paginate docs
-    href_string_list.append("page={0}")
+    href_string_list.append("page=")
```
matthew-robinson-ons, line 441: is this still needed? Should probably remove the reference to flask-paginate either way to avoid confusion

SteveScorfield: Absolutely. I missed this. I'm going to take a look now to see what can be improved.

matthew-robinson-ons (changes requested):
I think there is still more work to be done here unfortunately, the pagination is not displaying correctly on the messages tabs. The technical message tab doesn't display any pagination controls.

matthew-robinson-ons (approved):
Pagination is displaying and working correctly with the design system across RU, Respondent, Messaging and admin tabs.

arroyoAle (approved):
Changes look good and pagination works - happy with this

PR #920, opened 2024-03-12
RAS-1100 Separate out selected_ci_functions from main.js

LJBabbage:
# What and why?
Moves selected_ci_functions so it isn't called on every ROPS page and doesn't produce console errors 

image

# How to test?
Navigate to CI instrument upload page for an eQ and add a couple of instruments. Make sure the select all/unselect all button still works. Confirm for live CE for eQ when you view CI the page is fine. Also confirm error on console output is gone
# Jira


response_operations_ui/assets/scss/includes/_overrides.scss
```diff
@@ -187,3 +187,19 @@ button.hidden {
 .error-message {
     color: #d0021b;
 }
+
+.ons-tabs__list {
```
LJBabbage, line 191: This was incorrectly implemented in this PR https://github.com/ONSdigital/response-operations-ui/pull/882. The CSS was put into the auto generated code, so would be overwritten if anyone called make minify. This is the source file, the PR put it in the destination file

response_operations_ui/static/css/main.css
```diff
@@ -1,17 +1 @@
...
-
-.ons-tabs__list {
```
LJBabbage, line 3: Described above

response_operations_ui/templates/collection_exercise/ce-eq-instrument-section.html
```diff
@@ -8,6 +8,11 @@
 {% from "components/table/_macro.njk" import onsTable %}
 {% from "components/upload/_macro.njk" import onsUpload %}
 
+{% block head %}
```
LJBabbage, line 11: This has to go in the extending template, because of that it can't go a level down into the import. Because of this I have had to add an If to make sure it's not the view page. Not ideal, but I can't realistically restructure in this story

SteveScorfield (approved):
I've ran acceptance tests and have tested the UI. All work as expected. Code changes look fine. PR approved.

arroyoAle (approved):
Functionality unchanged and code looks good

PR #921, opened 2024-03-14
RAS-1088 Implement 48 hour grace period for survey change of status

arroyoAle:
# What and why?
Updating confirm button to make it disabled when no radios can be selected
# How to test?
deploy this to your env, complete a survey and try to reverted to not started, the radio and the confirm button should be disabled
# Jira
[RAS-1088](https://jira.ons.gov.uk/browse/RAS-1088)


response_operations_ui/contexts/case.py
```diff
@@ -23,25 +23,36 @@ def build_response_status_context(
...
-def _generate_radios(allowed_transitions_for_case: dict, case_completed_time: datetime) -> list:
+def _generate_radios_section(
+    allowed_transitions_for_case: dict, case_completed_time: datetime, ru_ref: str, survey_id: str
+) -> dict:
     radios = []
+    enabled_radios = True
     for index, (event, status) in enumerate(allowed_transitions_for_case.items()):
         radio = {"id": f"state-{index + 1}", "label": {"text": status}, "value": event}
         if event == "COMPLETED_TO_NOTSTARTED" and case_completed_time:
+            enabled_radios = False
```
SteveScorfield, line 43: I'd move this boolean to something like:

 ```
       if event == "COMPLETED_TO_NOTSTARTED" and case_completed_time:
            complete_to_not_started_wait_time = app.config["COMPLETE_TO_NOT_STARTED_WAIT_TIME"]
            if (datetime.now() - case_completed_time) < timedelta(seconds=complete_to_not_started_wait_time):
                radio["attributes"] = {"disabled": "true"}
                radio["label"][
                    "description"
                ] = "Status can only be changed after 48 hours have passed since the submission"
                enabled_radios = False
            else:
                enabled_radios = True
```

File: response_operations_ui/contexts/case.py [hunk 2]
```diff
@@ -23,25 +23,36 @@ def build_response_status_context(
...
                     "description"
                 ] = "Status can only be changed after 48 hours have passed since the submission"
+            else:
+                enabled_radios = True
         radios.append(radio)
 
-    return radios
+    return {
+        "radios": radios,
+        "confirm_button_variant": None if enabled_radios else "disabled",
```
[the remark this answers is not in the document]
SteveScorfield: I've been looking at this and I can't see a better way so far.

SteveScorfield (commented):
I've had a look at code and it looks fine and is working. I have commented on the booleans in the if else.

response_operations_ui/contexts/case.py
arroyoAle: I think I may leave it as is because if we add another state change it may not require the 48 grace period so if one radio is enabled we want the button to be enabled - and if the disabled one is processed second in the loop, the button will remain disabled while a transition can be made

SteveScorfield: Ah okay. That makes.

SteveScorfield (approved):
Tested and approved.

PR #924, opened 2024-03-19
RAS-1105 Update email-validator

LJBabbage:
# What and why?
There is a vulnerability in Dnspython 2.5.0 which is used by email-validator, update email-validator to 2.1.1 which update Dnspython to 2.6.0

# How to test?
Make sure tests pass and everything works as before
# Jira

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -40,8 +40,8 @@ et_xmlfile = "*"
 Jinja2 = "*"
 MarkupSafe = "*"
 WTForms = "*"
-email_validator = "*"
 jwcrypto = "*"
+email-validator = "*"
 
 [dev-packages]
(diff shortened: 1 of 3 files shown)


Pipfile
```diff
@@ -40,8 +40,8 @@ et_xmlfile = "*"
 Jinja2 = "*"
 MarkupSafe = "*"
 WTForms = "*"
-email_validator = "*"
 jwcrypto = "*"
+email-validator = "*"
```
LJBabbage, line 44: It wanted a name change as well

matthew-robinson-ons (approved):
All looks good

arroyoAle (approved):

PR #932, opened 2024-04-22
Fix issue with minify

SteveScorfield:
# What and why?
This is the fix for the error found in rOps. This had nothing to do with the minify upgrades. It was down to an overlooked pipfile.lock. The flask-session has been pinned at 0.6.0. Version 0.7.0 cause breaking changes which were fixed by 0.8.0. However, 0.8.0 introduced another breaking change into our code. This caused a TypeError when checking the session date in the uaa_controller.py:

`if "permissions" not in session or session["permissions"]["expiry"] < datetime.now():
        refresh_permissions(user_id)`

A ticket to investigate this issue can be found here: https://jira.ons.gov.uk/browse/RAS-1147

# How to test?
Deploy this to your dev env, run acceptance tests. Log into rOps and click 'Create survey'. This should respond correctly and not error.

# Jira
https://jira.ons.gov.uk/browse/RAS-1110


arroyoAle (changes requested):
All works but should we pin the offending package in the Pipfile so it doesn't cause this issue again by mistake before we get round to sorting it out

SteveScorfield:
> All works but should we pin the offending package in the Pipfile so it doesn't cause this issue again by mistake before we get round to sorting it out

Agreed. I've pinned it now.

arroyoAle (approved):
Happy with this now, rOps works as expected

anwilkie (approved):
Tested out according to instructions, and it works okay.

PR #933, opened 2024-04-23
Remove alert-library.js

mfcrocker:
# What and why?
We suspected that alert-library.js wasn't being used due to Design System alerting and that seems to be the case. This PR removes it.

# How to test?
Smoke test, especially around error handling

# Jira


matthew-robinson-ons (commented):
Tested locally and all looks good, the flash session fix needs to be merged to fully test  as currently getting the internal server error when deployed through spinnaker.

SteveScorfield (approved):
Tested locally and works as expected. Once the fix has been merged with this, it will work fine in GCP. Approved

matthew-robinson-ons (approved):
tested locally and through spinnaker. All looks as expected

PR #936, opened 2024-04-25
Fix generate new enrolment code link

pricem14pc:
# What and why?

This fixes a bug in the "Generate new enrolment code" hyperlink. This link is currently only formed correctly if registered respondents exist for the RU/survey. If no respondents exist the link is malformed and doesn't work. Respondents who have not verified their accounts within 80 hours will be deleted leaving no active UAC codes and no way for the business to generate new codes.

The link should be formed and presented if an active IAC code doesn't already exist, even if no respondents are registered for the business.

# How to Test

- Create a collection exercise through to it going live.
- Get the IAC from the `/reporting-units/{ru_ref}/surveys/{survey_id}` page.
- Register a user with the code but don't verify them.
- Mouse over (but don't click) the "Generate new enrolment code" link and check its correct in the `/reporting-units/{ru_ref}/surveys/{survey_id}` page  (see the unit test in this PR for the expected correct format). You should see the respondent showing as pending verification in the page.
- Delete the party `respondent` record (inc its `business_respondent`, `enrolment` and `pending_enrolment` records). You can do this manually via a Postgres client or use the delete account scheduling jobs if you hack at the dates to make the account creation date more than 80 hours (this is what surfaced the production bug)
- Refresh the `/reporting-units/{ru_ref}/surveys/{survey_id}` page and check there are no respondents showing in the page.
- Click the "Generate new enrolment code" link in the `/reporting-units/{ru_ref}/surveys/{survey_id}` page  (see the unit test in this PR for the expected correct format). A new enrolment code should be presented, and this should now show in the `/reporting-units/{ru_ref}/surveys/{survey_id}` page instead of a "Generate new enrolment code" link.
- Register a new user with this new enrolment code.

https://jira.ons.gov.uk/browse/RAS-1149


response_operations_ui/views/reporting_units.py
```diff
@@ -213,12 +213,26 @@ def view_reporting_unit_survey(ru_ref, survey_id):
...
         unused_iac = case["iac"]
 
-    logger.info("Successfully gathered data to view reporting unit survey data", ru_ref=ru_ref, survey_id=survey_id)
-
     permissions = {
         "reporting_unit_edit": user_has_permission("reportingunits.edit"),
         "messages_edit": user_has_permission("messages.edit"),
     }
+
+    enrolment_code_hyperlink = ""
```
pricem14pc, line 221: No point delegating this to a method in the method in `contexts/reporting_units.py` as the signature would be the same amount of code and there's no other use cases for this link.

response_operations_ui/views/reporting_units.py
```diff
@@ -213,12 +213,26 @@ def view_reporting_unit_survey(ru_ref, survey_id):
...
+    if permissions["reporting_unit_edit"]:
+        enrolment_code_hyperlink = url_for(
+            "reporting_unit_bp.generate_new_enrolment_code",
+            case_id=case["id"],
+            collection_exercise_id=survey_collection_exercises[0]["id"],
```
pricem14pc, line 226: This doesn't seem to yield a deterministic ordering of the collection exercises. They should be sorted but the `[0]` indexed exercise appears to change per run.

tests/views/test_reporting_units.py
```diff
@@ -58,6 +58,8 @@
 
 with open(f"{project_root}/test_data/party/business_reporting_unit.json") as fp:
     business_reporting_unit = json.load(fp)
+with open(f"{project_root}/test_data/party/business_reporting_unit_no_enrolments.json") as fp:
```
pricem14pc, line 61: This config is for the scenario this PR is resolving. There are no respondents registered for the RU/survey and as such the hyperlink was malformed. This test is to confirm the fix.

tests/views/test_reporting_units.py
```diff
@@ -296,6 +300,51 @@ def test_get_reporting_unit_survey_reporting_unit_edit_role(self, mock_request):
...
+            "ru_name=Bolts+and+Ratchets+Ltd&amp;"
+            "trading_as=Bolts+and+Ratchets+Ltd&amp;"
+            "survey_ref=221&amp;"
+            "survey_name=BLOCKS"
+        )
+
+        self.assertEqual(response.status_code, 200)
+
+        self.assertTrue(
+            enrolment_code_hyperlink1.encode() in response.data or enrolment_code_hyperlink2.encode() in response.data
```
pricem14pc, line 345: This is a consequence of the CEs not being ordered correctly. The `collection_exercise_id` can be one of multiple exercises returned, as such the hyperlink in the test will contain one or other of those IDs. We should improve on this asap.

arroyoAle (approved):
Code looks good and changes fix the issue in rOps an new code is now generated with and without a user enrolled

response_operations_ui/views/reporting_units.py
pricem14pc: Fixed with test data change to `scheduledStartDateTime`

tests/views/test_reporting_units.py
pricem14pc: Fixed with data change to `scheduledStartDateTime`

arroyoAle (approved):
Looks good and tested and enrolment code generates correctly both when there is and there isn't a respondent

PR #937, opened 2024-05-03
RAS-1101: Bug in secure messaging pre prod - Message sends twice

SteveScorfield:
# What and why?
When a message is sent and there is a delay, the submit button remains active. The submit can then be clicked 'X' amount of times leading to an equivalent 'X' amount of duplicates. The 'loader' variant has been added to the submit button. The submit button will no longer be active, a loading wheel will appear on the button and will prevent the button being clicked.

# How to test?

- Run acceptance tests
- Pull this PR
- Add 'time.sleep(30)' to https://github.com/ONSdigital/response-operations-ui/blob/3c540c5df10ea9241481515477207d86ed0f9229/response_operations_ui/controllers/message_controllers.py#L162 to cause a delay
- Run this PR locally
- Send a message in rOps
- The submit button should look like this 

image

-After the delay, you should be redirected to the inbox overview page

Additional testing:
-Deploy this PR to your env
-Run acceptance tests

# Jira
https://jira.ons.gov.uk/browse/RAS-1101

### DIFF
--- a/_infra/helm/response-operations-ui/Chart.yaml
+++ b/_infra/helm/response-operations-ui/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version
-version: 3.1.70
+version: 3.1.71
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 3.1.70
+appVersion: 3.1.71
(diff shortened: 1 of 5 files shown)


arroyoAle (commented):
This all works as expected and solves this issue in rOps, but this issue would also happen in frontstage

matthew-robinson-ons (approved):
Works as intended within rops

arroyoAle (approved):

PR #940, opened 2024-05-10
RAS-1147 Upgrade to flask-session 0.8.0 update in rOps

arroyoAle:
# What and why?
Bumping `flask-session` to version 0.8.0 which had breaking changes to how it handled datetime objects. This casts the object to an iso8061 string to store it in the session json and then back when it needs to compare it to `datetime.now()`
# How to test?
- Deploy this PR to your environment
- sign into rOps with the default `uaa_user` account
- click around making sure no 500 are raised because of a comparison error between str and datetime objects
# Jira
[RAS-1147](https://jira.ons.gov.uk/browse/RAS-1147)


matthew-robinson-ons (changes requested):
When sending a message i'm getting a 500. The message does still send however you get a `TypeError: Encoding objects of type Markup is unsupported` in the logs.

PR #941, opened 2024-05-15
RAS-1175 Remove disable-send.js from ROPs

LJBabbage:
# What and why?
disable-send.js is no longer used and double submission is now being blocked by the design system 
# How to test?
Make sure when sending a secure message that you can not send it twice, the DS loader should spin instead
# Jira


matthew-robinson-ons (approved):
Works as expected and unable to find any other references to disable-send

mfcrocker (approved):

PR #943, opened 2024-05-21
Add test-html to Makefile

mfcrocker:
# What and why?
pytest-html offers a nice way to show test reports when devving locally. This exposes it through a make command

# How to test?
`make test-html`

# Jira


LJBabbage (changes requested):
same as the other PR
"djlint is going to complain if this is run more than once, we should set the directory to just our html templates, so it ignores htmlcov report html. It would also be nice to lob a very basic addition in the readme. It's not obvious that the best way to use it is visiting /ras-frontstage/htmlcov/index.html"

LJBabbage (approved):

matthew-robinson-ons (approved):
html report is created and other changes look good

PR #944, opened 2024-05-29
RAS-1111 Update ROPs to call new survey business end point

LJBabbage:
# What and why?
Updates the call in view_reporting_unit_survey to use the new party service endpoint to find respondents by survey_id and business_id. There is also a small amount of refactoring and removal as well. view_reporting_unit_survey still needs a fair amount of work, so I have left it largely untouched bar the change in this PR

# How to test?
Deploy this PR and seed the db with data. Go to ROPs and search for a Reporting unit, drill down to the survey page of a business and confirm it still works and displays the correct information for respondents
image
Basically the PR is updating the Respondents section of the image above to call from a different endpoint. It is worth adding some more respondents, changing their state, sending a message and doing anything else you can think of
# Jira

### DIFF
--- a/_infra/helm/response-operations-ui/Chart.yaml
+++ b/_infra/helm/response-operations-ui/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version
-version: 3.1.75
+version: 3.1.76
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 3.1.75
+appVersion: 3.1.76
--- a/response_operations_ui/contexts/reporting_units.py
+++ b/response_operations_ui/contexts/reporting_units.py
@@ -13,22 +13,17 @@ ENROLMENT_PENDING = "PENDING"
 
 
 def build_reporting_units_context(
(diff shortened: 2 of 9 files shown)


SteveScorfield (approved):
Changes have been tested based on testing params. Acceptance tests run against PR and all seems good. Code reviewed. Approved.

arroyoAle (approved):
There is a drop in test coverage which is indirectly caused by this PR as the missed lines are no longer being used by this section and the other section ([view_respondents](https://github.com/ONSdigital/response-operations-ui/blob/085fea1adac7ba57c292f855dc639f9f1334b43a/response_operations_ui/views/reporting_units.py#L109-L125)) has no test coverage, which means we miss lines  223-228, 232-244 of the `party_controller.py`. But not part of this card (maybe worth another card to update the references or add some tests). Otherwise everything looks fine and continues working as expected

LJBabbage:
> There is a drop in test coverage which is indirectly caused by this PR as the missed lines are no longer being used by this section and the other section ([view_respondents](https://github.com/ONSdigital/response-operations-ui/blob/085fea1adac7ba57c292f855dc639f9f1334b43a/response_operations_ui/views/reporting_units.py#L109-L125)) has no test coverage, which means we miss lines 223-228, 232-244 of the `party_controller.py`. But not part of this card (maybe worth another card to update the references or add some tests). Otherwise everything looks fine and continues working as expected

Difficult one, might not be used anymore, but as it's an endpoint hard to track down

PR #946, opened 2024-06-10
Remove markup from message flashing

anwilkie:
# What and why?
The markup import was no longer compatible with newer versions of flask session, so it needed to be removed.

# How to test?

- Deploy
- Run acceptance tests
- Navigate to response-ops, find a secure message conversation thread, send a message, then close and re-open the conversation. Ensure that the correct pop-up messages appear in the box after each action and that the "view conversation" link works.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-1177)

### DIFF
--- a/response_operations_ui/templates/messages.html
+++ b/response_operations_ui/templates/messages.html
@@ -6,24 +6,22 @@
 {% set page_title = "Messages" %}
 {% set hide_breadcrumbs = true %}
 {% block main %}
-    {% with messages = get_flashed_messages() %}
-        {% if messages %}
-            {% call
-                onsPanel({
-                    "variant":"success",
-                    "iconType": "check",
-                    "classes": "ons-u-mb-s",
-                    "bodyAttributes": {
-                        "data-qa": "success-body"
-                    }
-                })
-            %}
-                {% for message in messages %}
-                    <p id="flashed-message-{{ loop.index }}">{{ message }}</p>
-                {% endfor %}
-            {% endcall %}
-        {% endif %}
-    {% endwith %}
+      {% if thread_id %}
+          {% call
+              onsPanel({
+                  "variant":"success",
+                  "iconType": "check",
+                  "classes": "ons-u-mb-s",
+                  "bodyAttributes": {
+                      "data-qa": "success-body"
+                  }
+              })
+          %} 
+              {% if thread_url %}
+                  <p id="flashed-message"><a href="{{thread_url}}">View conversation</a></p>
+              {% endif %}
+          {% endcall %}
+      {% endif %}
     <div class="ons-grid ons-grid--flex ons-grid--between">
         <div class="ons-grid__col">
             {% if breadcrumbs %}
--- a/response_operations_ui/templates/messages.html
+++ b/response_operations_ui/templates/messages.html
@@ -19,7 +19,7 @@
                 })
             %}
                 {% for message in messages %}
-                    <p id="flashed-message-{{ loop.index }}">{{ message }}</p><a>{{test_value}}</a>
+                    <p id="flashed-message-{{ loop.index }}">{{ message }}</p>
                 {% endfor %}
             {% endcall %}
         {% endif %}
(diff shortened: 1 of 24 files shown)


response_operations_ui/templates/messages.html
```diff
@@ -19,7 +19,11 @@
                 })
             %}
                 {% for message in messages %}
-                    <p id="flashed-message-{{ loop.index }}">{{ message }}</p>
+                    <p id="flashed-message-{{ loop.index }}">{{ message }}
+                        {% if thread_url %}
```
arroyoAle, line 23: If we ever need to extend this pattern to be used with other pages we might want it with a more extensible name like `flashed_url` or something. But this is also fine as this instance is specific to the messages and we have another flashing html that is used for everything else

response_operations_ui/views/messages.py
```diff
@@ -548,6 +545,8 @@ def view_selected_survey(selected_survey):  # noqa: C901
             ru_ref_filter=ru_ref_filter,
             tab_titles=_get_tab_titles(tab_counts, ru_ref_filter),
             show_pagination=bool(tab_counts["current"] > limit),
+            thread_id=thread_id,
+            thread_url=thread_url,
```
arroyoAle, line 549: As above this could be `flashed_url=thread_url` but is fine as is as well

response_operations_ui/views/messages.py
```diff
@@ -532,6 +517,18 @@ def view_selected_survey(selected_survey):  # noqa: C901
...
 
+        if thread_id:
+            thread_url = url_for(
+                "messages_bp.view_conversation",
+                thread_id=thread_id,
+                conversation_tab=conversation_tab,
+                page=page,
+                ru_ref_filter=ru_ref_filter,
+                business_id_filter=business_id_filter,
+            )
```
arroyoAle, line 528: This is not quite the right url its causing a different reaction from the site it needs the latest message bit added to the end 
```
thread_url = (
           url_for(
                    "messages_bp.view_conversation",
                    thread_id=thread_id,
                    page=page,
                    conversation_tab=conversation_tab,
                    ru_ref_filter=ru_ref_filter,
                    business_id_filter=business_id_filter,
                )
                + "#latest-message"
            )
```

arroyoAle (changes requested):
It works almost right there's just one bit missing from the url

response_operations_ui/views/messages.py
anwilkie: Fixed.

arroyoAle (approved):
Looks good, links work as expected and tests pass

response_operations_ui/views/messages.py
```diff
@@ -548,6 +548,8 @@ def view_selected_survey(selected_survey):  # noqa: C901
             ru_ref_filter=ru_ref_filter,
             tab_titles=_get_tab_titles(tab_counts, ru_ref_filter),
             show_pagination=bool(tab_counts["current"] > limit),
+            thread_id=thread_id,
```
LJBabbage, line 551: Why do you need to pass thread_id? the if in the template is thread_url

response_operations_ui/views/messages.py
```diff
@@ -560,6 +562,7 @@ def view_selected_survey(selected_survey):  # noqa: C901
             displayed_short_name=displayed_short_name,
             response_error=True,
             tab_titles=_get_tab_titles(),
+            thread_id=thread_id,
```
LJBabbage, line 565: Not sure why this would be here as well

response_operations_ui/views/messages.py
```diff
@@ -480,6 +464,7 @@ def view_selected_survey(selected_survey):  # noqa: C901
     limit = request.args.get("limit", default=10, type=int)
     conversation_tab = request.args.get("conversation_tab", default="open")
     ru_ref_filter = request.args.get("ru_ref_filter", default="")
+    thread_id = request.args.get("thread_id", default="")
```
LJBabbage, line 467: doesn't need the default, get has a standard default of None, which will be treated the same as empty string

response_operations_ui/views/messages.py
```diff
@@ -532,6 +517,21 @@ def view_selected_survey(selected_survey):  # noqa: C901
 
         pagination = pagination_processor(tab_counts["current"], limit, page, href)
 
+        if thread_id:
```
LJBabbage, line 520: I would be tempted to one line it (naturally over multiple lines), but being chunky it could be left. I would eyeball how it looked.

LJBabbage (changes requested):
This works for the survey mail box, but won't for technical or Respondent Feedback Team (misc) There should be a test added, plus a few extra comments to look at

tests/views/test_message.py
```diff
@@ -192,12 +192,13 @@ def test_technical_inbox_threads_list(self, mock_request, mock_get_jwt):
         of the messages.  Once we do, we can know what the requirements are (i.e., will messages of certain categories
         have survey, business and collection exercise information?)
         """
+        thread_id = "ff4537df-2097-4a73-a530-e98dba7bf28f"
```
LJBabbage, line 195: Strictly speaking this isn't quite right, although this will give line coverage, the fact that no additional assert has been added tells me that something is not right, whether it be another assert is needed or the code isn't working as expected. We should also have clear tests that focus on one functional part, there should be a test just for a thread_id being passed to the page and the result, and it should be named appropriately

SteveScorfield (approved):
Tested and works correctly. Code looks fine. Approved.

tests/views/test_message.py
```diff
@@ -210,6 +209,24 @@ def test_technical_inbox_threads_list(self, mock_request, mock_get_jwt):
             'href="/respondents/respondent-details/ff4537df-2097-4a73-a530-e98dba7bf28f"'.encode(), response.data
         )
 
+    @requests_mock.mock()
+    @patch("response_operations_ui.controllers.message_controllers._get_jwt")
+    def test_technical_inbox_threads_list_with_thread_id(self, mock_request, mock_get_jwt):
+        print("START OF TEST")
```
LJBabbage, line 215: This isn't ready for review, please look through the code again

tests/views/test_message.py
```diff
@@ -212,14 +212,11 @@ def test_technical_inbox_threads_list(self, mock_request, mock_get_jwt):
     @requests_mock.mock()
```
LJBabbage, line 212: Ok, so you might need an approach change. When writing tests the first thing to do is make sure it's obvious the process you are testing and you do that with the name and assert. If the test gets complicated, you then you bring in Given, When, Then to help.
So whats this test for? Checking a link appears when a thread_id is present in the url, which unfortunately that name doesn't tell me that and the assert either. That assert string is far to ambiguous and definitely doesn't confirm there now is a link. We want tests as obvious as possible so when another dev comes along they read it and instantly now the action it is testing

Equally I am not a fan of the that jwt blah thing either, I know it's been used elsewhere but for me that's just proliferating rubbish. It doesn't have to 100% accurate the return and you don't need to change anywhere else, just try not to copy code like that

LJBabbage (approved):

PR #952, opened 2024-07-19
RAS-1147 Upgrade to flask-session 0.8.0 update in rOps

SteveScorfield:
# What and why?
Bumping `flask-session` to version latest version which had breaking changes (for 0.8.0) to how it handled datetime objects.  With the updated code from [RAS-1177](https://jira.ons.gov.uk/browse/RAS-1177), the latest version can now be used. 

# How to test?
- Deploy this PR to your environment
- sign into rOps with the default `uaa_user` account
- check that you can send a new message
- click around making sure no 500 are raised because of a comparison error between str and datetime objects

# Jira
[RAS-1147](https://jira.ons.gov.uk/browse/RAS-1147)


tests/views/test_sign_in.py
```diff
@@ -149,4 +150,16 @@ def test_sign_out_deleting_session_variables(self):
         self.assertEqual(response.status_code, 200)
         self.assertNotIn(b"Sign out", response.data)
         with self.client.session_transaction() as session:
-            self.assertEqual(session.get("next"), None)
+            self.assertIsNone(session.get("next"))
+
+        @requests_mock.mock()
+        def test_session_holds_permission_expiry_as_iso8601_date(self, mock_request):
```
LJBabbage, line 156: This is a detached test of datetime.fromisoformat and not testing our code. It should realistically be hitting line 472 on uaa_controller.py, it's just the code used on that line is the same as this at the moment. If in the future that line was changed, this test is affectively redundant or is not right

LJBabbage (changes requested):
Works fine, but the test isn't quite right

tests/controllers/test_uaa_controller.py
```diff
@@ -282,3 +285,34 @@ def test_reset_user_password_by_id_password_change_success(self, mock_request):
             output = uaa_controller.change_user_password_by_id(user_id, user_password)
             self.assertEqual(output.status_code, 200)
             self.assertEqual(output.json(), password_change_success_json)
+
+    @requests_mock.mock()
+    def test_successful_processing_of_permissions_expiry_with_a_iso8601_date(self, mock_request):
```
LJBabbage, line 290: We realistically are testing the _If_ is processed correctly, the result of which is refresh is called. With that in mind we should mock that function and confirm just that. At the moment the test is generic I haven't broken. It also means we can get rid of the other mocks, something like 

```    
   @patch("response_operations_ui.controllers.uaa_controller.refresh_permissions")
    def test_user_has_permission_refresh_permissions_isoformat(self, refresh_permissions):
        
        with self.app.test_request_context():
            session["permissions"] = {
                "groups": test_groups,
                "expiry": datetime.isoformat(datetime.now() + timedelta(minutes=-5)),
            }
            uaa_controller.user_has_permission("oauth.approvals", user_id)
            refresh_permissions.assert_called_with(user_id)

Comment (LJBabbage) [reply tests/controllers/test_uaa_controller.py:290]: The function name(s) should also reflect the action, I am refreshing the permissions using a isoformat date

File: tests/controllers/test_uaa_controller.py [hunk 3]
```diff
@@ -282,3 +285,34 @@ def test_reset_user_password_by_id_password_change_success(self, mock_request):
...
+    def test_successful_processing_of_permissions_expiry_with_a_datetime(self, mock_request):
+        mock_request.post(url_uaa_token, json={"access_token": self.access_token}, status_code=201)
+        mock_request.get(url_uaa_user_by_id, json=uaa_user_by_id_json, status_code=200)
+        with self.app.test_request_context():
+            session["permissions"] = {"groups": test_groups, "expiry": str(datetime.now() + timedelta(minutes=5))}
+            self.assertTrue(uaa_controller.user_has_permission("oauth.approvals", user_id))
+
+    @requests_mock.mock()
+    def test_unsuccessful_processing_of_permissions_expiry_with_an_incorrect_datetime_format(self, mock_request):
+        mock_request.post(url_uaa_token, json={"access_token": self.access_token}, status_code=201)
```
LJBabbage, line 310: You don't have to patch and assert a call here, however your mocks mock_request.post(url_uaa_token, json={"access_token": self.access_token}, status_code=201) etc are directly related to the refreshing function, which it won't reach because it falls over before then. (i.e they can be removed)

LJBabbage (changes requested):

tests/controllers/test_uaa_controller.py
SteveScorfield: These have been updated.

tests/controllers/test_uaa_controller.py
SteveScorfield: This has been updated

matthew-robinson-ons (approved):
All working after the version bump

PR #955, opened 2024-07-31
Potential fix for RAS-1147

SteveScorfield:
# What and why?
After testing and deploying https://github.com/ONSdigital/response-operations-ui/pull/952 it was found that if a user was already logged in then their session with an incorrect iso date format would still persist. This would lead to the error that the original ticket was meant to fix. I have added a try to check the date is of isoformat and if it is not, then the except will deal with converting it into the correct format.

# How to test?

1. Deploy rOps version 3.1.78 to your dev env
2. Sign into rOps and navigate to surveys
3. Deploy version 3.1.79 and refresh rOps or move to another screen
4. An ValueError should appear with regards to an incorrect isoformat 
5. Deploy this branch
6. Refresh the screen, this error should no longer appear

Also run the unit tests

# Jira
https://jira.ons.gov.uk/browse/RAS-1147

### DIFF
--- a/_infra/helm/response-operations-ui/Chart.yaml
+++ b/_infra/helm/response-operations-ui/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version
-version: 3.1.80
+version: 3.1.81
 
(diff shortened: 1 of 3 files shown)


SteveScorfield:
Not needed

PR #961, opened 2024-08-09
Fix secure message return to inbox

SteveScorfield:
# What and why?
Return to the correct inbox was not working correctly for some technical and misc messages when there is recorded survey ID in the secure_message table. The code has been updated so that all technical and misc messages will return to the correct inbox.

# How to test?

1. Run unit tests
2. Deploy this PR to your dev env and run the acceptance tests
3. Create a technical message and misc message via Frontstage
4. In rOps got to navigate to the 2 messages and click on them
5. Click the return button and you should be redirected to the correct inbox

# Jira
https://jira.ons.gov.uk/browse/RAS-1224


anwilkie (approved):

anwilkie (approved):

response_operations_ui/views/messages.py
```diff
@@ -106,6 +106,12 @@ def view_conversation(thread_id):
     category = thread_conversation["category"]
     refined_thread = [_refine(message, category) for message in reversed(thread_conversation["messages"])]
     latest_message = refined_thread[-1]
+
+    # In some cases the survey_id is added to the technical and misc messages. This causes rerouting to the survey inbox
+    # By removing the survey_id from the message thread, we are redirected back to either technical or misc inboxes
```
pricem14pc, line 111: While this may resolve the issue I guess the question is whether the `survey_id` can be omitted to technical and misc messages in the first place?

LJBabbage (changes requested):
This is only part of the story, perhaps it's not explicit in the card, but the original story was to make sure the green message box, with a link to message (closed, message sent) is present. Unfortunately this still doesn't pass that information for these inboxes

response_operations_ui/views/messages.py
SteveScorfield: This was an erroneous decision based on poor data. It has been removed.

anwilkie (changes requested):
When I deployed and tested this PR out, I made the following observation:
1. Create a technical message on frontstage.
2. On r-ops, locate the message and assign the thread to a particular survey.
3. Re-assign the thread back to technical/misc.
4. In the message thread view, click on the "return to messages" button at the bottom or the "Messages" hyperlink at the top.

You will be redirected to the inbox for the survey that the thread used to be assigned to, rather than the technical/misc inbox.

SteveScorfield:
@anwilkie sorry I forgot to update the test steps. You no longer need to make changes in the DB.

response_operations_ui/templates/conversation-view/cv-back-link.html
```diff
@@ -1 +1,4 @@
+{% if thread_data.category != 'SURVEY' %}
```
LJBabbage, line 1: This feels wrong to me, and down to the fact we are passing something we shouldn't, line 186 in messages.py  shouldnt pass it, if not needed

response_operations_ui/views/messages.py
```diff
@@ -130,6 +130,12 @@ def view_conversation(thread_id):
                         form, thread_id=refined_thread[0]["thread_id"], category=thread_conversation["category"]
                     )
                 )
+                # Required for _view_select_survey to know what inbox to correctly redirect to
```
LJBabbage, line 133: All of this session stuff isn't the best, this extends a muddle, where we should just pass cleanly. It can stay, but we really need a story which removes setting session variable like this, as this gets ugly if people jump around the page by url

tests/test_data/message/thread_misc.json
```diff
@@ -0,0 +1,66 @@
+{
```
LJBabbage, line 1: There is a lot of additional replication here, isn't there nothing else in the code that you can't reuse/extend?

tests/test_data/message/thread_technical.json
```diff
@@ -0,0 +1,66 @@
+{
```
LJBabbage, line 1: Even if there is nothing to reuse/extend, these 2 JSON files can be minimised, are you purely changing the category?

tests/views/test_message.py
```diff
@@ -1612,6 +1615,128 @@ def test_conversation_reply_no_edit_permission(self, mock_request, mock_get_jwt)
...
 
+    @requests_mock.mock()
+    @patch("response_operations_ui.controllers.message_controllers._get_jwt")
+    @patch("flask_login.utils._get_user")
+    def test_flash_redirect_to_technical_inbox(self, mock_request, current_user, mock_get_jwt):
+        with self.client.session_transaction() as session:
+            session["messages_survey_selection"] = "technical"
+        with self.client.session_transaction() as session:
+            session["user_id"] = "test-id"
+        mock_get_jwt.return_value = "blah"
```
LJBabbage, line 1626: Not a fan of this, although I realised you have copied it

tests/views/test_message.py
```diff
@@ -1612,6 +1615,128 @@ def test_conversation_reply_no_edit_permission(self, mock_request, mock_get_jwt)
         self.assertNotIn("Close conversation".encode(), response.data)
         self.assertNotIn("Send message".encode(), response.data)
 
+    @requests_mock.mock()
```
LJBabbage, line 1618: Not a fan of these tests, sorry. Do we have to use the client? because it really is causing a lot of mocking, maybe we are do but it feels a lot of boiler plate between the tests (duplication), can it not be done cleaner?  I also don't really understand the difference between test_flash_redirect_to_technical_inbox and test_return_to_technical_inbox

LJBabbage (changes requested):

response_operations_ui/templates/conversation-view/cv-back-link.html
SteveScorfield: This has been updated and removed.

response_operations_ui/views/messages.py
SteveScorfield: This has been updated.

tests/test_data/message/thread_misc.json
SteveScorfield: These have been removed in favour of updating the 'category' in thread_json directly.

tests/test_data/message/thread_technical.json
SteveScorfield: As above.

tests/views/test_message.py
SteveScorfield: I had meant to update them but forgot. This has now been done.

tests/views/test_message.py
SteveScorfield: This has been improved along with the other tests.

The reason for the split in tests is that one is using the return to survey link whilst the other is being redirected via the send message action. I've updated the names of the tests to better reflect this.

response_operations_ui/views/messages.py
```diff
@@ -115,6 +115,10 @@ def view_conversation(thread_id):
     if latest_message["unread"] and _can_mark_as_unread(latest_message):
         message_controllers.remove_unread_label(latest_message["message_id"])
 
+    if category != "SURVEY":
+        # Required to produce the correct back link for the messages
+        refined_thread[0]["survey"] = ""
```
LJBabbage, line 120: Realistically we should be doing this in the dict create stage, i.e line 783. To create it then change it is not something we want to do, however I see that refine is being used for other things, but I reckon it's viable in all those situations. With that being said you can leave as it is, as it is already a bit of a mess, perhaps do a nicer var rather than updating the dict.

tests/views/test_message.py
```diff
@@ -1612,6 +1579,118 @@ def test_conversation_reply_no_edit_permission(self, mock_request, mock_get_jwt)
...
+        mock_request.get(url_get_threads_list, json=thread_list)
+        mock_request.get(url_get_surveys_list, json=survey_list)
+        mock_request.get(url_messages + "/count", json={"total": 1}, status_code=200)
+        mock_request.post(url_sign_in_data, json={"access_token": self.access_token}, status_code=201)
+        response = self.client.post(
+            "/messages/threads/fb0e79bd-e132-4f4f-a7fd-5e8c6b41b9af", data=self.message_form, follow_redirects=True
+        )
+        self.assertIn("Message sent.".encode(), response.data)
+        self.assertIn("Messages".encode(), response.data)
+        self.assertIn("RFT Messages".encode(), response.data)
```
LJBabbage, line 1630: Is that what the inbox is called RFT?

LJBabbage: One of these tests in each inbox needs to assert that a link is present, to return to the message

LJBabbage (changes requested):
A few minor issues, I think we might have to accept that all the tests in this area need work, the mocks are terrible, but good work on the tidy up. The only other query is the pipfile.lock update, it shouldn't really be done here

response_operations_ui/views/messages.py
SteveScorfield: Good spot. I've moved this to the dictionary.

tests/views/test_message.py
SteveScorfield: Yeah it's called the "RFT Messages" inbox.

I've updated the tests.

response_operations_ui/views/messages.py
```diff
@@ -813,7 +814,10 @@ def _refine(message: dict, category: str = "SURVEY") -> dict:
         refined_message["from_internal"] = False
 
     if message.get("survey_id"):
-        refined_message["survey"] = get_survey_short_name_by_id(message.get("survey_id"))
+        if category != "SURVEY":
```
LJBabbage, line 817: Done a deeper dive and this isn't going to work. I also didn't fully appreciate the empty string is not really appropriate either, this is endpoint jumping due to poor endpoint creation, the empty string causes it to access a different endpoint, which we should avoid

LJBabbage (changes requested):

response_operations_ui/views/messages.py
```diff
@@ -130,6 +130,7 @@ def view_conversation(thread_id):
                         form, thread_id=refined_thread[0]["thread_id"], category=thread_conversation["category"]
                     )
                 )
+                flash("Message sent.")
```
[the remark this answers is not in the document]
SteveScorfield: Updatted

tests/views/test_message.py
```diff
@@ -758,14 +754,11 @@ def test_conversation_fail(self, mock_request, mock_get_jwt):
     @patch("response_operations_ui.controllers.message_controllers._get_jwt")
     @patch("flask_login.utils._get_user")
     def test_conversation_reply(self, mock_request, current_user, mock_get_jwt):
-        with self.client.session_transaction() as session:
-            session["messages_survey_selection"] = "ASHE"
+        self._client_session_("Ashe")
```
arroyoAle, line 757: Is this not meant to be caps?

tests/views/test_message.py
```diff
@@ -964,20 +955,16 @@ def test_get_close_conversation_confirmation_page_for_categories(self, mock_requ
     @requests_mock.mock()
     @patch("response_operations_ui.controllers.message_controllers._get_jwt")
     def test_close_conversation(self, mock_request, mock_get_jwt):
-        sign_in_with_permission(self, mock_request, user_permission_messages_edit_json)
-        with self.client.session_transaction() as session:
-            session["messages_survey_selection"] = "Ashe"
-        mock_get_jwt.return_value = "blah"
+        self._client_session_("Ashe")
```
arroyoAle, line 958: Again

tests/views/test_message.py
```diff
@@ -987,15 +974,13 @@ def test_close_conversation(self, mock_request, mock_get_jwt):
     @requests_mock.mock()
     @patch("response_operations_ui.controllers.message_controllers._get_jwt")
     def test_close_conversation_http_error(self, mock_request, mock_get_jwt):
+        self.client.get("/logout", follow_redirects=True)
         sign_in_with_permission(self, mock_request, user_permission_messages_edit_json)
-        with self.client.session_transaction() as session:
-            session["messages_survey_selection"] = "Ashe"
-        mock_get_jwt.return_value = "blah"
+        self._client_session_("Ashe")
```
arroyoAle, line 979: As above

tests/views/test_message.py
```diff
@@ -1004,10 +989,8 @@ def test_close_conversation_http_error(self, mock_request, mock_get_jwt):
     @requests_mock.mock()
     @patch("response_operations_ui.controllers.message_controllers._get_jwt")
     def test_reopen_conversation(self, mock_request, mock_get_jwt):
-        sign_in_with_permission(self, mock_request, user_permission_messages_edit_json)
-        with self.client.session_transaction() as session:
-            session["messages_survey_selection"] = "Ashe"
-        mock_get_jwt.return_value = "blah"
+        self._client_session_("Ashe")
```
arroyoAle, line 992: As above

tests/views/test_message.py
```diff
@@ -1063,11 +1044,8 @@ def test_calculate_page_zero_threads(self):
     @requests_mock.mock()
     @patch("response_operations_ui.controllers.message_controllers._get_jwt")
     def test_mark_unread_returns_expected_flash_message(self, mock_request, mock_get_jwt):
-        mock_get_jwt.return_value = "blah"
-
-        with self.client.session_transaction() as session:
-            session["messages_survey_selection"] = "ASHE"
-
+        self._client_session_("Ashe")
```
arroyoAle, line 1047: As above

tests/views/test_message.py
```diff
@@ -1083,18 +1061,16 @@ def test_mark_unread_returns_expected_flash_message(self, mock_request, mock_get
...
         """
         limit = 10
         page = 2
         thread_id = "fb0e79bd-e132-4f4f-a7fd-5e8c6b41b9af"
 
-        sign_in_with_permission(self, mock_request, user_permission_messages_edit_json)
-        with self.client.session_transaction() as session:
-            session["messages_survey_selection"] = "Ashe"
-        mock_get_jwt.return_value = "blah"
+        self._client_session_("Ashe")
```
arroyoAle, line 1072: As above

arroyoAle: There are more below but the spirit is here

arroyoAle (commented):
Looks good and works just a lot of the same comment with a question

tests/views/test_message.py
SteveScorfield: I was just following the other tests but yeah you are correct. I've updated them

LJBabbage (approved):

arroyoAle (approved):
Works as expected

PR #969, opened 2024-12-06
RAS-1397 Remove party respondents/email calls from ROPS

LJBabbage:
# What and why?
These functions aren't used and cause a degree of confusion when checking what is calling to the respondents/email
# How to test?
Not sure there is anything to test, just make sure the functions aren't used
# Jira


Makefile
```diff
@@ -13,14 +13,14 @@ load-design-system-templates:
 	pipenv run ./scripts/load_templates.sh $(DESIGN_SYSTEM_VERSION)
 
 lint:
-	pipenv check ./response_operations_ui ./tests --ignore=70612
+	pipenv check ./response_operations_ui ./tests --ignore=70612 --ignore=73711
```
LJBabbage, line 16: Only added so this will build, there is another PR I have raised which will fix this

matthew-robinson-ons (approved):
Looks good, couldn't find any other uses. Will review the other card to fix the added ignore.

matthew-robinson-ons (commented):
This repo needs a CODEOWNERS file.

Makefile
arroyoAle: I think we may be able to remove the other ignore as well and its not coming up when running the lint command

arroyoAle (approved):
Looks good, not used anywhere and the functionality that would use it still works without it in place

Makefile
LJBabbage: I've just merged the other one, so will see if I can get rid of both, as well as adding in CODEOWNERS

PR #972, opened 2025-01-09
Fix my account error summary

matthew-robinson-ons:
# What and why?
The error summary for the my account page was mislabelled. This PR fixes the summary and solves the focus order problem it caused, highlighted in the DAC report. 
# How to test?
Deploy this PR and check a correct error summary is generated at the top of the page instead of an info summary. 
# Jira
https://jira.ons.gov.uk/browse/RAS-1361


arroyoAle (approved):
Top panel is now an error and draws initial focus which fixes the DAC report issue

mfcrocker (approved):

PR #974, opened 2025-01-15
Add tablist role to RU tabs

mfcrocker:
# What and why?
DAC report - in order for screenreader software to properly detect that tabs are part of the same group, the tablist role needs to be added to the <ul> element

# How to test?
Do a respondent search and inspect the HTML

# Jira
RAS-1363


matthew-robinson-ons (approved):
tablist role is now showing on the reporting-units-tabs page

anwilkie (approved):

PR #986, opened 2025-05-01
RAS-1546 Create new 'Collection instruments step 2 of 2' page

LJBabbage:
# What and why?
Adds in the summary page for CIR 
# How to test?
create a collection exercise in rops, then navigated to http://localhost:8085/surveys/<survey_short_name>/<period_id>/view-sample-ci/summary,  it should be empty, but works. Now add a collection instrument into the exercise and return to the page, it should now show the table populated
# Jira

### DIFF
--- a/response_operations_ui/templates/collection_exercise/view-sample-ci-summary.html
+++ b/response_operations_ui/templates/collection_exercise/view-sample-ci-summary.html
@@ -1,0 +1,42 @@
+{% extends "layouts/base.html" %}
+{% from "components/table/_macro.njk" import onsTable %}
+
+{% set  formTypeCIVersionTable = {
+    "id": "form-type-ci-version",
+    "ths": [{
+        "value": "EQ form type"
+    },
+    {
+        "value": "CIR version"
+    }]
+}%}
+
+{% set formTypeCIVersionTableDataRows = [] %}
+
+{% for collection_instrument in collection_instruments %}
+    {% do formTypeCIVersionTableDataRows.append(
+        {
+            "tds": [
+                {
+                    "value": collection_instrument["classifiers"]["form_type"]
+                },
+                {
+                  "value": "Nothing selected",
+                  "tdClasses": "ons-u-ta-centre"
+                }, 
+                {
+                  "value": '<a href="">Choose a version</a>',
+                  "tdClasses": "ons-u-ta-right"
+                }
+              ] 
+          }
+      )
+    %}
+    {% endfor %}
+    {% do formTypeCIVersionTable | setAttribute("trs", formTypeCIVersionTableDataRows) %}
+
+{% block main %}
+    <h1 class="ons-u-fs-xl" >Collection instruments step 2 of 2</h1>
+    <h3 class="ons-u-fs-m">Choose a CIR version for each eQ form type</h3>
+    {{ onsTable(formTypeCIVersionTable) }}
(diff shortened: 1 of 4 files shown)


response_operations_ui/templates/collection_exercise/view-sample-ci-summary.html
```diff
@@ -0,0 +1,42 @@
...
+              ] 
+          }
+      )
+    %}
+    {% endfor %}
+    {% do formTypeCIVersionTable | setAttribute("trs", formTypeCIVersionTableDataRows) %}
+
+{% block main %}
+    <h1 class="ons-u-fs-xl" >Collection instruments step 2 of 2</h1>
+    <h3 class="ons-u-fs-m">Choose a CIR version for each eQ form type</h3>
```
SteveScorfield, line 40: Shouldn't the initialism 'eQ' here be 'EQ' to match that of the Figma document?

SteveScorfield (changes requested):

SteveScorfield (commented):
Everything works as expected and the unit tests work. Code looks good. Just one change required I believe.

response_operations_ui/templates/collection_exercise/view-sample-ci-summary.html
LJBabbage: In the beginning it was always eQ, but it seems to have changed over time, I will update it

matthew-robinson-ons: very minor but formtype is one word in all the figma diagrams too.

matthew-robinson-ons (approved):
Looks good, page appears correctly and code looks good

response_operations_ui/views/collection_exercise.py
```diff
@@ -190,6 +190,12 @@ def view_collection_exercise(short_name, period):
     )
 
 
+def _validate_exercise(exercise: dict, period: str, short_name: str):
```
matthew-robinson-ons, line 193: is this a preferable approach to doing a try catch within view_sample_ci_summary?

LJBabbage: I just copied an existing pattern and rationalised it. However I hadn't noticed this isn't correct and isn't working as expected, get_collection_exercises_by_survey is raising a uncaught ApiError into the view, darn

response_operations_ui/templates/collection_exercise/view-sample-ci-summary.html
LJBabbage: I assumed that this was a mistake on their part, formtype isn't a word and it not being capitalised so isn't a name. Perhaps I should leave as is and let them correct it

response_operations_ui/views/collection_exercise.py
LJBabbage: Muddy puddle time, the code needs a refactor in this area. get_collection_exercise_by_period is the main issue, get_collection_exercises_by_survey does raise an uncaught exception, but it isn't dreadful. Its get_collection_exercise_by_period that should encapsulate the failing of a period to be found. Unfortunately it's used in 8 places, but all for the same thing, so could be rationalised. I'm going to run away and hide from this one and create a different story to correct this muddle

SteveScorfield (approved):
I'm happy with the change. Approved.

PR #994, opened 2025-05-12
Navigation to/from new 'Choose CIR version for EQ formtype' Page

mfcrocker:
# What and why?
This adds breadcrumbs and a save button to the CIR selection page. The save button just returns you to the CE overview.

# How to test?
Check the breadcrumb and save button send you to the right places

# Jira
RAS-1570


response_operations_ui/views/collection_exercise.py
```diff
@@ -1138,7 +1138,17 @@ def view_sample_ci_summary(short_name: str, period: str) -> str:
...
+        {"text": "Back to CIR versions", "url": "/surveys/" + short_name + "/" + period + "/view-sample-ci/summary"},
+        {},
+    ]
+    return render_template("collection_exercise/ci-versions.html", form_type=form_type, breadcrumbs=breadcrumbs)
+
+
+@collection_exercise_bp.route("/<short_name>/<period>/view-sample-ci/summary/<form_type>", methods=["POST"])
+@login_required
+def save_ci_versions(short_name: str, period: str, form_type: str) -> str:
+    return redirect(url_for("collection_exercise_bp.view_collection_exercise", short_name=short_name, period=period))
```
arroyoAle, line 1151: Exactly what the card is asking for, but it feels wrong to me, I think it should be sending the user back to the 'previous' page, the step 2 of 2 with the big table since they might be going through the CIs one by one to update the versions. But that might get done on a different card

response_operations_ui/templates/collection_exercise/ci-versions.html
```diff
@@ -1,5 +1,18 @@
 {% extends "layouts/base.html" %}
+{% from "components/button/_macro.njk" import onsButton %}
 
 {% block main %}
 <h1>Choose CIR version for EQ formtype {{ form_type }}</h1>
-{% endblock %}
+<form action="" class="ons-form" method="post">
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+    <div class="ons-grid__col">
```
arroyoAle, line 8: I don't think this is enough for the design system since this is a standalone html it needs a wrapping grid class and the save button should to be in line with the header as in the design

arroyoAle (changes requested):
Mostly fine just an extra indent that shouldn't be there and a comment about the overall journey but that's separate

response_operations_ui/views/collection_exercise.py
mfcrocker: As discussed, I won't be changing something that was refined by the whole team

response_operations_ui/templates/collection_exercise/ci-versions.html
```diff
@@ -1,5 +1,18 @@
 {% extends "layouts/base.html" %}
+{% from "components/button/_macro.njk" import onsButton %}
 
 {% block main %}
 <h1>Choose CIR version for EQ formtype {{ form_type }}</h1>
-{% endblock %}
+<form action="" class="ons-form" method="post">
```
LJBabbage, line 6: You don't need the action="" here

response_operations_ui/templates/collection_exercise/ci-versions.html
```diff
@@ -1,5 +1,18 @@
...
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+    <div class="ons-grid__col">
+        {{
+        onsButton({
```
LJBabbage, line 10: Could do with a small reformat, just alignment

response_operations_ui/views/collection_exercise.py
```diff
@@ -1138,7 +1138,17 @@ def view_sample_ci_summary(short_name: str, period: str) -> str:
 @login_required
 def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
 
-    return render_template("collection_exercise/ci-versions.html", form_type=form_type)
+    breadcrumbs = [
```
LJBabbage, line 1141: f string is the norm nowadays

LJBabbage (changes requested):
Works fine, just some tidy up pieces and the need for test coverage

LJBabbage (approved):

arroyoAle (approved):
Looks good, works as it should

PR #996, opened 2025-05-14
remove lodash.template dependency

anwilkie:
# What and why?
This PR removes the lodash.template dependency, which has been deprecated for 6 years and has not received updates. Its presence in the service was causing an unresolvable dependabot alert.

# How to test?

# Jira

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -172,7 +172,6 @@
         "lodash": "^4.17.21",
         "lodash._reinterpolate": "^3.0.0",
         "lodash.debounce": "^4.0.8",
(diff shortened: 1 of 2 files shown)


arroyoAle (approved):
I can't find its usage anywhere and rops still works along with the tests so happy with its removal

matthew-robinson-ons (changes requested):
lodash.template is related to our javascript. I wasn't able to run make minify to test the javascript bundle without doing further changes so not sure this is working as intended? There is also lodash.templatesettings which is an associated package which will be redundant with lodash.template removed.

PR #1001, opened 2025-05-28
RAS-1618: Collection instruments step 2 of 2 page - content

SteveScorfield:
# What and why?
As part of the ongoing CIR integration, there was a requirement to add additional content to the CIR page 2 of 2 page. This PR accomplishes this.

# How to test?
- Deploy this PR to your dev env
- Navigate to the CIR page 2 of 2 after adding a CI to a new or existing CE
- The new content text should be present
- Run unit and acceptance tests

# Jira
https://jira.ons.gov.uk/browse/RAS-1618

### DIFF
--- a/tests/views/test_collection_exercise.py
+++ b/tests/views/test_collection_exercise.py
@@ -2858,6 +2858,11 @@ class TestCollectionExercise(ViewTestCase):
         )
         self.assertIn("Return to 000000 Collection exercise".encode(), response.data)
         self.assertIn("Choose a version".encode(), response.data)
+        self.assertIn(
+            "A collection instrument is only created once both an EQ formtype and a CIR \n    "
+            "version are selected.<br>Each formtype requires a CIR version.".encode(),
+            response.data,
+        )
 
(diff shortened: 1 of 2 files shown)


tests/views/test_collection_exercise.py
```diff
@@ -2858,6 +2858,14 @@ def test_view_sample_ci_summary(
...
         self.assertIn("Return to 000000 Collection exercise".encode(), response.data)
         self.assertIn("Choose a version".encode(), response.data)
+        self.assertIn(
+            "A collection instrument is only created once both an EQ formtype and a CIR".encode(),
+            response.data,
+        )
+        self.assertIn(
+            "version are selected.<br>Each formtype requires a CIR version.".encode(),
+            response.data,
+        )
```
LJBabbage, line 2868: Going forward we don't need to test pure content on a page like this. Using it as verification that we are on a certain page or it's functionally working is where we should stop

LJBabbage (approved):

PR #1004, opened 2025-06-03
CIR radio buttons

mfcrocker:
# What and why?
Turns the big JSON blob into nice radio buttons to select the right option. Will need to be modified when the data persistence card goes in to store the right data

# How to test?
Check the radio boxes appear and are formatted correctly

# Jira
RAS-1553


matthew-robinson-ons (commented):
Implementation works and can view the radio buttons. One correction to the nothing selected button and some comments about the naming but looks good as an approach.

LJBabbage (approved):
Works fine, could do with a bit of formatting, i.e onsRadios and as Matthew says cir_version.ci_version could do with a tweak

matthew-robinson-ons (approved):

PR #1014, opened 2025-06-24
RAS-1544 Introduce CI's * section and the EQ formtypes count

LJBabbage:
# What and why?
Introduces CIR version count for CE page
# How to test?
Make sure feature flag on and off work, a CE which is eQ, one that is SEFT and one that is SEFT and eQ

N.B due to me having to juggle all the combinations i.e "eQ", "SEFT" and "SEFT and eQ" and how this was implemented I cant match the exact design on the card (i.e the link has to go next to the type, with no link for CIR version). 
# Jira


response_operations_ui/templates/collection_exercise/ce-ci-section.html
```diff
@@ -30,6 +30,7 @@
   {% endif %}
   
     {% for ci_type in context["ci_table"]["ci_details"] %}
+      {% set ciLink = '<a href="' + ci_type.url + '" id="view-choose-upload-ci-' + ci_type.type + '">' + ci_type.link_text + '</a>' if ci_type.type != "cir" %}
```
matthew-robinson-ons, line 33: At first glance `ci_type.type != "cir"` was a bit confusing but it makes sense in the context of the page.

PR #1016, opened 2025-06-26
404 cir error

LJBabbage:
# What and why?
Update 404 CIR return to new text
# How to test?
Make sure when no CIR data is returned that the right text is shown
# Jira


arroyoAle (approved):
Looks good, when a 404 is returned from the api we get the new message

PR #1015, opened 2025-06-25
POC flow for save CIR version

pricem14pc:
# What and why?

This is a fully functional POC reference PR, to demonstrate the logical flow of the "Select and Save CIR"

### Redis cache

This PR also introduces an exemplar Redis cache strategy for retrieving the `survey_ref` by survey short name and the `list_of_cir_metadata_objects`. These were chosen due to their static nature (although we do need to be careful with the CIR).

While this doesn't necessarily have to be part of the delivery for the Jira task, it is included out of curiosity of how we can reduce API calls.

It is functionally complete, including some gaps from a previous card around populating the radio options that was needed for this save feature.

# How to test?

Ensure a Wiremock is deployed to your namespace with appropriate mappings, then deploy this PR and step through the save CIR features.

# Jira

https://jira.ons.gov.uk/browse/RAS-1604

PR #1018, opened 2025-07-02
RAS-1548: Select and Save a CIR Version

SteveScorfield:
# What and why?

From [POC](https://github.com/ONSdigital/response-operations-ui/pull/1015)

As part of the ongoing CIR improvement work, we need to be able to select and save a CIR version. We also need to store it temporarily within the service. This PR accomplishes this.

# How to test?
1. Deploy this PR to your env
2. Run Wiremock (in ras-rm-docker repo) and point ROPS to it
3. Navigate to the DB and locate the 'registry_instrument' table
4. Check that there are no records
5. Log on to ROPS and goto survey QBS (139) create a new CE (you only need a bare minimum one)
6. Click 'Choose' for the EQ formtypes
7. Add forms 0001 and 0002 and select both
8. Click 'Continue to choose versions'
9. Select any of the formtypes and click 'Choose a version'
10. Select a version and click 'Save'
11. Repeat so that both formtypes have a version
12. Navigate back to the 'registry_insrument' table
13. There should be 2 records for formtypes 0001 and 0002
14. Go back to ROPS and goto the formtypes again for the CE
15. Select a one of the formtypes and click 'Choose a version'
16. This time, select 'Nothing selected' and click 'Save' 
17. Navigate back to the 'registry_insrument' table
18. Check that the formtype record has been deleted from the DB and the other is still present

Run acceptance and unit tests

# Jira
https://jira.ons.gov.uk/browse/RAS-1548


pricem14pc (commented):
Looks like you've copied my POC comments as-is. These weren't intended to be kept

response_operations_ui/views/collection_exercise.py
```diff
@@ -1162,17 +1163,22 @@ def view_sample_ci_summary(short_name: str, period: str) -> str:
 @collection_exercise_bp.route("/<short_name>/<period>/view-sample-ci/summary/<form_type>", methods=["GET"])
 @login_required
 def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
-    survey = survey_controllers.get_survey_by_shortname(short_name)
-    survey_ref = survey.get("surveyRef")
+    redis_cache = RedisCache()
+    survey = redis_cache.get_survey_by_shortname(short_name)
```
pricem14pc, line 1167: This gives us minor gains at this stage, as the subsequent `build_collection_exercise_details` then calls the same (non cache) method again anyway. A separate card will be progressed to improve that flow.

response_operations_ui/views/collection_exercise.py
```diff
@@ -1194,13 +1200,41 @@ def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
 @login_required
 def save_ci_versions(short_name: str, period: str, form_type: str):
     ci_version = request.form.get("ci-versions")
+    ce_details = build_collection_exercise_details(short_name, period, include_ci=True)
```
pricem14pc, line 1203: Again, a heavy operation to simply get the exercise_id. We will progress a separate strategy for this outside of this PR. We _could_ cache this `build_collection_exercise_details` method here, as long as we know to call it only for static data it contains, and not for anything that might change regularly. So cache it for the sole purpose of persisting the collection exercise ID.

response_operations_ui/views/collection_exercise.py
```diff
@@ -1194,13 +1200,41 @@ def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
 @login_required
 def save_ci_versions(short_name: str, period: str, form_type: str):
```
LJBabbage, line 1201: Again a bit heavy on the view

response_operations_ui/views/collection_exercise.py
```diff
@@ -1194,13 +1200,41 @@ def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
...
         return redirect(url_for("collection_exercise_bp.view_sample_ci_summary", short_name=short_name, period=period))
-    return redirect(url_for("collection_exercise_bp.view_collection_exercise", short_name=short_name, period=period))
+    else:
+        redis_cache = RedisCache()
+        survey_ref = redis_cache.get_survey_by_shortname(short_name).get("surveyRef")
+        list_of_cir_metadata_objects = None
+        if survey_ref:
+            try:
+                list_of_cir_metadata_objects = redis_cache.get_cir_metadata(survey_ref, form_type)
+            except ExternalApiError as e:
```
LJBabbage, line 1216: Proceed if exception and deal as None?

response_operations_ui/views/collection_exercise.py
```diff
@@ -1194,13 +1200,41 @@ def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
...
+                logger.info("Error Retrieving CIR metadata", survey_ref=survey_ref, form_type=form_type, error=e)
+
+            cir_metadata_object = next(
+                (node for node in list_of_cir_metadata_objects if node["guid"] == ci_version), None
```
LJBabbage, line 1220: Node doesn't mean much to me

response_operations_ui/common/redis_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    get_survey_by_shortname,
+)
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+class RedisCache:
+    SURVEY_EXPIRY = 600  # 10 mins
+
+    def get_cir_metadata(self, survey_ref, formtype):
```
LJBabbage, line 20: get with a sneaky set in it

response_operations_ui/common/redis_cache.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if not result:
+            logger.info("Key not in cache, getting value from CIR service", key=redis_key)
+            result = get_cir_metadata(survey_ref, formtype)
+            current_app.redis.set(redis_key, json.dumps(result), self.SURVEY_EXPIRY)
+            return result
+
+        return json.loads(result.decode("utf-8"))
+
+    def get_survey_by_shortname(self, short_name):
```
LJBabbage, line 43: Same as above, the function isn't name quite correct, considering there is a set in here. Realistically these should be separated

response_operations_ui/views/collection_exercise.py
```diff
@@ -1194,13 +1200,41 @@ def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
...
+        redis_cache = RedisCache()
+        survey_ref = redis_cache.get_survey_by_shortname(short_name).get("surveyRef")
+        list_of_cir_metadata_objects = None
+        if survey_ref:
+            try:
+                list_of_cir_metadata_objects = redis_cache.get_cir_metadata(survey_ref, form_type)
+            except ExternalApiError as e:
+                logger.info("Error Retrieving CIR metadata", survey_ref=survey_ref, form_type=form_type, error=e)
+
+            cir_metadata_object = next(
```
LJBabbage, line 1219: I assume by this that there is more than one coming back from redis_cache.get_cir_metadata

response_operations_ui/views/collection_exercise.py
```diff
@@ -1194,13 +1200,41 @@ def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
...
+                (node for node in list_of_cir_metadata_objects if node["guid"] == ci_version), None
+            )
+            eq_list = ce_details["collection_instruments"]["EQ"]
+            instrument_id = next((ci["id"] for ci in eq_list if ci["classifiers"].get("form_type") == form_type), None)
```
LJBabbage, line 1223: This feels like it shouldn't be in a view and be a function, where it can be explained. Looks like a recursive process to find the CI matching the form type

response_operations_ui/common/redis_cache.py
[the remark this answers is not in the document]
pricem14pc: Shame it's Python ;-)

response_operations_ui/common/redis_cache.py
pricem14pc: As above, I think this is fine. The set is not referring to the underlying data object being updated (i.e. in the db) it's the caching part of the flow. The method name (rightly) mirrors the existing/original non-cache method it's implementing the cache for

response_operations_ui/views/collection_exercise.py
pricem14pc: Yes, its the list of CIR version objects as per the response from the CIR service itself.

response_operations_ui/views/collection_exercise.py
pricem14pc: I've extracted and added some doc

response_operations_ui/controllers/collection_instrument_controllers.py
```diff
@@ -333,6 +350,46 @@ def delete_registry_instruments(collection_exercise_id: str, form_type: str):
...
+    guid: str,
+    instrument_id: str,
+    published_at: str,
+    survey_id,
+) -> str:
+    url = (
+        f'{app.config["COLLECTION_INSTRUMENT_URL"]}/collection-instrument-api/1.0.2/'
+        f"registry-instrument/exercise-id/{collection_exercise_id}"
+    )
+    payload = {
```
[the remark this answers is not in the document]
pricem14pc: I don't think so, it's so closely coupled with the `PUT` request that follows it I don't see any reason to extract or let anything else form it.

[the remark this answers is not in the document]
LJBabbage: I try to keep functions that make external calls light, but that is my preference, so can be ignored

response_operations_ui/common/redis_cache.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Gets the cir_metadata from redis or the cir service
+
+        :param short_name: str: the qualifying part of the redis key
+                                (response-operations-ui:survey:<SURVEY_REF>:<FORMTYPE>)
+        :return: Result from either the cache or the CIR service
+        """
+        redis_key = f"response-operations-ui:cir:{survey_ref}:{formtype}"
+        try:
+            result = current_app.redis.get(redis_key)
```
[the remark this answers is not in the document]
pricem14pc: Although currently duplicated in each cache method, this is intended to be a generic pattern so the `result` is as good a name as any (hit, result, object etc)

response_operations_ui/common/redis_cache.py
```diff
@@ -0,0 +1,75 @@
...
+        :return: Result from either the cache or survey service
+        """
+        redis_key = f"response-operations-ui:survey:{short_name}"
+        try:
+            result = current_app.redis.get(redis_key)
+        except RedisError:
+            logger.error("Error getting value from cache, please investigate", key=redis_key, exc_info=True)
+            result = None
+
+        if not result:
```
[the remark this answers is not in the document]
pricem14pc: Yep, I think a more generic cache strategy with lambda functions or simply abstracting these out to a single call that takes the key and fetch function as parameters. I'll raise a card to do this separately.

response_operations_ui/views/collection_exercise.py
pricem14pc: In the GET I think I did it so it returns a nicer message to the user than a 500 page. This will most likely result in the generic 500 page in subsequent methods. But yes, Ideally I'll look to return a message saying it can't be saved at this point rather than letting it fail further down.

response_operations_ui/views/collection_exercise.py
```diff
@@ -1162,17 +1163,22 @@ def view_sample_ci_summary(short_name: str, period: str) -> str:
 @collection_exercise_bp.route("/<short_name>/<period>/view-sample-ci/summary/<form_type>", methods=["GET"])
 @login_required
 def view_ci_versions(short_name: str, period: str, form_type: str) -> str:
-    survey = survey_controllers.get_survey_by_shortname(short_name)
-    survey_ref = survey.get("surveyRef")
+    redis_cache = RedisCache()
```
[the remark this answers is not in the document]
pricem14pc: Agree, I've added this to the Jira card to address the generic cache strategy

response_operations_ui/views/collection_exercise.py
```diff
@@ -1162,17 +1163,22 @@ def view_sample_ci_summary(short_name: str, period: str) -> str:
...
+        ce_details["collection_exercise"]["id"], form_type
+    )
     try:
-        cir_metadata = cir_controller.get_cir_metadata(survey_ref, form_type)
+        cir_metadata = redis_cache.get_cir_metadata(survey.get("surveyRef"), form_type)
         # Conversion to make displaying the datetime easier in the template
         for ci in cir_metadata:
             ci["published_at"] = datetime.fromisoformat(ci["published_at"]).strftime("%d/%m/%Y at %H:%M:%S")
+            ci["selected"] = ci["guid"] == (registry_instrument["guid"] if registry_instrument else False)
     except ExternalApiError as e:
```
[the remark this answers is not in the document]
pricem14pc: I think this is ok? ExternalApiError gets thrown, user friendly error message is set and displayed in the view. The logic in `for ci in cir_metadata:` _should_ be skipped over

matthew-robinson-ons (approved):
Tested the changes and all working as expected. Happy with the resolved conversations and the potential future changes.

PR #1021, opened 2025-07-18
RAS-1636 Collection exercise page – CI counts validation

LJBabbage:
# What and why?

# How to test?

# Jira

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -42,6 +42,7 @@ MarkupSafe = "*"
 WTForms = "*"
 jwcrypto = "*"
 email-validator = "*"
+pytest-mock = "*"
 
 [dev-packages]
 black = "*"
--- a/response_operations_ui/contexts/collection_exercise.py
+++ b/response_operations_ui/contexts/collection_exercise.py
@@ -1,6 +1,10 @@
 from flask import current_app as app
 from flask import url_for
 
+from response_operations_ui.controllers.collection_instrument_controllers import (
+    get_cir_instrument_count,
+)
+
 REMINDER_EMAILS = {
     "reminder": "First reminder",
     "reminder2": "Second reminder",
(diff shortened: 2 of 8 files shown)


response_operations_ui/contexts/collection_exercise.py
```diff
@@ -242,7 +248,10 @@ def _build_ci_table(
             ci_details.append(_ci_details_item("eq", "EQ formtypes", ci_count, eq_url, ci_table_link_text))
 
             if app.config["CIR_ENABLED"]:
-                ci_details.append(_ci_details_item("cir", "CIR version", 0))
+                cir_count = get_cir_instrument_count(ce_id)["registry_instrument_count"]
```
LJBabbage, line 251: This isn't ideal, but the count is tucked inside the context builder

Pipfile
```diff
@@ -42,6 +42,7 @@ MarkupSafe = "*"
 WTForms = "*"
 jwcrypto = "*"
 email-validator = "*"
+pytest-mock = "*"
```
LJBabbage, line 46: I had to update the full pipfile lock, when adding pytest-mock I was getting a strange dependency failure which I couldn't resolve, so had to update everything

anwilkie (changes requested):
I'm getting acceptance test failures after deploying this PR and the associated collection instrument PR (and I made sure the tests were okay prior to deploying).

matthew-robinson-ons (changes requested):
Code looks good and tested eq,seft and eqandseft with and without cir enabled. Able to take a survey all the way through to live and ready for live button is only appearing with the cir count matches the EQ formtypes. As cir_enabled is true in dev though we will need acceptance test changes as the ce will never go live as the counts won't match.

matthew-robinson-ons (approved):
Acceptance tests passing and the code changes look good and functionally work with the other CIR additions. We have a slight gap highlighted again on this card that if you remove a formtype from the CE the CIR version remains and so the count will also remain unchanged here as well. [RAS-1656](https://officefornationalstatistics.atlassian.net/browse/RAS-1656) will address this though so not blocking.

anwilkie (approved):

response_operations_ui/contexts/collection_exercise.py
pricem14pc: I think this is fine

PR #1023, opened 2025-07-31
Add table headers and cir version to CI table

matthew-robinson-ons:
# What and why?
This PR ensures that cir versions are visible on the view sample summary screen when the user has view permissions or the collection exercise has been set to live. It also includes the table headers to improve visibility. 
# How to test?
Check the table headers and cir version are visible for users with view permissions and users with edit permissions.
Check the table headers and cir version are visible on all CE states e.g. Created, Live, Ended etc. 

**N.B.** This needs testing with the feature flag on and off as this page is visible outside of the CIR journey entrypoint!
# Jira
https://officefornationalstatistics.atlassian.net/jira/software/c/projects/RAS/boards/1943

### DIFF
--- a/response_operations_ui/templates/collection_exercise/ce-collection-instrument-eq-section-redacted.html
+++ b/response_operations_ui/templates/collection_exercise/ce-collection-instrument-eq-section-redacted.html
@@ -26,58 +26,60 @@ ons-u-mt-l">
             <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
             <input type="hidden" name="csrf_token" value="" />
             <input type="hidden" name="ce_id" value="{{ ce.id }}" />
-            <div class="ons-grid ons-col-12@m ons-grid-flex--vertical-center ons-grid--no-wrap@m ons-u-mt-m ons-u-mb-m">
-                {% for eq_ci in all_cis_for_survey|batch(3, ' ') %}
+           <div class="ons-grid ons-col-12@m ons-grid-flex--vertical-center ons-grid--no-wrap@m ons-u-mt-m ons-u-mb-m">
+              {% if cir_enabled %}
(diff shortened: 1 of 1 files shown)


LJBabbage (approved):
All good. We probably want to be a bit more elegant with Nothing Selected, as these little if statements on existence in templates aren't great. But not needed for this card

pricem14pc (changes requested):
Personaly, I don't think it's worth dotting the `cir_enabled` logic through the template. I'd just bite the bullet and have one logic switch (with two versions of the table content). It'll make it easier to reason about, and also to remove the cir feature flag when we're done. Case in point, the table itself is inside the form_type for loop so is written out multiple times per formtype as things stand. That at least needs changing.

response_operations_ui/templates/collection_exercise/ce-collection-instrument-eq-section-redacted.html
```diff
@@ -26,8 +26,40 @@
             <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
             <input type="hidden" name="csrf_token" value="" />
             <input type="hidden" name="ce_id" value="{{ ce.id }}" />
-            <div class="ons-grid ons-col-12@m ons-grid-flex--vertical-center ons-grid--no-wrap@m ons-u-mt-m ons-u-mb-m">
-                {% for eq_ci in all_cis_for_survey|batch(3, ' ') %}
+           <div class="ons-grid ons-col-12@m ons-grid-flex--vertical-center ons-grid--no-wrap@m ons-u-mt-m ons-u-mb-m">
```
matthew-robinson-ons, line 29: Have left the existing flow the same outside of CIR, but refactored the table rendering outside of the for loops and removed some redundant batching for when we more to the CIR flow.

PR #1025, opened 2025-08-12
RAS-1665 Save button - user permissions

LJBabbage:
# What and why?
This PR hides the save button if you do not have permissions
# How to test?
Check you have a save button for CIR version changes, and not if you don't
# Jira


matthew-robinson-ons (approved):
Does what's intended, this page is a bit strange for a view user now though as it is all about 'choosing' but not for this card.

PR #1026, opened 2025-09-15
Update cir versions page for view users

matthew-robinson-ons:
# What and why?
This PR updates the CIR versions page to show a no access page if the user does not have edit permissions or if the survey is not editable (It is no longer in created state). In both these scenarios the only way to reach this page is through the URL, so until a view journey is created they shouldn't be able to access the page. 
The text of the page has been updated to reflect the figma and recent UX design. 
# How to test?
Check the CIR versions page is still accessible and usable as an edit user.
Check that view users see the no access page.
Check that surveys that are not in created state also see the no access page. 
# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1639

### DIFF
--- a/response_operations_ui/templates/errors/no-permission-error.html
+++ b/response_operations_ui/templates/errors/no-permission-error.html
@@ -9,7 +9,7 @@
             "classes": "ons-u-mb-l"
         })
     %}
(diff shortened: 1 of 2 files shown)


SteveScorfield (approved):
Tested fully and works as expected. Changes are fine. Approved.

PR #1027, opened 2025-09-18
Testing fully npm removal

SteveScorfield:
# What and why?
This PR is only intended for reference. It was created to test what happens with npm removal. It shows us that the packages we use are not required. Full testing would still be required before this could be used.

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1688


pricem14pc (commented):
There are README references to the front end frameworks (e.g. nvm and npm) that would need removing I guess?

PR #1031, opened 2025-10-02
Remove all npm (and other packages) not required in ROPS

arroyoAle:
# What and why?

# How to test?

# Jira


README.md
```diff
@@ -85,19 +57,35 @@ To update to a different version of the Design System:
...
-Similarly the js files are minimised into one file, `all.js.min`
+It is a manual step to minimise the JavaScript and scss.
 
-It is a manual step to minimise the JavaScript
+Ensure you have Node.js version >=10 installed.  The recommended way to do this is to use Creationix Node Version Manager, which works on Linux and MacOSX systems:
 
 ```bash
-make minify-install
-make minify
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
```
matthew-robinson-ons, line 72: 0.40.3 is the most recent version of nvm. This version didn't work when i attempted a clean install

README.md
```diff
@@ -85,19 +57,35 @@ To update to a different version of the Design System:
...
 ```
 
+After this, run
+
+```bash
+nvm use
```
matthew-robinson-ons, line 78: Had to use nvm install not nvm use

README.md
```diff
@@ -85,19 +57,35 @@ To update to a different version of the Design System:
...
-When the application is run the scss files are converted into css and they are minimised into one file, `all.css.min`
+Each of these working files are minimised into respective css and js files which are loaded when the application is run.
+The scss files are minimised into one file, `all.css` and the js files are minimised into two files, `main.js.min` and 
+`selected_ci_functions.min.js`
 
-Similarly the js files are minimised into one file, `all.js.min`
+It is a manual step to minimise the JavaScript and scss.
 
-It is a manual step to minimise the JavaScript
+Ensure you have Node.js version >=10 installed.  The recommended way to do this is to use Creationix Node Version Manager, which works on Linux and MacOSX systems:
```
matthew-robinson-ons, line 69: version >=20

README.md
arroyoAle: Updated to the most recent version

arroyoAle:
> Couple of documentation changes and some queries around node usage. The changes work and found no issues in the UI but we should be able to move to the design system for most/all of the JS functionality we are currently implementing. The design system has a select all checkboxes component which would replace the `selected_ci_functions`, the `validate_ci` is essentially disabling the radio button on incorrect file type upload which might also be possible and the `read_csv` is probably the most functionally complex and might not be covered by the DS. If not for this card we should explore the DS for the functionality on a separate card.

I don't think its worth replacing the select all functionality, unless we really want to, since it is going to be removed once CIR is enabled, I have removed the `validate_ci` because I cannot see how its being used, and I can't see how we would have the `read_csv` functionality only using the design system

PR #1073, opened 2026-06-01
RAS-1925: Investigate Rops interaction with collection exercise

SteveScorfield:
# What and why?
Multiple calls are being made to the CE service when navigating through a CE in ROPS. These calls are mostly unnecessary when we are just populating lists and data. For instance if I were to select a specific period of a CE from the list, it populates the page with data, already retrieved from the list page. Equally, when navigating back to the CE list page, it repopulates the data again. This is really unnecessary. From what I found, there are about 7 unnecessary calls being made to the CE service. All this data can be contained within a Redis session. Using a Redis session to store this data and refreshing it only when changes are made/session expiry ensures that the minimum of calls are made. I have made some changes that can be viewed as a starting point and doesn’t necessarily need to be implemented this way. 

# How to test?
1. Deploy this locally
2. Play about creating CEs

Note: As this is a concept, the tests are not working correctly. I didn't want to fully update the tests on a concept piece of work.

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1925

### DIFF
--- a/response_operations_ui/common/redis_cache.py
+++ b/response_operations_ui/common/redis_cache.py
@@ -5,6 +5,7 @@ from flask import current_app
 from redis.exceptions import RedisError
 from structlog import wrap_logger
 
+from response_operations_ui.controllers import collection_exercise_controllers
 from response_operations_ui.controllers.cir_controller import get_cir_metadata
 from response_operations_ui.controllers.survey_controllers import (
     get_survey_by_shortname,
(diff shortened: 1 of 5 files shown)


response_operations_ui/common/redis_cache.py
```diff
@@ -99,6 +100,35 @@ def get_survey_list(self) -> dict:
...
+    
+    def update_collection_exercises(self, survey_id: str) -> dict:
+        redis_key = f"{self.APPLICATION_KEY}:ce:{survey_id}"
+        self.remove_old_collection_exercises(survey_id)
+        logger.info("Updating collection exercises", key=redis_key)
+        result = collection_exercise_controllers.get_collection_exercises_by_survey(survey_id)
+        self.set(redis_key, json.dumps(result), self.EXPIRY)
+        return result
+    
+    def remove_old_collection_exercises(self, survey_id: str) -> str:
```
arroyoAle, line 127: Is there a reason we'd need to remove them from the cache instead of relying on the TTL to clear the exercise

arroyoAle (commented):
This looks like a good start, but I think we need a bit more information about where collection exercise is being called from and which ones might be redundant. Also might be worth a thought about wether redis is appropriate for this, although we do this for the surveys so it might be a similar, but since we create/edit collection exercise with more regularity we might want to always request the most up to date information from collection exercise.

